PR #32, opened 2018-09-19
Add release scripts

jpkrohling:
This commit adds support for releasing the operator. It adds a couple of scripts that are supposed to be executed by travis. The scripts build and push docker images, as well as change the 'deploy/operator.yaml' to use this latest release tag. Once that is done, the change to the operator.yaml is committed, a release is tagged and pushed to GitHub.


jpkrohling:
@objectiser This is a first draft, to start a discussion. It goes against the idea of having the "stream", as master will always have a `deploy/operator.yaml` that is fixed to use the last release. Not sure it's a good thing or bad thing yet.

Basically, this would work in a similar way as other projects: tag `release/v1.2.3` and the `v1.2.3` will be released.

Another thing this changes from the current behavior is that this will not publish an image to `latest`, but to BRANCH (master, for instance). I could add a conditional on the `publish-images.sh` to update `latest` whenever we are building `master`, but wanted your opinion before.

pkg/version/main.go
```diff
@@ -34,8 +33,7 @@ func Get() Version {
 	}
 
 	return Version{
-		Operator:    jaeger, // for now, we just follow the same versioning name as the main components
-		GitCommit:   commit,
+		Operator:    commit,
```
objectiser: Should this be renamed from `commit` to `operator`?

jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/32)

PR #35, opened 2018-09-20
Fix the run goal

geoand:
Instead of relying on the binary being present, we just use go run.
We also set all the necessary compilation flags, environment variables
and trap the interrupt signal in order to avoid having Make fail when
the execution of the goal is interrupted

Signed-off-by: Georgios Andrianakis <geoand@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ e2e-tests: es crd build docker push
 	@go test ./test/e2e/... -kubeconfig $(KUBERNETES_CONFIG) -namespacedMan ../../deploy/test/namespace-manifests.yaml -globalMan ../../deploy/crd.yaml -root .
 
 run: crd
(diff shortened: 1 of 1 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/35)

geoand:
Sounds good, I'll take a look

jpkrohling:
You might want also to add the `-ldflags`, as mentioned on #37

geoand:
Sure thing

geoand:
@jpkrohling The parenthesis are wrong, I need to update to use brackets :)

Unfortunately the `|| true` does not work when the goal is interrupted (it's the first thing I tried), hence the need to trap that signal explicitly

geoand:
I fixed the parenthesis issue.

Also I am also on Linux, so I don't really know how this will behave on MacOS, although I would be surprised if it didn't work :P

geoand:
Ah yes, good catch! Fixed :0

Issue #30, opened 2018-09-14
Switch to disable the ingress resource


secat:
Is it currently possible to disable the creation of the ingress resource?

If not, I think it would be a good feature to update the CRD spec with an ingress configuration section.

jpkrohling:
Would you like to give it a shot? Should be relatively easy to do.

secat:
Yes, sure I can implement the feature.

For the switch specification, I was thinking of something like:
```yaml
apiVersion: io.jaegertracing/v1alpha1
kind: Jaeger
metadata:
  name: my-jaeger
spec:
  ingress:
    enabled: false
```

What do you think about it?

jpkrohling:
Right now, the ingress is applied only to the query deployment. So, to avoid any ambiguity, it would be better to have it as a child of the JaegerQuery struct.

secat:
Should I close this issue?

jpkrohling:
I'm closing. Thanks for the PR!

sneko:
Hi @secat @jpkrohling, I tried to disable ingress but it doesn't work:
```
spec:
  strategy: all-in-one
  all-in-one:
    image: jaegertracing/all-in-one:1.7
    options:
      log-level: info
      memory:
        max-traces: 10000
  query:
    ingress:
      enabled: false

```

Could you help me please? Thanks

jpkrohling:
@sneko good catch, looks like we need something like this for the `all-in-one` as well:

https://github.com/jaegertracing/jaeger-operator/blob/e0849b584483716a3565267679146dda0b036a9d/pkg/deployment/query.go#L118-L127

Here:
https://github.com/jaegertracing/jaeger-operator/blob/e0849b584483716a3565267679146dda0b036a9d/pkg/deployment/all-in-one.go#L145-L150

Would you be interested in sending a PR for it?

jpkrohling:
Created #48 to handle the missing support in the all-in-one.

sneko:
@jpkrohling to be honest I'm quite busy on other stuff, I was justing looking at the operator to see advancement. I'm on my side using for now the Jaeger instance embedded with Istio Helm Chart.

But if I find some spare time I will dig into it (even if I'm not a Go programmer ^^)

BTW, thanks for your work 👍

jpkrohling:
That's completely understandable :) I have a number of other improvements on my queue right now, so, I'll leave it open and see if someone would pick it up as part of hacktoberfest :)

[...]

PR #36, opened 2018-09-20
Add query ingress enable switch

secat:
This PR addresses issue #30.

Update the jaeger query specification with an ingress specification containing an enable flag. When not set, defaults to creating the query ingress object resource.


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/36)

secat:
___
*[pkg/apis/io/v1alpha1/types.go, line 43 at r1](https://reviewable.io/reviews/jaegertracing/jaeger-operator/36#-LMvEDHkDy0_OMLXLW7i:-LMvj-k684wUI0G9n4Xc:bs7ivdm) ([raw file](https://github.com/jaegertracing/jaeger-operator/blob/db47374c4f005ff3d7e9ed20a3d8c497b19b06cd/pkg/apis/io/v1alpha1/types.go#L43)):*
<details><summary><i>Previously, jpkrohling (Juraci Paixão Kröhling) wrote…</i></summary><blockquote>

This struct could probably be named `JaegerIngressSpec` instead of `JaegerQueryIngressSpec`. If we ever need it for the collector or for a future new component, we could reuse it.
</blockquote></details>

Excellent idea, I will make the change.

secat:
___
*[pkg/apis/io/v1alpha1/types.go, line 51 at r1](https://reviewable.io/reviews/jaegertracing/jaeger-operator/36#-LMvETKy7t2ZQlxfAQZy:-LMvjC6kBJIyJM8qCEve:b-4b5d4o) ([raw file](https://github.com/jaegertracing/jaeger-operator/blob/db47374c4f005ff3d7e9ed20a3d8c497b19b06cd/pkg/apis/io/v1alpha1/types.go#L51)):*
<details><summary><i>Previously, jpkrohling (Juraci Paixão Kröhling) wrote…</i></summary><blockquote>

Any reason to have it as a pointer?
</blockquote></details>

It is in order to not break the default behavior. Currently a user doesn't have to specify if the ingress is enabled, it is by default.

Also, using a pointer helps detecting that case, i.e. if the flag is set or not. A `nil` pointer means it was not set, therefore use the default value.

If we do not use a pointer, the default value will be `false` if the user doesn't set the flag value.

Which direction should we go?

secat:
___
*[pkg/apis/io/v1alpha1/types.go, line 43 at r1](https://reviewable.io/reviews/jaegertracing/jaeger-operator/36#-LMvEDHkDy0_OMLXLW7i:-LMwBkVP2hdTbQiKoPjG:brotd86) ([raw file](https://github.com/jaegertracing/jaeger-operator/blob/db47374c4f005ff3d7e9ed20a3d8c497b19b06cd/pkg/apis/io/v1alpha1/types.go#L43)):*
<details><summary><i>Previously, secat (Serge Catudal) wrote…</i></summary><blockquote>

Excellent idea, I will make the change.
</blockquote></details>

I have pushed a commit which renames the struct from `JaegerQueryIngressSpec` to `JaegerIngressSpec`.

jpkrohling:
Thanks!

PR #37, opened 2018-09-20
Use "latest" as the image version when none is set

geoand:
The version can be empty when developing with "go run main.go" so to
avoid having Kubernetes fail to deploy Jaeger because the image is
incorrect, we use "latest" as the default

Signed-off-by: Georgios Andrianakis <geoand@gmail.com>


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/37)

geoand:
I am pretty new to `go` so yeah, I'm sure mine isn't the best possible solution. I will checkout the `ldflags` you mention and see what I can come up with

geoand:
I agree that we should add the `ldflags` in #35. I then close this PR ok?

PR #58, opened 2018-10-11
Inject sidecar in properly annotated pods

jpkrohling:
This PR adds support for auto-injection of Jaeger agents into target `Deployment` objects. Technically, this can be done for all objects containing `Pod` objects, like `StatefulSet` or `DaemonSet`, but this PR limits the first implementation to `Deployment`, to get it tested first. If there's demand, it's relatively easy to add support for other constructs.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -167,6 +167,35 @@ oc get routes
 
 NOTE: make sure to use `https` with the hostname/port you get from the command above, otherwise you'll see a message like: "Application is not available".
 
+== Auto injection of Jaeger Agent sidecars
+
+The operator can also inject Jaeger Agent sidecars in `Deployment` workloads, provided that the deployment has the annotation `inject-jaeger-agent` with a suitable value. The values can be either `"true"` (as string), or the Jaeger instance name, as returned by `kubectl get jaegers`. When `"true"` is used, there should be exactly *one* Jaeger instance for the same namespace as the deployment, otherwise, the operator can't figure out automatically which Jaeger instance to use.
+
+The following snippet shows a simple application that will get a sidecar injected, with the Jaeger Agent pointing to the single Jaeger instance available in the same namespace:
+
+[source,yaml]
+----
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: myapp
+  annotations:
+    inject-jaeger-agent: "true" # <1>
+spec:
+  selector:
+    matchLabels:
+      app: myapp
+  template:
+    metadata:
+      labels:
+        app: myapp
+    spec:
+      containers:
+      - name: myapp
+        image: acme/myapp:myversion
+----
+<1> Either `"true"` (as string) or the Jaeger instance name
+
 == Agent as DaemonSet
 
 By default, the Operator expects the agents to be deployed as sidecars to the target applications. This is convenient for several purposes, like in a multi-tenant scenario or to have better load balancing, but there are scenarios where it's desirable to install the agent as a `DaemonSet`. In that case, specify the Agent's strategy to `DaemonSet`, as follows:
--- a/pkg/deployment/query.go
+++ b/pkg/deployment/query.go
@@ -43,6 +43,13 @@ func (q *Query) Get() *appsv1.Deployment {
 	annotations := map[string]string{
 		"prometheus.io/scrape": "true",
 		"prometheus.io/port":   "16686",
+
+		// note that we are explicitly using a string here, not the value from `inject.Annotation`
+		// this has two reasons:
+		// 1) as it is, it would cause a circular dependency, so, we'd have to extract that constant to somewhere else
+		// 2) this specific string is part of the "public API" of the operator: we should not change
+		// it at will. So, we leave this configured just like any other application would
+		"inject-jaeger-agent": q.jaeger.Name,
 	}
 
 	return &appsv1.Deployment{
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -1,0 +1,99 @@
+package inject
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/api/core/v1"
+
+	"github.com/jaegertracing/jaeger-operator/pkg/apis/io/v1alpha1"
+	"github.com/jaegertracing/jaeger-operator/pkg/deployment"
+	"github.com/jaegertracing/jaeger-operator/pkg/service"
+)
+
+var (
+	// Annotation is the annotation name to look for when deciding whether or not to inject
+	Annotation = "inject-jaeger-agent"
+)
+
+// Sidecar adds a new container to the deployment, connecting to the given jaeger instance
+func Sidecar(dep *appsv1.Deployment, jaeger *v1alpha1.Jaeger) {
+	deployment.NewAgent(jaeger) // we need some initialization from that, but we don't actually need the agent's instance here
+
+	if jaeger == nil || dep.Annotations[Annotation] != jaeger.Name {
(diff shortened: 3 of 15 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/58)

pavolloffay:
Does it auto inject to all deployments in cluster or in namespace?

README.adoc
```diff
@@ -167,6 +167,35 @@ oc get routes
 
 NOTE: make sure to use `https` with the hostname/port you get from the command above, otherwise you'll see a message like: "Application is not available".
 
+== Auto injection of Jaeger Agent sidecars
+
+The operator can also inject Jaeger Agent sidecars in `Deployment` workloads, provided that the deployment has the annotation `inject-jaeger-agent` with a suitable value. The values can be either `"true"` (as string), or the Jaeger instance name, as returned by `kubectl get jaegers`. When `"true"` is used, there should be exactly *one* Jaeger instance for the same namespace as the deployment, otherwise, the operator can't figure out automatically which Jaeger instance to use.
```
pavolloffay: What happens if the there are more jaeger instances and inject-jaeger-agent=true? Does the deployment fail or it does not inject?

pkg/deployment/query.go
```diff
@@ -43,6 +43,13 @@ func (q *Query) Get() *appsv1.Deployment {
 	annotations := map[string]string{
 		"prometheus.io/scrape": "true",
 		"prometheus.io/port":   "16686",
+
+		// note that we are explicitly using a string here, not the value from `inject.Annotation`
+		// this has two reasons:
+		// 1) as it is, it would cause a circular dependency, so, we'd have to extract that constant to somewhere else
+		// 2) this specific string is part of the "public API" of the operator: we should not change
+		// it at will. So, we leave this configured just like any other application would
+		"inject-jaeger-agent": q.jaeger.Name,
```
pavolloffay: is this for tracing jaeger-query?

pkg/inject/sidecar.go
```diff
@@ -0,0 +1,99 @@
...
+var (
+	// Annotation is the annotation name to look for when deciding whether or not to inject
+	Annotation = "inject-jaeger-agent"
+)
+
+// Sidecar adds a new container to the deployment, connecting to the given jaeger instance
+func Sidecar(dep *appsv1.Deployment, jaeger *v1alpha1.Jaeger) {
+	deployment.NewAgent(jaeger) // we need some initialization from that, but we don't actually need the agent's instance here
+
+	if jaeger == nil || dep.Annotations[Annotation] != jaeger.Name {
```
pavolloffay: where is the `=true` case?

pkg/inject/sidecar.go
```diff
@@ -0,0 +1,99 @@
...
+	if dep.Annotations[Annotation] == "" {
+		logrus.Debugf("Not needed, annotation not present for %v", dep.Name)
+		return false
+	}
+
+	// this pod is annotated, it should have a sidecar
+	// but does it already have one?
+	hasSidecar := false
+	for _, container := range dep.Spec.Template.Spec.Containers {
+		if container.Name == "jaeger-agent" { // we don't labels/annotations on containers, so, we rely on its name
```
pavolloffay: can the name be consumed from a constant?

pkg/inject/sidecar.go
```diff
@@ -0,0 +1,99 @@
...
+			hasSidecar = true
```
pavolloffay: you can directly return false, no need to continue the loop

pkg/inject/sidecar.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+
+	return !hasSidecar
```
pavolloffay: viz previous comment this can return true.

pkg/inject/sidecar.go
```diff
@@ -0,0 +1,99 @@
...
+// Select a suitable Jaeger from the JaegerList for the given Pod, or nil of none is suitable
+func Select(target *appsv1.Deployment, availableJaegerPods *v1alpha1.JaegerList) *v1alpha1.Jaeger {
+	jaegerName := target.Annotations[Annotation]
+	if strings.ToLower(jaegerName) == "true" && len(availableJaegerPods.Items) == 1 {
+		// if there's only *one* jaeger within this namespace, then that's what
+		// we'll use -- otherwise, we should just not inject, as it's not clear which
+		// jaeger instance to use!
+		// first, we make sure we normalize the name:
+		jaeger := &availableJaegerPods.Items[0]
+		target.Annotations[Annotation] = jaeger.Name
```
pavolloffay: the target is not modified in the following loop

pkg/deployment/query.go
```diff
@@ -43,6 +43,13 @@ func (q *Query) Get() *appsv1.Deployment {
 	annotations := map[string]string{
 		"prometheus.io/scrape": "true",
 		"prometheus.io/port":   "16686",
+
+		// note that we are explicitly using a string here, not the value from `inject.Annotation`
+		// this has two reasons:
+		// 1) as it is, it would cause a circular dependency, so, we'd have to extract that constant to somewhere else
+		// 2) this specific string is part of the "public API" of the operator: we should not change
+		// it at will. So, we leave this configured just like any other application would
```
pavolloffay: I always prefer using a constant. Even we don't change it it's easier from the maintenance point of view.

pavolloffay:
btw. I find that "=true" option useful for simple deployments. Or we can default to the default name used by the operator.

README.adoc
jpkrohling: Then nothing happens, unless there's a Jaeger instance named "true" (which I'm not even sure it's a valid name). :)

pkg/deployment/query.go
jpkrohling: Yes

pkg/inject/sidecar.go
jpkrohling: In the `Select` method

pkg/inject/sidecar.go
jpkrohling: It could, yes. Perhaps it belongs to a refactoring, making all such names constants.
https://github.com/jaegertracing/jaeger-operator/issues/59

pkg/inject/sidecar.go
jpkrohling: This is modified because it might be "true", so, it becomes the Jaeger's name. It's not modified in the following loop because there, the annotation content == jaeger.Name already, so, there's nothing to normalize.

pkg/deployment/query.go
jpkrohling: A local constant, instead of the same constant as the one used in the `inject` package? Could be, in that case, I'll consider it for #59.

jpkrohling:
> I think the Jaeger instance name needs to support namespace scoping - it is very likely that the jaeger instance will be deployed into a different namespace

I actually thought about it when implementing, but I got some practical questions that I didn't want to figure out *right now*. I recorded #60 with the questions and I'd add support for inter-namespace injection in a follow up PR (as this here is already complex enough, IMO).

pkg/deployment/query.go
pavolloffay: I mean a single constant unique for the project if possible

jpkrohling: Then no, I'll keep it as string following the reasoning on the comment

pkg/inject/sidecar.go
jpkrohling: Done

cfontes:
Just pointing out that I would like to have it in Statefulsets!

jpkrohling:
@cfontes that's great to hear! Could you please open an issue with the request? Bonus points if you could describe your use case.

Issue #66, opened 2018-10-16
Update the Operator SDK


jpkrohling:
An upcoming release of the Operator SDK will break the compat with current master, which is what we currently have at `Gopkg.toml`.

This issue is to track:
- [x] Fix the operator SDK version in the dependency management file
- [ ] Bump our operator SDK version once their refactoring is done + fix the parts where it breaks us.

https://github.com/kubernetes-sigs/controller-runtime

jpkrohling:
The version 0.1.0 has been released a few days ago and did break the compatibility with previous versions. I'll start working on it now.

PR #78, opened 2018-11-01
Add support for injecting jaeger agent in annotated statefulsets.

### LINKED ISSUE #62: Inject sidecar in properly in annotated Statefulsets
I am testing out Jaeger as a distributed tracing solution, Jaeger-Operator sounded like the easiest way to do it, and here I am ( I also checked the helm chart, but this looks like the preferred way currently )

Operator worked great, but I can't use the very handy annotation below

>   annotations:
>     inject-jaeger-agent: "true"

Our system is an actor based system with sharded local state, and we use Statefulsets to maintain local state and instances together.

Of course I can use the side car in my YAML and I did that, but this annotation is a lot easier to maintain.

That is it!

annanay25:
Attempts to resolve #62.

Need to check test coverage.

### DIFF
--- a/pkg/controller/deployment/deployment_controller.go
+++ b/pkg/controller/deployment/deployment_controller.go
@@ -18,7 +18,7 @@ import (
 	"github.com/jaegertracing/jaeger-operator/pkg/inject"
 )
 
-// Add creates a new Deployment Controller and adds it to the Manager. The Manager will set fields on the Controller
+// Add creates a new Controller for Deployment and StatefulSets and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
 	return add(mgr, newReconciler(mgr))
@@ -26,13 +26,13 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileDeployment{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &ReconcileObject{client: mgr.GetClient(), scheme: mgr.GetScheme()}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Create a new controller
-	c, err := controller.New("deployment-controller", mgr, controller.Options{Reconciler: r})
+	c, err := controller.New("object-controller", mgr, controller.Options{Reconciler: r})
 	if err != nil {
 		return err
 	}
@@ -43,33 +43,39 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to primary resource StatefulSet
+	err = c.Watch(&source.Kind{Type: &appsv1.StatefulSet{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
-var _ reconcile.Reconciler = &ReconcileDeployment{}
+var _ reconcile.Reconciler = &ReconcileObject{}
 
-// ReconcileDeployment reconciles a Deployment object
-type ReconcileDeployment struct {
+// ReconcileObject reconciles a Kubernetes object
+type ReconcileObject struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
 	client client.Client
 	scheme *runtime.Scheme
 }
 
-// Reconcile reads that state of the cluster for a Deployment object and makes changes based on the state read
-// and what is in the Deployment.Spec
+// Reconcile reads that state of the cluster for a Kubernetes object and makes changes based on the
+// state read and what is in the Object Spec
 // Note:
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
-func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (r *ReconcileObject) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	log.WithFields(log.Fields{
 		"namespace": request.Namespace,
 		"name":      request.Name,
-	}).Debug("Reconciling Deployment")
+	}).Debug("Reconciling the object")
 
-	// Fetch the Deployment instance
-	instance := &appsv1.Deployment{}
-	err := r.client.Get(context.Background(), request.NamespacedName, instance)
+	// Fetch the object instance
+	instance := *runtime.Object{}
+	err := r.client.Get(context.Background(), request.NamespacedName, *instance)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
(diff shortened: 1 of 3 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/78)

jpkrohling:
Let me know once it's ready to be reviewed.

annanay25:
@jpkrohling request for review.

annanay25:
Thanks for the comprehensive review! 

I understand that we need to write sidecar injection for multiple pod types, I am looking into making the `inject` package more generic by accepting all deployment types. Will update this PR soon.

pkg/stub/handler.go
```diff
@@ -61,7 +61,7 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 			logrus.Errorf("failed to update %v", o)
 			return err
 		}
-	case *appsv1.Deployment:
+	default:
```
annanay25: To accept a larger class of Pod types without replicating code, I added this change. 

However, this would cause conflicts with sdk.List here - https://github.com/jaegertracing/jaeger-operator/pull/78/files#diff-2d7ceed365f1ea47bba386dac808cb07R72 - where we need explicit type checking.

annanay25:
@jpkrohling I was a little confused as to how to proceed here - https://github.com/jaegertracing/jaeger-operator/pull/78#discussion_r232498890

Your thoughts?

pkg/stub/handler.go
jpkrohling: Instead of accepting a more generic pod type there, you could just list all the types that should be included in this branch, like:

```golang
case *appsv1.Deployment:
case *appsv1.StatefulSet:
  ....
```

jpkrohling:
> Your thoughts?

Sorry, I missed that part. I just replied with a suggestion that might work.

pkg/stub/handler.go
annanay25: Oh. Did you mean a switch-case fall-through? We would still have to write code for type checking the `sdk.Object` in the inject package. I've made some changes, do let me know if OK. @jpkrohling

jpkrohling: Something like that, but not quite: you can keep the two cases collapsed into one, like the example in previous comment. With more context, the example would be like this:

```golang
func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
	switch o := event.Object.(type) {
	case *v1alpha1.Jaeger:
            ....
	case *appsv1.Deployment:
	case *appsv1.StatefulSet:
            if inject.Needed(o) {...}
	}
	return nil
```

The `inject.Needed` would have to be changed. Perhaps instead of accepting a `sdk.Object` (or `runtime.Object`, it could accept two parameters: a map[string]string with the annotations and a `PodTemplateSpec`. The same would probably work with the other functions, like `inject.Select` and `inject.Sidecar`. Adding another supported object type (DaemonSet) would then just be a matter of adding it to the switch case in the handler.

jpkrohling:
Heads up: I'm starting to work on #66, which will probably conflict with this PR. Please, wait a couple of days to work on this issue, as I'm not sure how much of it would be reusable there.

annanay25:
Understood. I'll hold this off. Thanks!

jpkrohling:
The migration to the new operator SDK is now complete. Please, rebase and update this PR.

pkg/inject/sidecar.go
```diff
@@ -24,28 +25,32 @@ const (
...
-	if jaeger == nil || dep.Annotations[Annotation] != jaeger.Name {
-		logrus.Debugf("Skipping sidecar injection for deployment %v", dep.Name)
-	} else {
-		decorate(dep)
-		logrus.Debugf("Injecting sidecar for pod %v", dep.Name)
-		dep.Spec.Template.Spec.Containers = append(dep.Spec.Template.Spec.Containers, container(jaeger))
+	switch o := obj.(type) {
+	case *appsv1.Deployment:
+	case *appsv1.StatefulSet:
+		if jaeger == nil || obj.Annotations[Annotation] != jaeger.Name {
```
annanay25: @jpkrohling even with type assertions here, I'm running into errors like 

```
obj.Annotations undefined (type "github.com/jaegertracing/jaeger-operator/vendor/k8s.io/apimachinery/pkg/runtime".Object has no field or method Annotations)
```

pkg/controller/deployment/deployment_controller.go
```diff
@@ -43,33 +43,39 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 		"namespace": request.Namespace,
 		"name":      request.Name,
-	}).Print("Reconciling Deployment")
+	}).Print("Reconciling the Object")
 
-	// Fetch the Deployment instance
-	instance := &appsv1.Deployment{}
-	err := r.client.Get(context.Background(), request.NamespacedName, instance)
+	// Fetch the object instance
+	instance := *runtime.Object{}
```
annanay25: Can either do type checking in this function or in each of the functions in the `inject` package.

pkg/inject/sidecar.go
jpkrohling: You should use `o` instead of `obj` here :)

annanay25: I went hunting everywhere. Thanks for pointing this out!

pkg/controller/deployment/deployment_controller.go
annanay25: The `Sidecar` function internally calls `decorate` which uses other parameters of the object, so it would be useful to pass the object around. Though there is #104 to consider.

annanay25:
@jpkrohling the case statement grouping as implemented here - https://github.com/jaegertracing/jaeger-operator/pull/78/files#diff-f88c99835d4b2acda532e36123761825R91 - is not evaluating the `*appsv1.Deployment` section (its breaking out without running the part under `*appsv1.StatefulSet`)

annanay25:
@jpkrohling I updated this patch, but there's a bunch of replicated code in the inject package which I'm not sure how to workaround.

annanay25:
> My main problem with the current code is that it won't scale nicely for all types that include a PodSpec/PodTemplateSpec, like batches, jobs, daemon sets, ..

Completely agree. We need a better way to do this.

I think we need more than the `PodTemplateSpec` in the `decorate` function for example, where we're working with the object namespace:

```
// Append the namespace to the app name. Using the DNS style "<app>.<namespace>""
// which also matches with the style used in Istio.
if len(o.Namespace) > 0 {
    app += "." + o.Namespace
} else {
    app += ".default"
}
```

jpkrohling:
> I think we need more than the PodTemplateSpec in the decorate function for example, where we're working with the object namespace:

That's fine, you can just pass the namespace then. Is that all you'd need? PodTemplateSpec + Namespace?

jpkrohling:
I think there are quite a few merge conflicts now. Would you like to fix them, or should I close it?

annanay25:
I'm going to take a stab at this soon :)

annanay25:
@jpkrohling could you tell me if the functions look good? I'll correct the tests accordingly :)

jpkrohling:
Looks like the build is failing:

```
pkg/controller/deployment/deployment_controller.go:90:9: expected switch expression, found assignment (missing parentheses around composite literal?)
```

annanay25:
`runtime.Object` cannot be type casted into `meta/v1.Object` - I'm missing the point here

jpkrohling:
> runtime.Object cannot be type casted into meta/v1.Object - I'm missing the point here

I mean, instead of having almost the same code for `appsv1.Deployment` and `appsv1.StatefulSet`, we could try having `meta/v1.Object` instead. IIRC, we needed something specific for `Deployment` or `Statefulset` in only one line, right?

jpkrohling:
@annanay25, I'm closing this one, as we are currently working on the webhook approach (see #171), which would make this obsolete. If you still want to work on this before we move on with that other approach, feel free to reopen.

PR #98, opened 2018-11-08
Update generated code and reverted change to 'all-in-one' in CRD

jpkrohling:
cc @clyang82 


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/98)

jpkrohling:
> LGTM - shouldn't the duplicate volume/mount names be flagged as errors?

We could -- would you mind opening an issue to discuss it? My current view is that an inner object might want to override an outer object (ie: all components should mount this volume under this name, except for this one component, which should mount that other volume under the same name)

PR #100, opened 2018-11-08
Add OAuth Proxy to UI when on OpenShift

### LINKED ISSUE #89: Protect the UI with OpenShift OAuth Proxy
Depends on #88. When the operator is deployed on OpenShift, the Query UI should be protected by default by the [OpenShift OAuth Proxy](https://github.com/openshift/oauth-proxy).

jpkrohling:
Closes #89.


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/100)

jpkrohling:
This is currently blocked by https://github.com/openshift/oauth-proxy/issues/95.

Unit tests are passing, as well as end-to-end tests, which execute only on Kubernetes. 

Manually testing this on OpenShift, I can see that all the objects are created as required, but I cannot login due to the issue listed above.

So, this is ready for an initial review but shouldn't be merged until the login issue is fixed.

pkg/controller/production_test.go
```diff
@@ -93,11 +97,15 @@ func TestDelegateProductionDepedencies(t *testing.T) {
...
+
 	if hasDaemonSet {
-		assert.Len(t, objs, 6)
-	} else {
-		assert.Len(t, objs, 5)
+		expectedNumObjs++
+	}
+
+	if hasOAuthProxy {
+		expectedNumObjs++
```
objectiser: value of expectedNumObjs is not being asserted

pkg/inject/oauth-proxy.go
```diff
@@ -0,0 +1,58 @@
...
+	"k8s.io/api/core/v1"
+
+	"github.com/jaegertracing/jaeger-operator/pkg/account"
+	"github.com/jaegertracing/jaeger-operator/pkg/apis/io/v1alpha1"
+	"github.com/jaegertracing/jaeger-operator/pkg/service"
+)
+
+// OAuthProxy injects an appropriate OpenShift proxy into the given deployment
+func OAuthProxy(jaeger *v1alpha1.Jaeger, dep *appsv1.Deployment) *appsv1.Deployment {
+	if jaeger.Spec.Ingress.OAuthProxy == nil || !*jaeger.Spec.Ingress.OAuthProxy {
```
objectiser: Shouldn't this also check for platform? All-in-one and production controllers don't check for platform before calling.

pkg/inject/oauth-proxy_test.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func TestOAuthProxyContainerIsNotAddedByDefault(t *testing.T) {
+	jaeger := v1alpha1.NewJaeger("TestOAuthProxyContainerIsNotAddedByDefault")
+	dep := OAuthProxy(jaeger, deployment.NewQuery(jaeger).Get())
+	assert.Len(t, dep.Spec.Template.Spec.Containers, 1)
+	assert.Equal(t, "jaeger-query", dep.Spec.Template.Spec.Containers[0].Name)
+}
+
+func TestOAuthProxyContainerIsAdded(t *testing.T) {
```
objectiser: May need to be updated if platform=openshift check added.

pkg/inject/oauth-proxy_test.go
```diff
@@ -0,0 +1,59 @@
...
+	jaeger := v1alpha1.NewJaeger("TestOAuthProxyContainerIsAdded")
+	b := true
+	jaeger.Spec.Ingress.OAuthProxy = &b
+	dep := OAuthProxy(jaeger, deployment.NewQuery(jaeger).Get())
+	assert.Len(t, dep.Spec.Template.Spec.Containers, 2)
+	assert.Equal(t, "oauth-proxy", dep.Spec.Template.Spec.Containers[1].Name)
+}
+
+func TestOAuthProxyTLSSecretVolumeIsAdded(t *testing.T) {
```
objectiser: Any particular reason for these this test and next assertions not being collapsed into above two tests?

pkg/service/query.go
```diff
@@ -22,6 +24,9 @@ func NewQueryService(jaeger *v1alpha1.Jaeger, selector map[string]string) *v1.Se
 			Name:      GetNameForQueryService(jaeger),
 			Namespace: jaeger.Namespace,
 			Labels:    selector,
+			Annotations: map[string]string{
+				"service.alpha.openshift.io/serving-cert-secret-name": GetTLSSecretNameForQueryService(jaeger),
```
objectiser: As mentioned above, it would be good if things like this could be moved into a platform specific SPI.

pkg/service/query_test.go
```diff
@@ -19,4 +20,21 @@ func TestQueryServiceNameAndPorts(t *testing.T) {
...
+	selector := map[string]string{"app": "myapp", "jaeger": name, "jaeger-component": "query"}
+
+	jaeger := v1alpha1.NewJaeger(name)
+	b := true
+	jaeger.Spec.Ingress.OAuthProxy = &b
+	svc := NewQueryService(jaeger, selector)
+
+	assert.Equal(t, fmt.Sprintf("%s-query", name), svc.ObjectMeta.Name)
+	assert.Len(t, svc.Spec.Ports, 1)
+	assert.Len(t, svc.Spec.Ports, 1)
```
objectiser: duplicate line

pkg/account/oauth-proxy.go
```diff
@@ -0,0 +1,46 @@
...
+	return &v1.ServiceAccount{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "ServiceAccount",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      OAuthProxyAccountNameFor(jaeger),
+			Namespace: jaeger.Namespace,
+			Annotations: map[string]string{
+				"serviceaccounts.openshift.io/oauth-redirectreference.primary": getOAuthRedirectReference(jaeger),
```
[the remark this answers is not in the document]
jpkrohling: I started with that, but it felt weird to have `if openshift` all over the code base. 

The idea behind the current approach is that we'd be deciding about the OAuth Proxy based on the `enabled` flag, and let the controller deal with setting the appropriate flag value based on the platform. The `normalize` operation will set it to false if not on OpenShift, and default to true if on OpenShift (leaving it as is if explicitly set).

pkg/account/oauth_proxy_test.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/jaegertracing/jaeger-operator/pkg/apis/io/v1alpha1"
+)
+
+func TestOAuthRedirectReference(t *testing.T) {
+	b := true
+	jaeger := v1alpha1.NewJaeger("TestOAuthRedirectReference")
+	jaeger.Spec.Ingress.OAuthProxy = &b
+
+	assert.Contains(t, getOAuthRedirectReference(jaeger), jaeger.Name)
+}
```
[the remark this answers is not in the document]
jpkrohling: Not sure what exactly to test there, but I'll add an assertion on the account's name.

pkg/controller/production_test.go
jpkrohling: Nice catch!

pkg/inject/oauth-proxy.go
jpkrohling: The `normalize()` operation will set the flag to `false` if the platform doesn't support the OAuth proxy. I should actually remove the reference to OpenShift from this `godoc`.

pkg/inject/oauth-proxy_test.go
jpkrohling: Not really, just that it's easier to spot what's wrong when the assertion matches the test name.

pkg/service/query.go
jpkrohling: Good catch. I think this will be the only place outside of the `inject` package (and the `normalize`) that we have OpenShift-specific knowledge for the OAuth Proxy (the other feature being Routes vs. Ingresses). For now, I'll just add a check on the OAuth Proxy flag and wait for a second implementation before extracting a provider interface.

deploy/examples/disable-oauth-proxy.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: io.jaegertracing/v1alpha1
```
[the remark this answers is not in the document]
objectiser: +1

pkg/account/oauth-proxy.go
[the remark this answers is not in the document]
objectiser: Just thinking about future platforms - might be better to define the structure now to make it easier for other platforms to be added - as otherwise it could be quite messy to try to decouple the binary decisions between k8s and openshift at a later date.

pkg/inject/oauth-proxy_test.go
objectiser: Fair enough :)

README.adoc
```diff
@@ -176,6 +176,19 @@ oc get routes
...
+
+[source,yaml]
+----
+apiVersion: io.jaegertracing/v1alpha1
+kind: Jaeger
+metadata:
+  name: disable-oauth-proxy
+spec:
+  ingress:
+    oauth-proxy: false
```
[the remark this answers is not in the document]
objectiser: Yes that would be good

objectiser:
Refactoring to support a platform SPI will be handled in a separate PR.

PR #126, opened 2018-11-23
Remove vendor from scm

jpkrohling:


jpkrohling:
The Operator SDK [adds the `vendor` to the SCM by default](https://github.com/operator-framework/operator-sdk-samples/tree/master/memcached-operator/vendor), but this is not something we usually do with our repositories.

A big downside was shown during the update of the Operator SDK, with the diff being mostly about changes in the vendor directory.

jpkrohling:
As discussed with @objectiser via IRC, there's no clear winner about checking in the `vendor`. So, we'll keep it as is for now.

Issue #142, opened 2018-11-30
with-cassandra example is broken


pavolloffay:
Example `with-cassandra` is broken. It deploys all components successfully but c* storage is misconfigured and jaeger UI responds with something like `Could not achieve consistency LOCAL_ONE`.

pavolloffay:
done in #145

PR #145, opened 2018-11-30
Smoke test

### LINKED ISSUE #143: Add e2e tests which create spans and query jaeger-query
The current itegration tests are not bulletproof. They verify that deployment was successful, but that is not enough as the storage might return an error on read/write when it's not properly configured.

pavolloffay:
Resolves #143 #142 

Use port forwad to query and collector. Report a span to collector and query it.

This proves that cassandra tests incorrectly configures c* resulting in error on reads and writes.

 review by commits - checked in vendor dir...


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/145)

pavolloffay:
> Could you add a call to the SmokeTest from a test that is known to work? This is just to "prove" that the SmokeTest is working.

The PR fixes that test and also C* example.

pavolloffay:
> The test did not pass for me, though:

did you run `dep ensure`? 

@objectiser could you please also try? On my machine it passes fine (I know we like claims like these)

pavolloffay:
>Is it needed? If so, then the vendor changes should be included in this PR.

The vendor changes are checked in. What else is needed except adding it to `Gopkg` and including vendor sources?

pavolloffay:
indeed there was an issue with the test. Also note that it's using port forwarding on standard jaeger ports so make sure those ports are free: 16686, 14268. 

k8s client 1.13.0 will allow us to use random ports. (It allows now but there is no getter on the forwarded port). I could not update that dep now.

PR #155, opened 2018-12-04
Add Elasticsearch index cleaner as cron job

pavolloffay:
CronJob for https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/esCleaner.py#L11


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/155)

test/elasticsearch.yml
```diff
@@ -16,27 +16,21 @@ spec:
     spec:
       containers:
         - name: elasticsearch
-          image: docker.elastic.co/elasticsearch/elasticsearch:5.6.0
+          image: docker.elastic.co/elasticsearch/elasticsearch:5.6.12
```
pavolloffay: I have bumped patch version. And disabled xpack, we should not use it anyways. There an official ES6x without licensed xpack, but I would rather keep using the lowest supported version.

pkg/apis/io/v1alpha1/jaeger_types.go
```diff
@@ -150,6 +151,14 @@ type JaegerDependenciesSpec struct {
 	ElasticsearchNodesWanOnly   bool   `json:"elasticsearchNodesWanOnly"`
 }
 
+// JaegerEsIndexCleanerSpec holds the options related to es-index-cleaner
+type JaegerEsIndexCleanerSpec struct {
+	Enabled      bool   `json:"enabled"`
+	NumberOfDays int    `json:"numberOfDays"`
```
pavolloffay: Used 7 days as default. TTL in c* is 2 day though...

pkg/apis/io/v1alpha1/jaeger_types.go
```diff
@@ -150,6 +151,14 @@ type JaegerDependenciesSpec struct {
 	ElasticsearchNodesWanOnly   bool   `json:"elasticsearchNodesWanOnly"`
 }
 
+// JaegerEsIndexCleanerSpec holds the options related to es-index-cleaner
+type JaegerEsIndexCleanerSpec struct {
+	Enabled      bool   `json:"enabled"`
```
pavolloffay: By default disabled. 

I think we should enable it by default along with dependencies job. I would rather do it in a separate PR to keep this minimal.

pkg/apis/io/v1alpha1/jaeger_types.go
pavolloffay: The script supports other configuration properties, see https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/es/esCleaner.py#L11

index prefix, es hostname, I have not exposed them directly but they are derived from storage properties map.

PR #161, opened 2018-12-06
Fix log when spak depenencies are used with unsupported storage

### LINKED ISSUE #157: Log when dependency job is enabled with unsupported storage
Follow-up from #140: that PR adds a `log.Info` on the default case (Enabled=false for Cassandra and ES), while it doesn't let the user know that the dependency job isn't installed when Enabled=true and storage isn't Cassandra or ES. 

It should:

1. Not log anything on the standard, common case. If everything is as it should, there's no need to log anything
1. Log (debug or info) when Enabled=true and the storage isn't ES or Cassandra. This is because the flag is `false` by default, meaning that the user has explicitly set the value to `true`, so, there's a clear expectation that the job will be installed.

pavolloffay:
Resolves #157 

### DIFF
--- a/pkg/strategy/all-in-one.go
+++ b/pkg/strategy/all-in-one.go
@@ -79,11 +79,11 @@ func (c *allInOneStrategy) Create() []runtime.Object {
 		}
 	}
 
-	if cronjob.SupportedStorage(c.jaeger.Spec.Storage.Type) {
-		if c.jaeger.Spec.Storage.SparkDependencies.Enabled {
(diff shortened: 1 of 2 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/161)

pavolloffay:
Changed, I will merge this as there will be conflicts in #162 and I won't be available. I will fix any outstanding issues in follow pr.

PR #172, opened 2019-01-18
WIP - Use recommended labels

jpkrohling:

### DIFF
--- a/pkg/deployment/collector.go
+++ b/pkg/deployment/collector.go
@@ -39,7 +39,7 @@ func NewCollector(jaeger *v1alpha1.Jaeger) *Collector {
 func (c *Collector) Get() *appsv1.Deployment {
 	logrus.Debugf("Assembling a collector deployment for %v", c.jaeger)
 
-	selector := c.selector()
+	labels := c.labels()
 	trueVar := true
 	replicas := int32(c.jaeger.Spec.Collector.Size)
 
@@ -75,7 +75,7 @@ func (c *Collector) Get() *appsv1.Deployment {
 			Kind:       "Deployment",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-collector", c.jaeger.Name),
+			Name:      c.name(),
 			Namespace: c.jaeger.Namespace,
 			OwnerReferences: []metav1.OwnerReference{
 				metav1.OwnerReference{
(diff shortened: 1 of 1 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/172)

pkg/deployment/collector.go
```diff
@@ -148,12 +148,26 @@ func (c *Collector) Get() *appsv1.Deployment {
...
+		service.NewCollectorService(c.jaeger, c.labels()),
 	}
 }
 
-func (c *Collector) selector() map[string]string {
-	return map[string]string{"app": "jaeger", "jaeger": c.jaeger.Name, "jaeger-component": "collector"}
+func (c *Collector) labels() map[string]string {
+	return map[string]string{
+		"app.kubernetes.io/name":       c.name(),
+		"app.kubernetes.io/instance":   c.jaeger.Name,
```
objectiser: Looks reasonable - only one I wasn't sure about was the `instance`, where the [docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) describe it as "A unique name identifying the instance of an application" with an example of `wordpress-abcxzy`. But for our purpose it enables objects to be selected by various criteria.

jpkrohling: I'll apply the same change to the other objects then. Thanks!

objectiser: Don't forget we will also need to support the `app` label for a while - was thinking that should be same as `app.kubernetes.io/name` value?

jpkrohling: Do you have existing applications in mind, or is there a case of an application that still reads the `app` instead of `app.kubernetes.io`? If it's about maintaining backwards compatibility, we need to keep `app: jaeger`.

objectiser: I guess the two areas to consider is:
1) Istio - which needs the `app` label and `app:jaeger` should be fine
2) Kiali - raised by Jay? where unique `app` label seemed to be a problem - but possibly this is not a blocker

jpkrohling: If they are also moving to use `app.kubernetes.io` instead of `app`, then I'd keep `app: jaeger` for backwards compatibility.

objectiser: Unclear what the timeframe is for them changing, but yes probably the best approach.

jpkrohling:
This is now ready to be reviewed.

jpkrohling:
Tests added

objectiser:
Thanks :+1:

PR #184, opened 2019-01-29
WIP - Implemented updates to replica size

jpkrohling:


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/184)

jpkrohling:
This is the first draft for the "update CR" story. I tried a couple of different approaches, but in the end, I think having specialized methods with knowledge about the strategies is the best way to move forward.

If we can agree on the general approach taken in this PR, I'll add tests (unit, e2e) and documentation.

jpkrohling:
To try this out:

```
$ kubectl apply -f deploy/examples/two-replicas.yaml 
jaeger.io.jaegertracing/two-replicas created

$ kubectl get pods
...
two-replicas-fb766d855-6g589   1/1       Running   0          14s
two-replicas-fb766d855-7qcz9   1/1       Running   0          14s

$ kubectl get deployments
NAME           DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
two-replicas   2         2         2            2           53s

$ kubectl edit jaeger two-replicas ## change allInOne size to 3
jaeger.io.jaegertracing/two-replicas edited

$ kubectl get pods
NAME                           READY     STATUS    RESTARTS   AGE
...
two-replicas-fb766d855-67h6v   1/1       Running   0          29s
two-replicas-fb766d855-6g589   1/1       Running   0          1m
two-replicas-fb766d855-7qcz9   1/1       Running   0          1m

$ kubectl get deployments
NAME           DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
two-replicas   3         3         3            3           2m

$ kubectl edit jaeger two-replicas ## change it back to 2 replicas
jaeger.io.jaegertracing/two-replicas edited

$ kubectl get pods
NAME                           READY     STATUS    RESTARTS   AGE
...
two-replicas-fb766d855-6g589   1/1       Running   0          2m
two-replicas-fb766d855-7qcz9   1/1       Running   0          2m

$ kubectl get deployments
NAME           DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
two-replicas   2         2         2            2           2m
```

jpkrohling:
> For now, to support this specific update case (i.e. replicas), this is fine. Couple of points:

The idea is to evolve it to support other update types. Just updating the Spec would also work in a similar way, but a migration scenario would be placed in a similar location, but with a more complex logic.

>     1. I believe the new example has been added simply to enable testing against all-in-one - but would prefer not to have an extra example file just for replica.

Alright.

>     2. Wondering if the update should still be strategy specific - doing the size change on components that are not part of the strategy seems wrong. What is the impact of it testing the size of collector, ingester and query when those deployments don't exist?

When the strategy is `allInOne`, the `Size` in the collector/ingester/query specs is expected to be `nil`. This value is checked as the first thing in the `updateReplicaSizeDeployment` func, returning if it's `nil`. We can certainly do some extra sanity checks here.

I think the update should not be strategy-specific, but perhaps strategy-aware. This is because we'll have cases in the future where the update involves switching from one strategy to another.

There's an edge case where someone sets the strategy to `allInOne` and specifies a `query` YAML node with a size value (any value). I *expect* the `r.client.Get()` call to then fail with a `non existing object` message, but that's something to test in the unit tests.

objectiser:
I think that is what I was thinking - as it is quite valid for a user to have a CR with all components defined - but the strategy essentially focuses on only the ones that are relevant. Not sure we should be reporting errors just because the same CR is then applied for update, where the all-in-one is only actually used.

jpkrohling:
> Not sure we should be reporting errors just because the same CR is then applied for update, where the all-in-one is only actually used.

Absolutely, we can handle this case, either by ignoring the error if it's a simple "not found" (as we do in the main reconcile logic already), or by checking the strategy. I would prefer *not* to do it based on the strategy, mainly because the logic behind the strategy chooser is somewhere else down the code.

jpkrohling:
I'm closing this one, as we'll have the replica size change as part of the generic update approach.

PR #191, opened 2019-02-01
Use es-operator

### LINKED ISSUE #183: Use elasticsearch-operator for ES deployment
We would like to use https://github.com/openshift/elasticsearch-operator to manage ES deployment. This should be optional and users can still bring managed ES instances. The elasticsearch-operator uses this ES image https://github.com/openshift/origin-aggregated-logging/tree/master/elasticsearch.

There are a couple of problems with the eslasticsearch-operator. It's openshift-common-logging specific (uses their image), however we will make it more generic to be used with any APP and vanilla ES image.

The following steps are required to make the es-operator work with jaeger operator:
* Allow es-operator to watch every namespace https://github.com/openshift/elasticsearch-operator/pull/74 
* add searchguard rules to ECL Elasticsearch image https://github.com/openshift/origin-aggregated-logging/pull/1500/files
* create `Role` and `RoleBinding` when es-operator is being used (follow openshift-ansible/roles/openshift_logging_elasticsearch/templates/loggin-jaeger-rolebinding.ym)
* create certificates https://github.com/openshift/cluster-logging-operator/blob/master/scripts/cert_generation.sh
* OLM deploys es-operator and jaeger operator, jaeger operator just creates es CR
* pass /var/run/secrets/kubernetes.io/serviceaccount/token bearer token to jaeger (it's my service accout)
* mount certificates to jaeger

```
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: jaeger-data-manager
+  namespace: {{ namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: jaeger-data-manager
+subjects:
+  - kind: ServiceAccount
+    namespace: {{ role_namespace }}
+    name: {{ role_user }}


+---
kind: Role
metadata:
  annotations:
    rbac.authorization.kubernetes.io/autoupdate: "true"
  name: jaeger-data-manager
  namespace: {{ namespace }}
rules:
  - apiGroups:
      - metrics.openshift.io
    resources:
      - jaeger
    verbs:
      - edit or manage?
```

pavolloffay:
Resolves #183 

depends on: https://github.com/openshift/origin-aggregated-logging/pull/1500

- [x] create ES CR inside the operator
- [x] use client certs
- [x] create separate secret for jaeger
- [ ] integration tests

The elasticsearch has to be deployed manually at the moment
https://github.com/openshift/elasticsearch-operator, The ES image has to with this change https://github.com/openshift/origin-aggregated-logging/pull/1500
```
make deploy
oc create -f hack/cr.yaml - not necessary Jaeger operator deploys CR
```
If the es deployment has been created before jaeger deployement the ES pod has to be restarted so it can load correct certificates - actually it waits for volume with certificates, but needs to be restarted if certs are changed - jaeger redeploy. We might create a different volume with certs per jaeger deployment.

Note that ES requires limits settings on node:
`minishift ssh -- 'sudo sysctl -w vm.max_map_count=262144'`


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/191)

pavolloffay:
I will keep follow up issues in this comment

* use separate secret for jaeger client certs - now all certs are in one secret
* allow multiple jaeger instances access the same elasticsearch storage
* use certificates with index-cleaner and spark-dependencies
* deploy ES by creating CR and configuring jaeger storage flags and mounting volumes

pkg/strategy/production.go
```diff
@@ -102,6 +102,20 @@ func (c *productionStrategy) Create() []runtime.Object {
 		}
 	}
 
+	if strings.ToLower(c.jaeger.Spec.Storage.Type) == "elasticsearch" {
+		// TODO query and collector will be failing until ES is up - we could wait on readiness probe
+		// TODO we could create ES CR if no es.server-urls are provided (and ES CRD is defined)
+		err := storage.CreateESCerts()
```
pavolloffay: @jpkrohling @objectiser 

We should decide who and under what conditions the ES instance will be created. 

I would like to create ES CR (assuming es-operator is deployed by OLM and will create ES instance) if:
* flag `es.server-urls`  is not defined 
* there is ES CRD (optional)

so I will create ES here and configure our components to talk to it - mount secrets and use appropriate flags.

objectiser: Sounds reasonable to me, as to use an existing ES cluster they would need to provide the URL.

pavolloffay:
We should follow up on:

- [ ] all certificates (secrets) are created when ES deployment is created. I have noticed that common-logging operator first checks if there is a secret and if yes it gets the cert from there.
- [ ] The ES spec is embedded directly in storage spec. We should consider creating `storage.elasticsarch` and `storage.cassandra` - I would prefer to do it in a separate PR.

pavolloffay:
It's just a log print if it makes problems we can remove. 

e2e tests passed
```
 ploffay  ~/projects/golang/src/github.com/jaegertracing/jaeger-operator   es-operator  NAMESPACE=pavolloffay make e2e-tests                                                       2:07 
Formatting code...
Building...
Sending build context to Docker daemon  84.97MB
Step 1/6 : FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 ---> 5f269bf34c29
 ---> Using cache
 ---> 92cd24463204
 ---> Using cache
 ---> e680be3c29b0
Step 4/6 : USER nobody
 ---> Using cache
 ---> 5283fd5b7f41
 ---> 0be97d0b9a48
 ---> Running in 51f4860e97d5
Removing intermediate container 51f4860e97d5
 ---> 8f80cd0c0086
Successfully built 8f80cd0c0086
Successfully tagged pavolloffay/jaeger-operator:latest
Pushing image pavolloffay/jaeger-operator:latest...
Running end-to-end tests...
ok  	github.com/jaegertracing/jaeger-operator/test/e2e	261.688s
```

PR #213, opened 2019-02-21
Removed constraint on namespace when obtaining available Jaeger instances

### LINKED ISSUE #60: Inject sidecars from different namespaces
It's quite possible that Jaeger instances will live in one namespace and applications in another. We need a way then to inject sidecars into deployments from one namespace pointing to Jaeger instances in another namespace.

Tasks:
- [x] Figure out whether the `JaegerSpec` should have a list of namespaces the operator should watch
- [x] The RBAC implications: what are the failure scenarios and how to recover from that
- [ ] On OpenShift, one namespace cannot see things from another namespace by default. What should we do here?

jpkrohling:
Fixes #60 by removing the namespace constraint when querying for available Jaegers, as Kubernetes will do the right thing and deliver only the ones that the Operator's account can see.

@gregoryfranklin: do you think you'd be able to try this one out?

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -298,6 +298,8 @@ spec:
 ----
 <1> Either `"true"` (as string) or the Jaeger instance name
 
+A complete sample deployment is available at link:./deploy/examples/business-application-injected-sidecar.yaml[`deploy/examples/business-application-injected-sidecar.yaml`]
+
 == Agent as DaemonSet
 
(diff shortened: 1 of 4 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/213)

jpkrohling:
@pavolloffay could you please review this one?

jpkrohling:
Done, thanks!

PR #217, opened 2019-02-21
Added capability of detecting the platform

### LINKED ISSUE #133: Automatically detect platform
At the moment jaeger-operator has to be configured with `--platform=openshift` to find out information about the platform: k8s, openshift. This leads to duplication of deployment files and docs. 

Maybe we can get the platform from the orchestration API and adjust behavior based on that.

jpkrohling:
Fixes #133 by implementing a way of detecting when we are running in OpenShift. If a "route" resource cannot be found, we assume we are running in Kubernetes.

When running in OpenShift, this is th expected output:
```
$ make run
customresourcedefinition.apiextensions.k8s.io/jaegers.io.jaegertracing created
INFO[0000] Go                                            arch=amd64 os=linux version=go1.11.1
INFO[0000] operator-sdk                                  version=v0.4.1
DEBU[0000] Attempting to auto-detect the platform       
```

And this is the expected output for Kubernetes:
```
$ make run
customresourcedefinition.apiextensions.k8s.io/jaegers.io.jaegertracing created
INFO[0000] Go                                            arch=amd64 os=linux version=go1.11.1
INFO[0000] operator-sdk                                  version=v0.4.1
DEBU[0000] Attempting to auto-detect the platform       

```

This PR also removes the content about `--platform=openshift`, which is still available to override the auto-detect routine.


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/217)

README.adoc
```diff
@@ -44,7 +44,7 @@ The operator is now ready to create Jaeger instances!
 
 === OpenShift
 
-The instructions from the previous section also work on OpenShift given that the `operator-openshift.yaml` is used instead of `operator.yaml`. Make sure to install the RBAC rules, the CRD and the operator as a privileged user, such as `system:admin`.
+The instructions from the previous section also work on OpenShift. Make sure to install the RBAC rules, the CRD and the operator as a privileged user, such as `system:admin`.
```
pavolloffay: Could be the whole section removed? Instead have just one deploy section. Less text to maintain and allows us  catching differences between platforms sooner.

pkg/cmd/start/main.go
```diff
@@ -57,7 +60,7 @@ func NewStartCommand() *cobra.Command {
 	cmd.Flags().String("openshift-oauth-proxy-image", "openshift/oauth-proxy:latest", "The Docker image location definition for the OpenShift OAuth Proxy")
 	viper.BindPFlag("openshift-oauth-proxy-image", cmd.Flags().Lookup("openshift-oauth-proxy-image"))
 
-	cmd.Flags().String("platform", "kubernetes", "The target platform the operator will run. Possible values: 'kubernetes' and 'openshift'")
+	cmd.Flags().String("platform", "auto-detect", "The target platform the operator will run. Possible values: 'kubernetes' and 'openshift'")
```
pavolloffay: do we need to keep this flag at all? I would rather remove it altogether and rely on auto detection.

pavolloffay: We can add it once we need it.

jpkrohling: If our auto-detect logic fails for some reason in some strange situations, it's good to provide a fallback to the user. So, let's keep it.

README.adoc
jpkrohling: You mean, the whole OpenShift section?

pavolloffay: yes

jpkrohling: I'll open an issue to discuss it. I think it still has value, especially about logging as admin to create the objects that can't be created with an ordinary user.

pavolloffay: There can be a sentence saying log as admin and use oc instateak of kubectl

jpkrohling: As far as I know, `kubectl` can be used instead of `oc` for OpenShift as it will cleanly translate to the appropriate commands. I opened an issue (#222) to discuss this further. I'm OK in removing it if the Kubernetes steps can be followed for OpenShift as well.

PR #229, opened 2019-02-25
Set shards and replicas based on redundancy policy

pavolloffay:
https://issues.jboss.org/browse/TRACING-386

### DIFF
--- a/pkg/storage/elasticsearch.go
+++ b/pkg/storage/elasticsearch.go
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"fmt"
 	"strings"
 
 	"k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/229)

pavolloffay:
I have added more tests and slightly changed ES CR. If node count is higher than 3 (N) it creates CR with 3 master nodes and N-3 client,data nodes. 

It's not recommended to deploy more than 3 master nodes in general and ESO even does not allow it:
```
time="2019-02-25T15:55:43Z" level=error msg="error syncing key (myproject/elasticsearch): Failed to reconcile Elasticsearch deployment spec: Invalid master nodes count. Please ensure there are no more than 3 total nodes with master roles"
```

pavolloffay:
At the moment there is a bug in ES operator, it does not create deployment correctly for 3 master and N other nodes https://jira.coreos.com/browse/LOG-340

pavolloffay:
@jpkrohling could you please re-review and/or merge?

pavolloffay:
> How do you even know this exists :D

I heard old dudes talking about it... I thought that I will give it a try on k8s!

> Could you add a reference to the bug? Is there a public place where the bug can be seen? I tried opening the JIRA, but I don't seem to have access to it O_o

It's private jira, once it's fixed we don't have to change anything as it is the correct topology. I have just added a comment here to let people know.

PR #173, opened 2019-01-21
Add Jaeger CSV and Package for OLM integration and deployment of the …

awgreene:
…operator through OperatorHub

- This pull requests introduces a CSV and Package which can be bundled together and used by the [Operator Lifecycle Manager](https://github.com/operator-framework/operator-lifecycle-manager) to install, manage, and upgrade the Jaeger Operator in a cluster.

- The Jaeger operator versions available to all Kubernetes clusters using OLM can be updated by submitting a pull request to the [Community Operators GitHub repo](https://github.com/operator-framework/community-operators) that includes the latest CSVs, CRDs, and Packages.

- The CSV was created based of the [documentation provided by OLM](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Documentation/design/building-your-csv.md) and the [documentation](https://github.com/operator-framework/community-operators/blob/master/docs/marketplace-required-csv-annotations.md) provided by [OperatorHub](https://github.com/operator-framework/operator-marketplace).

- The operator and the CSV were tested using the [scorecard functionality recently introduced](https://github.com/operator-framework/operator-sdk/pull/758) to the [Operator-sdk](https://github.com/operator-framework/operator-sdk). Currently, the Jaeger operator scores a 4/18 on the scorecard tests.

OLM integration could be improved by future code changes to the Operator:
- Addressing the operator-sdk scorecard tests that failed.
- Adding additional information to the Jaeger CSV based on the [CSV documentation provided by OLM](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Documentation/design/building-your-csv.md)
- Adding additional information to the Jaeger CSV based on the [CSV documentation provided by OperatorHub](https://github.com/operator-framework/community-operators/blob/master/docs/marketplace-required-csv-annotations.md)

PR #247, opened 2019-03-01
Fixed the bump of release in deploy artifacts

jpkrohling:
Since #173, we have a few more files to bump the version during the releases.
Since #217, we don't have the `operator-openshift.yaml` file anymore, causing the `release.sh` [to fail](https://travis-ci.org/jaegertracing/jaeger-operator/jobs/499756950#L484).

This PR fixes both scenarios and bumps the necessary versions to the latest ones.


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/247)

jpkrohling:
@awgreene -- I realize that I should use `operator-sdk olm-catalog gen-csv --csv-version` instead of messing with the files directly, but the Operator SDK version we are using (0.4.0) doesn't support that command yet. Right now, it's too risky for us to use any other version of the SDK.

PR #249, opened 2019-03-01
Store back the CR only if it has changed

### LINKED ISSUE #231: Reconcile loop called every second on OpenShift, but not on Kubernetes
When running in Kubernetes, the reconcile loop is being called only when the CR is changed, while in OpenShift it's being called every second. 

https://groups.google.com/d/msg/operator-framework/DZynMacOgoY/RWaSFxFUBwAJ

#203 might be caused by this.

jpkrohling:
Fixes #231 by comparing the original CR with the post-reconcile state, issuing an update only in case the CR has changed.

### DIFF
--- a/pkg/controller/jaeger/jaeger_controller.go
+++ b/pkg/controller/jaeger/jaeger_controller.go
@@ -3,6 +3,7 @@ package jaeger
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/249)

pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -109,16 +111,13 @@ func (r *ReconcileJaeger) Reconcile(request reconcile.Request) (reconcile.Result
 		return reconcile.Result{}, err
 	}
 
-	logFields.Info("Configured Jaeger instance")
```
pavolloffay: Do you want to remove it? We could indicate the changes on line 119

jpkrohling: Yeah, I wanted to remove it. As someone managing the cluster, I'm not sure this would be useful at all in the day-to-day. It's certainly useful to get a positive confirmation in the beginning, but considering that there will be one change every 5 seconds (due to the metrics on #248), this would cause too much noise in the logs.

PR #257, opened 2019-03-04
Use only ES data nodes to calculate shards

### LINKED ISSUE #253: Change the number of shards and replicas when ES is scaled up/down
We derive the number of shards and replicas based on the number of ES nodes deployed. These numbers are then injected to query and collector pods specs.

So if the number of ES nodes changes should we change the number of shards?  Note that the number of shards can be changed only via reindex API https://discuss.elastic.co/t/how-change-the-number-of-shards-and-replicas-in-elasticsearch/84736. The number of replicas can be changed via index mappings API. However as we use daily indices we can just wait for the new index to be created with updated settings.

* once we change number of ES nodes we should probably restart query and collector with updated es flags. Then new daily indices will have properly set number of shards on replicas.
* Should be the number of shards and replicas based only on the number of data nodes? Now we use the total number of nodes.

pavolloffay:
Resolves #253 


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/257)

jpkrohling:
Needs `make format`. Other than that, LGTM

PR #267, opened 2019-03-05
Add rollover support

pavolloffay:

This at the moment does not work with self-provisioned ES due to curator settings: https://github.com/openshift/origin-aggregated-logging/pull/1517


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/267)

jpkrohling:
This will conflict with #265. Please wait to work on this, let's get that one merged first.

pavolloffay:
Test CR
```
# setup an elasticsearch with `make es`
kind: Jaeger
metadata:
  name: simple-prod
spec:
  strategy: production
  storage:
    type: elasticsearch
    options:
      es:
        use-aliases: true
```

Self-provisioned ES does not at the moment - my PR with SG config has been merged but the image is not yet available.

pavolloffay:
this is ready for re-review

pavolloffay:
> Why don't you just use Seconds()? time.ParseDuration("1m30s").Seconds() returns 90, so, you can just use something like return pythonUnits{units: seconds, count: int(time.ParseDuration("1m30s").Seconds())}

I am with what is there now, it's easier to read/debug.

pavolloffay:
> .WithField("readTTL", jaeger.Spec.Storage.Rollover.ReadTTL) so that the user knows what we attempted to parse.

the value is present I have modified the logging to produce nicer logs :)

PR #279, opened 2019-03-07
Block until objects have been created and are ready

### LINKED ISSUE #277: Block until objects have been created and are ready
In some parts of the controller, during the reconciliation logic, we check for existing objects, such as deployments. A timing condition might cause the check for the state of a object to fail, as the object might not have been created yet. Ideally, we would wait until the object has been created *and* available, instead of just available.

https://github.com/jaegertracing/jaeger-operator/blob/898bb4312ff74f829f01b1c1c0440a722aaf90bc/pkg/controller/jaeger/deployment.go#L48-L52

https://github.com/jaegertracing/jaeger-operator/blob/898bb4312ff74f829f01b1c1c0440a722aaf90bc/pkg/controller/jaeger/deployment.go#L74-L76

jpkrohling:
Closes #277 

### DIFF
--- a/pkg/controller/jaeger/elasticsearch.go
+++ b/pkg/controller/jaeger/elasticsearch.go
@@ -6,7 +6,9 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/apps/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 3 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/279)

jpkrohling:
I tried to come up with a way to test this, but I failed, as we need to reliably execute some code at exactly between the "create" command and the "list" in the "waitFor...". I'm afraid this is one of those cases where it's not easy/feasible to test.

Suggestions?

pkg/controller/jaeger/elasticsearch.go
```diff
@@ -61,6 +63,14 @@ func waitForAvailableElastic(c client.Client, es esv1alpha1.Elasticsearch) error
 	return wait.PollImmediate(time.Second, 2*time.Minute, func() (done bool, err error) {
 		depList := corev1.DeploymentList{}
 		if err = c.List(context.Background(), client.MatchingLabels(es.Labels).InNamespace(es.Namespace), &depList); err != nil {
+			if k8serrors.IsNotFound(err) {
```
pavolloffay: Did you run into this issue in general for any deployment? Interestingly I haven't see this for ES and the controller should reach this block before the deployment is found

jpkrohling: I did *not* run into this problem with ES, but in theory, it could also happen there. I guess it happens more often with deployments because we have two of them (query/collector)?

PR #288, opened 2019-03-08
WIP: Use random port forward

### LINKED ISSUE #154: Use random ports in port forward and smoketests
We use port forward in smoke tests introduced in https://github.com/jaegertracing/jaeger-operator/pull/145.

The test forwards collector and query ports on host machine. However it uses standard ports (e.g. 16686) so the test might fail if there is jaeger server running locally.

Forwarder allows to use random port if it is specified as `:16686` or `0:16686`, however random ports are not exposed in the currently used version of go-client. Version 1.13.0 added API to get used ports.

pavolloffay:
Resolves #154 

Still WIP there was a bug in k8s which does not change the port https://github.com/kubernetes/client-go/blob/kubernetes-1.13.4/tools/portforward/portforward.go#L201

It is fixed in current master https://github.com/kubernetes/client-go/blob/master/tools/portforward/portforward.go#L209 1.15.0+ version

### DIFF
--- a/test/e2e/cassandra.go
+++ b/test/e2e/cassandra.go
@@ -75,7 +75,7 @@ func cassandraTest(t *testing.T, f *framework.Framework, ctx *framework.TestCtx)
 	if err != nil {
 		return err
 	}
-	//defer portForw.Close()
+	defer portForw.Close()
 	defer close(closeChan)
 	ports, err := portForw.GetPorts()
 	if err != nil {
--- a/test/e2e/cassandra.go
+++ b/test/e2e/cassandra.go
@@ -71,11 +71,15 @@ func cassandraTest(t *testing.T, f *framework.Framework, ctx *framework.TestCtx)
 	if err != nil {
 		return err
 	}
-	portForw, closeChan, err := CreatePortForward(namespace, jaegerPod.Name, []string{"16686", "14268"}, f.KubeConfig)
+	portForw, closeChan, err := CreatePortForward(namespace, jaegerPod.Name, []string{"0:16686", "0:14268"}, f.KubeConfig)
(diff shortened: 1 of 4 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/288)

pavolloffay:
@jpkrohling do you think we can update k8s API to 1.15? Or do we need to depend on 1.13 because of operator SDK?

jpkrohling:
Might be better to ask on the Operator SDK mailing list or forums, but I see that the they pinned the version to 1.13.4 for their master, even after releasing SDK 0.8.2/0.9.0. I guess that's a sign that something might not be working on newer versions :-)

https://github.com/operator-framework/operator-sdk/blob/master/go.mod#L113

Given that this is for testing purposes, let's keep the version brought by the Operator SDK.

jpkrohling:
Should we close this one?

kevinearls:
@jpkrohling I implemented use of random ports in tests with https://github.com/jaegertracing/jaeger-operator/pull/545

pavolloffay:
Not quite, we should rely on the random ports provided by k8s client. 

This is blocked until we update to k8s api 1.15.0+

jpkrohling:
Any news on this one, or should it be closed?

pavolloffay:
what version of the k8s are we using?

jpkrohling:
https://github.com/jaegertracing/jaeger-operator/blob/ec441bd0fcc6226686f93d5eb4970229a20ca415/go.mod#L33-L55

pavolloffay:
done in #862

PR #303, opened 2019-03-13
Enable archive button if archive storage is enabled

pavolloffay:


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/303)

pkg/apis/jaegertracing/v1/freeform.go
```diff
@@ -39,3 +39,12 @@ func (o FreeForm) MarshalJSON() ([]byte, error) {
 func (o FreeForm) IsEmpty() bool {
 	return len(o.json) == 0 || string(o.json) == "{}"
 }
+
+// GetMap returns a map created from json
+func (o FreeForm) GetMap() (map[string]interface{}, error) {
```
pavolloffay: Just a note...

I was not able to add `map[string]interface{}` to spec because deepcopy does not support it

objectiser: Ah yes - that rings a bell.

PR #311, opened 2019-03-13
Automatically Enable/disable depenencies tab

### LINKED ISSUE #163: Enable/disable dependencies tab in UI when dependency job is enabled/disabled

pavolloffay:
Resolves #163

### DIFF
--- a/pkg/strategy/controller.go
+++ b/pkg/strategy/controller.go
@@ -141,21 +141,42 @@ func normalizeUI(spec *v1.JaegerSpec) {
 			uiOpts = m
 		}
 	}
-	sOpts := spec.Storage.Options.Map()
-	// we respect explicit UI config
+	enableArchiveButton(uiOpts, spec.Storage.Options.Map())
+	disableDependenciesTab(uiOpts, spec.Storage.Type, spec.Storage.SparkDependencies.Enabled)
+	if len(uiOpts) > 0 {
+		spec.UI.Options = v1.NewFreeForm(uiOpts)
+	}
+}
+
+func enableArchiveButton(uiOpts map[string]interface{}, sOpts map[string]string) {
+	// respect explicit settings
 	if _, ok := uiOpts["archiveEnabled"]; !ok {
+		// archive tab is by default disabled
 		if strings.EqualFold(sOpts["es-archive.enabled"], "true") ||
 			strings.EqualFold(sOpts["cassandra-archive.enabled"], "true") {
 			uiOpts["archiveEnabled"] = true
 		}
 	}
+}
 
-	if _, ok := uiOpts["dependencies"]; !ok {
-		
+func disableDependenciesTab(uiOpts map[string]interface{}, storage string, depsEnabled *bool) {
+	// dependency tab is by default enabled and memory storage support it
+	if strings.EqualFold(storage, "memory") || (depsEnabled != nil && *depsEnabled == false) {
+		return
 	}
-
-	if len(uiOpts) > 0 {
-		spec.UI.Options = v1.NewFreeForm(uiOpts)
+	deps := map[string]interface{}{}
+	if val, ok := uiOpts["dependencies"]; ok {
+		if val, ok := val.(map[string]interface{}); ok {
+			deps = val
+		} else {
+			// we return as the type does not match
+			return
+		}
+	}
+	// respect explicit settings
+	if _, ok := deps["menuEnabled"]; !ok {
+		deps["menuEnabled"] = false
+		uiOpts["dependencies"] = deps
 	}
 }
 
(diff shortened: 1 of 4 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/311)

pkg/strategy/controller_test.go
```diff
@@ -246,38 +246,94 @@ func TestNormalizeUI(t *testing.T) {
...
-			j:        &v1.JaegerSpec{Storage: v1.JaegerStorageSpec{Options: v1.NewOptions(map[string]interface{}{"es.archive.enabled": "false"})}},
-			expected: &v1.JaegerSpec{Storage: v1.JaegerStorageSpec{Options: v1.NewOptions(map[string]interface{}{"es.archive.enabled": "false"})}},
+			j:        &v1.JaegerSpec{Storage: v1.JaegerStorageSpec{Type: "memory"}},
+			expected: &v1.JaegerSpec{Storage: v1.JaegerStorageSpec{Type: "memory"}},
 		},
 		{
 			j: &v1.JaegerSpec{Storage: v1.JaegerStorageSpec{Options: v1.NewOptions(map[string]interface{}{"es-archive.enabled": "true"})}},
 			expected: &v1.JaegerSpec{Storage: v1.JaegerStorageSpec{Options: v1.NewOptions(map[string]interface{}{"es-archive.enabled": "true"})},
-				UI: v1.JaegerUISpec{Options: v1.NewFreeForm(map[string]interface{}{"archiveEnabled": true})}},
+				UI: v1.JaegerUISpec{Options: v1.NewFreeForm(map[string]interface{}{"archiveEnabled": true, "dependencies": map[string]interface{}{"menuEnabled": false}})}},
```
objectiser: Isn't this wrong - if the storage type is not defined, it should default to `memory`, which should show the dependencies tab. The expected results is showing it disabled.

pkg/strategy/controller_test.go
```diff
@@ -246,38 +246,94 @@ func TestNormalizeUI(t *testing.T) {
...
+		{
+			uiOpts:   map[string]interface{}{"dependencies": map[string]interface{}{"menuEnabled": "respectThis"}},
+			storage:  "whateverStorage",
+			expected: map[string]interface{}{"dependencies": map[string]interface{}{"menuEnabled": "respectThis"}},
+		},
+		{
+			uiOpts:   map[string]interface{}{"dependencies": map[string]interface{}{"foo": "bar"}},
+			storage:  "whateverStorage",
+			expected: map[string]interface{}{"dependencies": map[string]interface{}{"foo": "bar", "menuEnabled": false}},
+		},
```
objectiser: Can you add one test with `test.enabled` set to `false`.

pkg/strategy/controller_test.go
pavolloffay: In general yes, but this tests only normalizeUI which works on any storage

pkg/strategy/controller.go
```diff
@@ -135,23 +135,48 @@ func normalizeRollover(spec *v1.JaegerEsRolloverSpec) {
...
+		if strings.EqualFold(sOpts["es-archive.enabled"], "true") ||
+			strings.EqualFold(sOpts["cassandra-archive.enabled"], "true") {
+			uiOpts["archiveEnabled"] = true
+		}
+	}
+}
+
+func disableDependenciesTab(uiOpts map[string]interface{}, storage string, depsEnabled *bool) {
+	// dependency tab is by default enabled and memory storage support it
+	if strings.EqualFold(storage, "memory") || (depsEnabled != nil && *depsEnabled == false) {
```
objectiser: Just realised, with the additional test, if dependencies are explicitly disabled, then menuEnabled should be false?

objectiser: so should the condition be `if strings.EqualFold(storage, "memory") || depsEnabled == nil || *depsEnabled == true) {`?

pavolloffay: You are right, changing

pkg/strategy/controller.go
```diff
@@ -161,7 +161,7 @@ func enableArchiveButton(uiOpts map[string]interface{}, sOpts map[string]string)
 
 func disableDependenciesTab(uiOpts map[string]interface{}, storage string, depsEnabled *bool) {
 	// dependency tab is by default enabled and memory storage support it
-	if strings.EqualFold(storage, "memory") || (depsEnabled != nil && *depsEnabled == false) {
+	if strings.EqualFold(storage, "memory") || (depsEnabled != nil && *depsEnabled == true) {
```
objectiser: I think it needs to be `... || depsEnabled == nil || *depsEnabled == true` - as if the `depsEnabled` is nil, then it means dependencies are enabled by default, doesn't it?

pavolloffay: The normalization of deps spec runs before - so we assuming correct settings here

objectiser:
@pavolloffay LGTM - although there is a test failure.

Issue #227, opened 2019-02-22
e2e tests do not work on OpenShift


pavolloffay:
The e2e tests does not work on openshift bc:
* https://github.com/jaegertracing/jaeger-operator/blob/610d415ca5dfe2f14b2f06f4a3bef3fa5520457b/test/e2e/daemonset.go#L143
* personally on minishift I could not select default namespace and our tests depend on services (es, cassandra) deployed in default namespce

jpkrohling:
cc @kear

kevinearls:
Most of the work here is done.  There are two small bugs left, #326 which is a fix to get the all-in-one-with-ingress test working on OpenShift, and #327 for the DaemonSet test, which is currently blocked by #322.  I think this can be closed.

PR #323, opened 2019-03-15
Update tests to work on OpenShift

kevinearls:
This is the second set of changes for #227.  I've done the following

- Changed several tests to use portForwarding rather than ingress as suggested by @pavolloffay .  This simplifies making them work both on Kubernetes and OpenShift
- Created a version of the daemonsettest called daemonset_with_ingress to cover ingress testing.  This currently does not work on OpenShift as we call WaitForIngress to make sure deplyment has finished, but on OpenShift a route gets created instead.  We need to either figure out how to find routes with kubeclient or start using a openshift client for this test.
- I've refactored a couple of things (specifically creating getVertxDeployment and getJaegerDefinition functions in daemonset test to try to reduce repetition and make test cases shorter.  I'll try to do more of this in the future.


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/323)

kevinearls:
@pavolloffay @objectiser @jpkrohling Please review

pavolloffay:
> We need to either figure out how to find routes with kubeclient or start using a openshift client for this test.

Routes can be queried with the same API as other objects https://github.com/jaegertracing/jaeger-operator/blob/fcb26f488a5960f37fd63221954ac5929c031a9e/pkg/controller/jaeger/route.go#L19

You can use this PR to see how we know that we are running on OCP or k8s https://github.com/jaegertracing/jaeger-operator/pull/217

kevinearls:
HI @objectiser I've updated this based on your previous comments.  Could you please take another look?

kevinearls:
@objectiser I'm looking at that now.  If possible though I'd like to get this merged and add that as a separate issue.

objectiser:
@kevinearls that is fine with me.

kevinearls:
@objectiser Thanks.  Can you merge it then, or do I need review from @pavolloffay and @jpkrohling too?

PR #324, opened 2019-03-15
Load back Elasticsearch certs from secrets

### LINKED ISSUE #320: Store Elasticsearch master certs (ca and key) into secret
We have to store ES master certs `ca.crt`, `ca.key` to a secret. These certs are needed to recreate client certs used by jaeger and curator. 

The recreation of secrets happens on every loop. The script for creating the certs checks whether ca and key are present on operator's filesystem  - if they are it uses them. The problem happens when the operator restarts - It will create new certs. 

There is also other issue, we store certs in one directory at the moment so all instances get same ca and key certs.

The actions:
- [ ] store master certs in a secret per instance
- [ ] load master certs to filesystem before generating certs
- [ ] set working directory for cert generation script to something which includes namespace and jaeger instance name

pavolloffay:
Resolves #320 

* avoids generating different certs if the operator restarts
* restores edited secrets by user
* stores certs in `/tmp/_cert/namespace/instance_name/uuid`

### DIFF
--- a/pkg/storage/elasticsearch_secrets.go
+++ b/pkg/storage/elasticsearch_secrets.go
@@ -104,7 +104,7 @@ func extractSecretsToFile(jaeger *v1.Jaeger, secrets []corev1.Secret, s ...secre
 	}
 	for _, sec := range s {
 		if secret, ok := secretMap[sec.instanceName(jaeger)]; ok {
-			if err := extractSecretToFile(jaeger, secret.Data, sec); err != nil {
+			if err := extractSecretToFile(getWorkingDir(jaeger), secret.Data, sec); err != nil {
 				return errors.Wrap(err, fmt.Sprintf("failed to extract secret %s", secret.Name))
 			}
 		}
@@ -112,9 +112,9 @@ func extractSecretsToFile(jaeger *v1.Jaeger, secrets []corev1.Secret, s ...secre
 	return nil
 }
 
-func extractSecretToFile(jaeger *v1.Jaeger, data map[string][]byte, secret secret) error {
+func extractSecretToFile(workingDir string, data map[string][]byte, secret secret) error {
 	for k, v := range secret.keyFileNameMap {
-		if err := writeToFile(getWorkingDir(jaeger), v, data[k]); err != nil {
+		if err := writeToFile(workingDir, v, data[k]); err != nil {
 			return err
 
 		}
@@ -207,7 +207,7 @@ func writeToFile(dir, file string, value []byte) error {
 	defer f.Close()
 	_, err = f.Write(value)
 	if err != nil {
-		// remove the file on failure - so it can be correctly created on the next iteration
+		// remove the file on failure - it can be correctly created in the next iteration
 		os.RemoveAll(path)
 		return err
 	}
--- a/pkg/storage/elasticsearch_secrets.go
+++ b/pkg/storage/elasticsearch_secrets.go
@@ -200,7 +200,6 @@ func writeToWorkingDirFile(dir, toFile string, value []byte) error {
 	if err := os.MkdirAll(dir, os.ModePerm); err != nil {
 		return err
 	}
-	//#nosec G302 Expect file permissions to be 0600 or less (Confidence: HIGH, Severity: MEDIUM)
 	f, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE, 0600)
 	if err != nil {
 		return err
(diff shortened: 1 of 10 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/324)

pkg/storage/elasticsearch_secrets.go
```diff
@@ -40,54 +58,81 @@ var esSecret = secret{
...
+		createSecret(jaeger, jaegerSecret.instanceName(jaeger), getWorkingDirContents(getWorkingDir(jaeger), jaegerSecret.keyFileNameMap)),
+		createSecret(jaeger, curatorSecret.instanceName(jaeger), getWorkingDirContents(getWorkingDir(jaeger), curatorSecret.keyFileNameMap)),
 	}
 }
 
 // CreateESCerts runs bash scripts which generates certificates
-func CreateESCerts() error {
-	return createESCerts(certScript)
+// The secrets are pulled back to FS in case of operator restart
+// The scrip checks if secrets are expired or need to be regenerated
```
objectiser: nit: script

pkg/storage/elasticsearch_secrets.go
```diff
@@ -145,3 +190,25 @@ func getFileContents(path string) []byte {
...
 	return contents
 }
+
+func writeToWorkingDirFile(dir, toFile string, value []byte) error {
+	// first check if file exists - we prefer what is on FS to revert users editing secrets
+	path := getFilePath(dir, toFile)
+	if _, err := os.Stat(path); err == nil {
+		return nil
+	}
+	if err := os.MkdirAll(dir, os.ModePerm); err != nil {
```
objectiser: Thought this method would only be called for overwriting the secret data with existing secret - so shouldn't the file already exist? Not a big issue, but just wanted to clarify my understanding.

pkg/storage/elasticsearch_secrets_test.go
```diff
@@ -1,55 +1,44 @@
...
-			ca, err := ioutil.ReadFile(workingDir + "/ca.crt")
+		if s.Name == jaegerSecret.instanceName(j) {
+			ca, err := ioutil.ReadFile(tmpWorkingDir + "/foo/ca.crt")
 			assert.NoError(t, err)
 			assert.Equal(t, map[string][]byte{"ca": ca}, s.Data)
 		}
 	}
 }
 
 func TestCreteSecret(t *testing.T) {
```
objectiser: nit :"TestCreateSecret" - not part of PR, but would be good to fix.

objectiser:
@pavolloffay Sorry forgot to mention code coverage needs fixing.

pkg/storage/elasticsearch_secrets.go
pavolloffay: This method is called on every loop. The previous statement returns if file exists. If the file does not exists we have to create before creating the file.

pavolloffay:
> Do we have to fail here? Can't we recover from this?

We do fail in the controller if we cannot get objects. So maybe better to keep it the same for consistency.

> The secrets should be set directly on the strategy afterwards instead of being passed to the strategy chooser. Otherwise, it implies that the secrets are somehow part of the selection algorithm, which is not the case.

Unfortunately, the `str := r.runStrategyChooser(instance)` already runs strategy and sets all objects. To do what you propose we would have to refactor to choose the strategy first and then run it.

> I guess I missed this during the initial review, but for the record: a cleaner solution is to use os.TempDir(), as /tmp might not be the right place.

Let's address this when it actually will be an issue. Btw we do create the dir if it does not exists.

> Users should be instructed to change the secret instead of the Operator's file system.

No, we trust the operator and not the objects. What you propose means that users could also edit deployments and jaeger would change the CR appropriately.

pavolloffay:
> As we talked on IRC, the best would be to have the secret as the source of truth. This is consistent with other Kubernetes/OpenShift objects. If the secret does not exist, the operator will generate the appropriate certs and update the secret. Later on, the Operator creates the cert in the file system and synchronizes during every reconciliation, in case the backing secret changes.

At the moment we don't officially support users providing their own certs - secrets. That said this PR assures that secrets are what operator created. If a users tries to edit them it will be reverted. 

We should track users providing their own certs in a separate issue.

pavolloffay:
I have created https://github.com/jaegertracing/jaeger-operator/issues/330 for providing certs by user

PR #340, opened 2019-03-21
Update allInOneWithIngressTest to work on OpenShift

### LINKED ISSUE #326: Update allInOneWithIngressTest to work on OpenShift
This requires updating the test to get routes even when using the kubernetes client.  See https://github.com/jaegertracing/jaeger-operator/blob/fcb26f488a5960f37fd63221954ac5929c031a9e/pkg/controller/jaeger/route.go#L19

kevinearls:

This resolves #326 by adding the ability to find routes with the kubeclient.  

It also contains a minor change to the isOpenShift function, as I realized I did not need to pass it a framework object.

### DIFF
--- a/test/e2e/all_in_one_test.go
+++ b/test/e2e/all_in_one_test.go
@@ -2,7 +2,9 @@ package e2e
 
 import (
 	goctx "context"
+	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -10,11 +12,14 @@ import (
 	"testing"
 	"time"
 
+	osv1 "github.com/openshift/api/route/v1"
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
 	"github.com/operator-framework/operator-sdk/pkg/test/e2eutil"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/net/context"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
 )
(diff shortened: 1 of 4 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/340)

kevinearls:
@objectiser  @pavolloffay @jpkrohling Please review.

test/e2e/all_in_one_test.go
```diff
@@ -74,43 +61,21 @@ func allInOneTest(t *testing.T, f *framework.Framework, ctx *framework.TestCtx)
 	return e2eutil.WaitForDeployment(t, f.KubeClient, namespace, "my-jaeger", 1, retryInterval, timeout)
 }
 
+
```
objectiser: nit: unnecessary line gaps

test/e2e/all_in_one_test.go
```diff
@@ -119,28 +84,50 @@ func allInOneWithIngressTest(t *testing.T, f *framework.Framework, ctx *framewor
...
+			return err
+		}
+
+		url = fmt.Sprintf("https://%s/api/services", routeUrl)
+		transport := &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		httpClient= http.Client{Timeout: 30 * time.Second, Transport: transport}
+	} else {
+		ingress, err := WaitForIngress(t, f.KubeClient, namespace, "my-jaeger-with-ingress-query", retryInterval, timeout)
```
objectiser: Wondering why the names are different?
@jpkrohling I think this is because the route is named with `-query` whereas the ingress isn't. Do we actually need this distinction or would it be better to be consistent? (either way - i.e. add `-query` to ingress or remove it from route)?

test/e2e/all_in_one_test.go
```diff
@@ -119,28 +84,50 @@ func allInOneWithIngressTest(t *testing.T, f *framework.Framework, ctx *framewor
...
+		if err != nil {
+			return err
+		}
+
+		if len(ingress.Status.LoadBalancer.Ingress) != 1 {
```
objectiser: Should this test be performed in the openshift branch aswell?

test/e2e/all_in_one_test.go
```diff
@@ -260,3 +248,53 @@ func allInOneWithUIConfigTest(t *testing.T, f *framework.Framework, ctx *framewo
 		return true, nil
 	})
 }
+
+func getJaegerDefinition(namespace string, name string) *v1.Jaeger {
```
objectiser: `getJaegerAllInOneDefinition` - just to be clear what it is returning.

test/e2e/all_in_one_test.go
jpkrohling: AFAIR, we don't access ingresses by its name, because it might need some DNS changes. So, we get the ingress IP and use it instead. Routes create a hostname based on the backing service name (query in our case) that could probably be reached by the client if it uses something like `nip.io`.

objectiser: Sounds reasonable - was mistaken anyway, route and ingress are both named with `-query` suffix. So ok as is - only possible change could be to move the `WaitForDeployment` above the if/else as independent of whether using openshift or k8s?

kevinearls: I could move the WaitForDeployment, but in the Kubernetes case WaitForIngress serves the same purpose.  Let me know what you prefer, I can do it either way.

objectiser: Might be better to move, as not sure whether the ingress being available guarantees that the deployment is available - so moving might avoid issues later with timing.

test/e2e/all_in_one_test.go
kevinearls: @objectiser what would be the equivalent for a route?  len(route.Status.Ingress)?

objectiser: Think so.

jpkrohling:
E2E tests are passing:

```
Running Smoke end-to-end tests...
ok  	github.com/jaegertracing/jaeger-operator/test/e2e	211.829s
Creating namespace default
service/cassandra created
statefulset.apps/cassandra created
ok  	github.com/jaegertracing/jaeger-operator/test/e2e	180.979s
statefulset.apps/elasticsearch created
service/elasticsearch created
Running Elasticsearch end-to-end tests...
ok  	github.com/jaegertracing/jaeger-operator/test/e2e	172.227s
```

Merging.

PR #346, opened 2019-03-25
Added instructions for daemonsets on OpenShift

jpkrohling:


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/346)

README.adoc
```diff
@@ -96,6 +92,36 @@ oc create \
 
 After the role is granted, switch back to a non-privileged user.
 
+Jaeger Agent can be configured to be deployed as a `DaemonSet` using a `HostPort` to allow Jaeger clients in the same node to discover the agent. In OpenShift, a `HostPort` can only be set when a special security context is set. A security context constraint allowing `DaemonSet` components to bind to a `HostPort` can be specified as follows:
```
objectiser: Shouldn't this text all be before the previous statement, switching back to the non-privileged user?

Security concern - isn't this potentially enabling non-privileged users to deploy daemonsets on host ports? Isn't there a way to just grant the permission to the Jaeger operator, and the non-privileged user is then only able to deploy a Jaeger instance, which indirectly deploys daemonset with hostport?

jpkrohling: > Shouldn't this text all be before the previous statement, switching back to the non-privileged user?

Possibly. The reasoning is that users will *only* need to read on if they plan to allow agents as daemonsets.

> Security concern - isn't this potentially enabling non-privileged users to deploy daemonsets on host ports?

Yes, and good question. Would you know who might be able to help us here?

jpkrohling:
@objectiser, @kevinearls I changed this PR with the results of my latest tests on OpenShift 4. Could you please take a look and give it a try as well?

@rcernich, do you think the instructions from this PR are sufficient to fix #459?

README.adoc
```diff
@@ -97,6 +97,33 @@ oc create \
 
 After the role is granted, switch back to a non-privileged user.
 
+Jaeger Agent can be configured to be deployed as a `DaemonSet` using a `HostPort` to allow Jaeger clients in the same node to discover the agent. In OpenShift, a `HostPort` can only be set when a special security context is set. A separate service account can be used by the Jaeger Agent with the permission to bind to `HostPort`, as follows:
+
+[source,bash]
+----
+oc create -f https://raw.githubusercontent.com/jaegertracing/jaeger-operator/master/deploy/hostport-scc-daemonset.yaml # <1>
```
objectiser: Need to update the URL as moved to `examples/openshift`

README.adoc
```diff
@@ -97,6 +97,33 @@ oc create \
...
+
+oc new-project myappnamespace
+oc create -f https://raw.githubusercontent.com/jaegertracing/jaeger-operator/master/deploy/service_account_jaeger-agent-daemonset.yaml # <2>
```
objectiser: Same here.

PR #350, opened 2019-04-02
Update test to run on OpenShift

### LINKED ISSUE #327: Update DaemonsetTest to work on OpenShift
This can be done after #322 is fixed

kevinearls:

This resolves #327 by taking actions that are needed to permit this test to run on OpenShift.  Part of the solution is a bit hacky as it uses os.Exec to call the "oc adm" command, which means the oc client needs to be on the path wherever this test is being run.

It would be preferable to use either the OpenShift Go client of the Rest API to do this, but we don't use those anywhere else in the tests and adding them would be non-trivial.  I will open an issue to do this in the future.

### DIFF
--- a/test/e2e/daemonset.go
+++ b/test/e2e/daemonset.go
@@ -42,11 +42,12 @@ func daemonsetTest(t *testing.T, f *framework.Framework, ctx *framework.TestCtx)
 	}
 
 	if isOpenShift(t) {
-		err = f.Client.Create(goctx.TODO(), hostportSccDaemonset(), cleanupOptions)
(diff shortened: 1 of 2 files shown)


jpkrohling:
This change is [Reviewable](https://reviewable.io/reviews/jaegertracing/jaeger-operator/350)

kevinearls:
@jpkrohling @objectiser  Please review.

kevinearls:
@objectiser @jpkrohling Can one of you guys merge this?  Juca I assume the code-review/reviewable failure is because you turned that off?

jpkrohling:
That's correct. Sorry I didn't merge it earlier, I was holding it off until I executed the tests locally and ran into #355, but I assume you tested it as well.

PR #371, opened 2019-04-10
Include elasticsearch statefulset nodes in availability check

### LINKED ISSUE #343: Wait for elasticseach should query also statefulsest
Before deploying jaeger we are waiting for es nodes to come UP. To do that we are querying only ES deployment objects. However ES operator creates also statfulset but only for single master nodes. We create single master nodes if the number of ES nodes is higher that 3.

https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/controller/jaeger/elasticsearch.go#L58

pavolloffay:
Resolves #343 

### DIFF
--- a/pkg/controller/jaeger/elasticsearch.go
+++ b/pkg/controller/jaeger/elasticsearch.go
@@ -83,6 +83,21 @@ func waitForAvailableElastic(c client.Client, es esv1.Elasticsearch) error {
 				available++
 			}
 		}
+		ssList := corev1.StatefulSetList{}
+		if err = c.List(context.Background(), client.MatchingLabels(labels).InNamespace(es.Namespace), &ssList); err != nil {
+			if k8serrors.IsNotFound(err) {
+				// the object might have not been created yet
+				log.WithFields(log.Fields{
+					"namespace": es.Namespace,
+					"name":      es.Name,
+				}).Debug("Elasticsearch cluster doesn't exist yet.")
+				return false, nil
+			}
+			return false, err
+		}
+		for _, d := range ssList.Items {
+			available += d.Status.ReadyReplicas
+		}
 		logrus.WithFields(logrus.Fields{
 			"namespace":      es.Namespace,
 			"name":           es.Name,


pavolloffay:
The coverage dropped, I think we do not test wait in other controller either. Not sure how to test it either in unit tests.

pkg/controller/jaeger/elasticsearch.go
```diff
@@ -83,6 +83,21 @@ func waitForAvailableElastic(c client.Client, es esv1.Elasticsearch) error {
 				available++
 			}
 		}
+		ssList := corev1.StatefulSetList{}
```
jpkrohling: What's the relationship between the deployment above and this stateful set? Is it sufficient to just check the stateful set?

pkg/controller/jaeger/elasticsearch.go
```diff
@@ -83,6 +83,21 @@ func waitForAvailableElastic(c client.Client, es esv1.Elasticsearch) error {
...
+				log.WithFields(log.Fields{
+					"namespace": es.Namespace,
+					"name":      es.Name,
+				}).Debug("Elasticsearch cluster doesn't exist yet.")
+				return false, nil
+			}
+			return false, err
+		}
+		for _, d := range ssList.Items {
+			available += d.Status.ReadyReplicas
```
jpkrohling: That may seem a bit confusing when debugging an issue: when there's a mismatch, which ones aren't ready?

Wouldn't it be better to have two counters, one for the deployments and one for the stateful sets?

pkg/controller/jaeger/elasticsearch.go
pavolloffay: SS is used exclusively for _only_ master nodes. The deployments are used for data or other combinations.

pkg/controller/jaeger/elasticsearch.go
pavolloffay: +1 I have split them into SS and deployments.

PR #382, opened 2019-04-29
Fix race condition in test, remove hard-coded namespace

### LINKED ISSUE #380: Fix race condition in e2e SmokeTest
The smoketest can fail here if the first span has not been flushed by the time the test does the http get.

https://github.com/jaegertracing/jaeger-operator/blob/master/test/e2e/smoketest.go#L46-L52

kevinearls:

This fixes #380 where the test would fail if we did the http get before the span got flushed.  I've also removed the hardcoded namespace of "default" from the es_index_cleaner_test.

### DIFF
--- a/test/e2e/smoketest.go
+++ b/test/e2e/smoketest.go
@@ -43,12 +43,10 @@ func SmokeTest(apiTracesEndpoint, collectorEndpoint, serviceName string, interva
 		bodyBytes, err := ioutil.ReadAll(resp.Body)
 		bodyString := string(bodyBytes)
 
+
 		if !strings.Contains(bodyString, "errors\":null") {
 			return false, errors.New("query service returns errors")
-		} else if (strings.Contains(bodyString, tStr)) {
-			return true, nil
-		} else {
-			return false, nil
 		}
+		return strings.Contains(bodyString, tStr), nil
 	})
 }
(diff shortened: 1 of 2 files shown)


kevinearls:
@pavolloffay @objectiser @jpkrohling Please review

test/e2e/smoketest.go
```diff
@@ -43,12 +43,12 @@ func SmokeTest(apiTracesEndpoint, collectorEndpoint, serviceName string, interva
 		bodyBytes, err := ioutil.ReadAll(resp.Body)
 		bodyString := string(bodyBytes)
 
-		if !strings.Contains(bodyString, "errors\":null") {
+		if (strings.Contains(bodyString, tStr)) {
+			return true, nil
+		} else if !strings.Contains(bodyString, "errors\":null") {
```
pavolloffay: Keep the errors check separate from the expected spas and move it above it

kevinearls: @pavolloffay Are you ok with these changes?

test/e2e/smoketest.go
```diff
@@ -45,10 +45,10 @@ func SmokeTest(apiTracesEndpoint, collectorEndpoint, serviceName string, interva
 
 		if !strings.Contains(bodyString, "errors\":null") {
```
pavolloffay: Can be written as

```golang
if !strings.Contains(bodyString, "errors\":null") {
			return false, errors.New("query service returns errors")
		} 

return  strings.Contains(bodyString, tStr)), nil
```

PR #389, opened 2019-05-03
Make sure to call ctx.cleanup if prepare() fails

### LINKED ISSUE #388: e2e tests don't clean up if they fail during operator deploy
If an e2e tests fails because the operator does not deploy during the prepare call, context.cleanup will not be called, and therefore the namespace that was created will not be deleted.

kevinearls:

This fixed #388   Currently if tests fail during the prepare() call ctx.cleanup() will not be called and the namespace created for the test does not get cleaned up.

### DIFF
--- a/test/e2e/cassandra.go
+++ b/test/e2e/cassandra.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
 	"github.com/operator-framework/operator-sdk/pkg/test/e2eutil"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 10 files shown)


test/e2e/cassandra.go
```diff
@@ -15,7 +16,11 @@ import (
 
 // Cassandra runs a test with Cassandra as the backing storage
 func Cassandra(t *testing.T) {
-	ctx := prepare(t)
+	ctx, err := prepare(t)
```
pavolloffay: Could we call `prepare` in the caller so each test would not have to repeat this block?

kevinearls: @pavolloffay Once I finish the other changes I am making for reporting it will be called in the Setup method, only once per suite.  https://github.com/jaegertracing/jaeger-operator/blob/master/test/e2e/all_in_one_test.go#L36-L44

I've run into this bug a few times though, so I wanted to get this done first.

kevinearls:
@pavolloffay @objectiser @jpkrohling  Please review.

PR #430, opened 2019-05-23
Workaround for Jaeger 1468 create traces so index cleaning doesn't fail

kevinearls:

This is a workaround for https://github.com/jaegertracing/jaeger/issues/1468  It creates at least one trace so that the es-index-cleaner jobs will not fail on account of there not being any indices.

### DIFF
--- a/test/e2e/elasticsearch_test.go
+++ b/test/e2e/elasticsearch_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


kevinearls:
@objectiser  @pavolloffay  Please review

kevinearls:
Good point.  I guess there is a race condition here, and this way the first trace gets created faster, so it's there the first time the es-index-cleaner runs?  Let me see if I can find another solution.

PR #454, opened 2019-05-30
Scan deployments for agent injection

rubenvp8510:


pkg/inject/sidecar.go
```diff
@@ -177,3 +177,34 @@ func hasEnv(name string, vars []corev1.EnvVar) bool {
...
+
+// RequireInjections deployments
+func RequireInjections(jaeger *v1.Jaeger, pods *v1.JaegerList, deps *appsv1.DeploymentList) []*appsv1.Deployment {
+	requireUpdates := make([]*appsv1.Deployment, 0)
+	for i := 0; i < len(deps.Items); i++ {
+		dep := &deps.Items[i]
+		if Needed(dep) {
+			if belongsToJaegerInstance(dep, jaeger, len(pods.Items) == 1) {
+				// a suitable jaeger instance was found! let's inject a sidecar pointing to it then
+				log.WithFields(log.Fields{
```
pavolloffay: I think you can call logger on jaeger instance `jager.Log()`

pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -220,5 +222,29 @@ func (r *ReconcileJaeger) apply(jaeger v1.Jaeger, str strategy.S) error {
...
 	}
 
+	if err := r.scanDeployments(jaeger); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *ReconcileJaeger) scanDeployments(jaeger v1.Jaeger) error {
```
pavolloffay: What does the func do? Does it scan for something or does it add sidecars?

rubenvp8510: The name is wrong, not intuitive at all, sorry

It suppose to add the sidecar for deployments that require that. But I think I found a better way for do this.

jpkrohling:
@rubenvp8510, are you still interested in working on this one? We might change in the future to use the webhook approach, which would render this one here obsolete, but this is indeed fixing a bug we have right now, IIRC.

jpkrohling:
We now have an auto-detect package (`pkg/autodetect`) with a process that runs periodically in the background. If you decide to continue working on this one, it would be good to use that feature.

rubenvp8510:
@jpkrohling Yes I want to update this PR, I'll check the autodetect stuff and update this.

Thanks!

rubenvp8510:
@jpkrohling I'm not totally convinced of using auto-detect feature on this, I updated the code, now the actual code uses that feature. but I need to get all deployments and scan for updates/injections. Not sure if it's a better way to do it, I have the problem that I cannot query on annotations.

jpkrohling:
@rubenvp8510, wasn't this implemented in another PR? Should this one be closed?

rubenvp8510:
@jpkrohling this is another issue, I think you are confusing this PR with https://github.com/jaegertracing/jaeger-operator/pull/453, which is the opposite.

This is about to inject the agent on existing deployments, the other is about deletion.  Not sure what is the decision regarding of this PR, Do we want to proceed in this way ?or are we going to work more on webhook admission controller?

pkg/inject/sidecar.go
```diff
@@ -62,6 +62,17 @@ func Sidecar(jaeger *v1.Jaeger, dep *appsv1.Deployment) *appsv1.Deployment {
 	return dep
 }
 
+// UpdateSideCar modify the deployment side car with the latest parameters if it's required.
+func UpdateSideCar(jaeger *v1.Jaeger, dep *appsv1.Deployment) bool {
```
jpkrohling: `s/SideCar/Sidecar`

jpkrohling:
@rubenvp8510 do you want to merge this, and address the remaining comments in a follow-up PR?

rubenvp8510:
> @rubenvp8510 do you want to merge this, and address the remaining comments in a follow-up PR?

Yes please! :)

jpkrohling:
Merged. Please, create an issue to track the remaining points.

Issue #421, opened 2019-05-17
Enable specifying serviceAccount and securityContext


mwieczorek:
When we create a new Jaeger resource, new serviceAccount with the same name is created and all jaeger components are run with this SA.
It's not possible to set securityContext for the Jaeger components

Enable specifying serviceAccount and securityContext in jaeger CR (top-level and on each component separately).

jpkrohling:
Would you be interested in contributing this feature?

mwieczorek:
I don’t do golang day to day, but I can try.
Of course if someone else from the community finds this „good first issue” and would like to do it - Please comment here and go ahead.

PR #456, opened 2019-06-03
Add support for securityContext and serviceAccount

mwieczorek:
Signed-off-by: mwieczorek <wieczorek-michal@wp.pl>

fix: #421


pkg/account/main.go
```diff
@@ -46,6 +47,21 @@ func getMain(jaeger *v1.Jaeger) *corev1.ServiceAccount {
...
+	case "query":
+		sa = util.Merge([]v1.JaegerCommonSpec{jaeger.Spec.Query.JaegerCommonSpec, jaeger.Spec.JaegerCommonSpec}).ServiceAccount
+	case "ingester":
+		sa = util.Merge([]v1.JaegerCommonSpec{jaeger.Spec.Ingester.JaegerCommonSpec, jaeger.Spec.JaegerCommonSpec}).ServiceAccount
+	case "all-in-one":
+		sa = util.Merge([]v1.JaegerCommonSpec{jaeger.Spec.AllInOne.JaegerCommonSpec, jaeger.Spec.JaegerCommonSpec}).ServiceAccount
+	}
+
+	if sa == "" {
+		return fmt.Sprintf("%s", jaeger.Name)
```
jpkrohling: Or just `return jaeger.Name` ?

pkg/account/main.go
```diff
@@ -46,6 +47,21 @@ func getMain(jaeger *v1.Jaeger) *corev1.ServiceAccount {
 }
 
 // JaegerServiceAccountFor prints service name for Jaeger instance
-func JaegerServiceAccountFor(jaeger *v1.Jaeger) string {
-	return fmt.Sprintf("%s", jaeger.Name)
+func JaegerServiceAccountFor(jaeger *v1.Jaeger, component string) string {
+	sa := ""
+	switch component {
+	case "collector":
```
jpkrohling: It would probably make sense to use a typed constant here, like, `CollectorComponent Component` and change the func's signature to `component Component`

pkg/account/main.go
mwieczorek: sure, first I just left the old version, but now it's changed.

pkg/account/main.go
mwieczorek: added new type/const

mwieczorek:
Added 2 new configuration options to readme file.
Let me know if I should provide more extended description somewhere else.

Also one question:
I see in https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/deployment/agent.go#L76 that agent daemonset has no explicit serviceAccount set. Is it by design or we should also add the configuration option there?

jpkrohling:
> agent daemonset has no explicit serviceAccount set

This is not by design, I see no problems in adding the configuration option for it as well.

pkg/util/util.go
```diff
@@ -10,6 +10,16 @@ import (
 	"github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
 )
 
+// Component represents type of jaeger component
+type Component string
+
+const (
```
jpkrohling: Not sure `util` is the right place for this. We have some constants already in the main `v1` package (github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1), but that might also not be suitable.

As all usages of this new constant are inside the `deployment` package, I'd keep it private there in a file named `const.go`. We can move it later if we need it somewhere else.

jpkrohling:
By the way, Travis fails here:

```
/home/travis/gopath/src/github.com/jaegertracing/jaeger-operator/pkg/util/util.go:17:2: exported const CollectorComponent should have comment (or a comment on this block) or be unexported
```

mwieczorek:
> This is not by design, I see no problems in adding the configuration option for it as well.

With default value `default` (as it is now) or `jaeger` (as it is for other components. Will it make a breaking change and should it be documented somewhere too - release notes etc?)

jpkrohling:
I would vote for `jaeger` + release notes, as I think the potential to break existing users is very low.

pkg/util/util.go
mwieczorek: I also use it in `account` package, so if I put it in `deployment` package I cannot use it in `account` (cyclic import).
Put it in `account` package? (there's `JaegerServiceAccountFor` func where it's used)

jpkrohling: Good idea, placing it close to the `JaegerServiceAccountFor` sounds logical to me.

PR #523, opened 2019-07-12
Avoid touching the original structure of the options

rubenvp8510:

Fixes Issue #304 

Changed the way that the options is serialized/deserialized, now options does not touch the yaml structure on an update.

This is similar to Freeform type, but it has internally a method for "flat" options, so this support both types of properties, condensed form and normal structure, but not a mix of them.

may be we can/should merge this type with Freeform ?

### DIFF
--- a/pkg/apis/jaegertracing/v1/options_test.go
+++ b/pkg/apis/jaegertracing/v1/options_test.go
@@ -118,3 +118,20 @@ func TestExposedMap(t *testing.T) {
 	o.UnmarshalJSON([]byte(`{"cassandra": {"servers": "cassandra:9042"}}`))
 	assert.Equal(t, "cassandra:9042", o.Map()["cassandra.servers"])
 }
+
+func TestMarshallRaw(t *testing.T) {
+	json := []byte(`{"cassandra": {"servers": "cassandra:9042"}}`)
+	o := NewOptions(nil)
+	o.json = json
+	bytes, err := o.MarshalJSON()
+	assert.NoError(t, err)
+	assert.Equal(t, bytes, json)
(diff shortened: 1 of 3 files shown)


pkg/apis/jaegertracing/v1/options_test.go
```diff
@@ -118,3 +118,20 @@ func TestExposedMap(t *testing.T) {
...
 	assert.Equal(t, "cassandra:9042", o.Map()["cassandra.servers"])
 }
+
+func TestMarshallRaw(t *testing.T) {
+	json := []byte(`{"cassandra": {"servers": "cassandra:9042"}}`)
+	o := NewOptions(nil)
+	o.json = json
+	bytes, err := o.MarshalJSON()
+	assert.NoError(t, err)
+	assert.Equal(t, bytes, json)
```
jpkrohling: What is the behavior before this change? `{"cassandra.servers": "cassandra:9042"}`?

rubenvp8510: The change this PR does is on marshal and unmarshall, previously it flats all options into condensed form, now I don't do that, because it could produce Jaeger resources with mixed condensed and structured form of options, I just leave the options intact.

So for this case will return the JSON in the way it is, this is:  `{"cassandra": {"servers": "cassandra:9042"}}`  , but from code perspective you can still access to a map with condensed form of properties.

I did in this way trying to preserve backwards compatibility, if someone is already using condensed form, it will work, same for users  using the structured form of options.

Other option is to change this to map[string]string, but that breaks the compatibility for users that uses structured form

PR #531, opened 2019-07-16
Use ES single redundancy by default

pavolloffay:
Related to https://github.com/jaegertracing/jaeger-operator/pull/501#discussion_r304001869

### DIFF
--- a/pkg/strategy/controller.go
+++ b/pkg/strategy/controller.go
@@ -133,7 +133,7 @@ func normalizeElasticsearch(spec *v1.ElasticsearchSpec) {
 		spec.NodeCount = 1
 	}
 	if spec.RedundancyPolicy == "" {
-		spec.RedundancyPolicy = esv1.ZeroRedundancy
+		spec.RedundancyPolicy = esv1.SingleRedundancy
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/strategy/controller.go
```diff
@@ -133,7 +133,7 @@ func normalizeElasticsearch(spec *v1.ElasticsearchSpec) {
 		spec.NodeCount = 1
 	}
 	if spec.RedundancyPolicy == "" {
-		spec.RedundancyPolicy = esv1.ZeroRedundancy
+		spec.RedundancyPolicy = esv1.SingleRedundancy
```
pavolloffay: This string option is used to specify replica shards. Instead of exposing an integer ECL decided to expose a human readable string. The number of nodes is taken into the account when calculating the replica shards.

https://github.com/jaegertracing/jaeger-operator/blob/0d6bd5fbd5b5ecb6f8138d7a5222689962dbd1aa/pkg/storage/elasticsearch.go#L200-L212

This should be defaulting to `SingleRedundancy` to keep backwards compatibility.

pkg/strategy/controller.go
```diff
@@ -133,7 +133,7 @@ func normalizeElasticsearch(spec *v1.ElasticsearchSpec) {
 		spec.NodeCount = 1
 	}
 	if spec.RedundancyPolicy == "" {
```
pavolloffay: Here we should be also checking the number of nodes. If it is a single node cluster the redundancy should be zero - https://angristan.xyz/elasticsearch-6-shard-replica-settings-for-single-node-cluster/ otherwise the health of the cluster will be yellow (not green :)).

I will create a separate issue for it.

pavolloffay:
I want to test it first.

PR #542, opened 2019-07-18
Run ES self provisioned e2e test on travis using OCP 3.9

### LINKED ISSUE #484: Run self es-self provisioned test on travis
Run `make e2e-tests-self-provisioned-es` on travis. The motivation is to catch issues early.

For example https://github.com/jaegertracing/jaeger-operator/issues/481 is not related to OCP version but to change in ES operator.

@kevinearls would you be interested in this issue?

pavolloffay:
Resolves #484

### DIFF
--- a/.travis/setupOpenShift.sh
+++ b/.travis/setupOpenShift.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -eoux pipefail
+
+export OPENSHIFT_VERSION=v3.9
+export KUBERNETES_VERSION=v1.14.0
+
+# setup insecure registry
+echo '{"registry-mirrors": ["https://mirror.gcr.io"], "mtu": 1460, "insecure-registries": ["172.30.0.0/16"] }' | sudo tee /etc/docker/daemon.json
+sudo service docker restart
+
+# install and run OCP
+sudo docker cp $(docker create docker.io/openshift/origin:$OPENSHIFT_VERSION):/bin/oc /usr/local/bin/oc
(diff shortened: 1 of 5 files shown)


deploy/role.yaml
```diff
@@ -43,6 +43,10 @@ rules:
 - apiGroups:
   - extensions
   resources:
+  - replicasets
```
pavolloffay: These are required to run on OCP 3.9

deploy/role.yaml
```diff
@@ -79,6 +83,7 @@ rules:
   - '*'
 - apiGroups:
   - apps
+  - extensions
```
pavolloffay: Required to run on OCP 3.9

pavolloffay:
travis is yellow, but when you click on it it is all green. Maybe travis reporting bug.

.travis/setupOpenShift.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -eoux pipefail
+
+export OPENSHIFT_VERSION=v3.9
```
jpkrohling: We should start with 4.1, and I think there's no `oc cluster up` for that.

pavolloffay: I don't think 4.1 will run on travis, it requires full virtualization and way more resources. This is the only way we can get it working here.

Other option is 3.11, but I prefer 3.9 due to a quick start and low resources utilization.

pavolloffay:
I have not seen issues with the streaming anymore, although it should not be related to this PR.

PR #545, opened 2019-07-22
Use random ports when port forwarding

kevinearls:

### DIFF
--- a/test/e2e/all_in_one_test.go
+++ b/test/e2e/all_in_one_test.go
@@ -166,7 +166,7 @@ func (suite *AllInOneTestSuite)  TestAllInOneWithUIConfig()  {
 	defer portForward.Close()
 	defer close(closeChan)
 
-	url := fmt.Sprintf("http://localhost:" + queryPort + "/%s/search", basePath)
+	url := fmt.Sprintf("http://localhost:%s/%s/search", queryPort, basePath)
 	c := http.Client{Timeout: time.Second}
 
 	req, err := http.NewRequest(http.MethodGet, url, nil)
--- a/test/e2e/all_in_one_test.go
+++ b/test/e2e/all_in_one_test.go
@@ -160,11 +160,13 @@ func (suite *AllInOneTestSuite)  TestAllInOneWithUIConfig()  {
 	require.NoError(t, err, "Error waiting for jaeger deployment")
 	defer undeployJaegerInstance(j)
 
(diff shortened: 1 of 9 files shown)


kevinearls:
@jpkrohling @pavolloffay @jpkrohling  Please review.  I'm hoping this will reduce test flakiness as we have had a lot of port forwarding errors.

test/e2e/utils.go
```diff
@@ -213,6 +215,15 @@ func undeployJaegerInstance(jaeger *v1.Jaeger) {
 	require.NoError(t, err)
 }
 
+func randomPortNumber() string {
```
pavolloffay: Random port implementations usually return a random port which is free. There is maybe a syscall which you could use to get the random port.

Note that k8s port-forward provides random port feature which I tried to use in https://github.com/jaegertracing/jaeger-operator/pull/288. To use it we would need to update k8s API to 1.15.

kevinearls: @pavolloffay  I've updated this so it should now return a random port which is free.  Can you review and merge if it looks ok?

pavolloffay: Could you please move this function to `portForward.go` and leave a TODO comment saying use port-forward from k8s once we upgrade k8s API to 1.15.0+

test/e2e/all_in_one_test.go
```diff
@@ -160,11 +160,13 @@ func (suite *AllInOneTestSuite)  TestAllInOneWithUIConfig()  {
...
 
-	portForward, closeChan := CreatePortForward(namespace, "all-in-one-with-ui-config", "jaegertracing/all-in-one", []string{"16686"}, fw.KubeConfig)
+	queryPort := randomPortNumber()
+	ports := []string{queryPort + ":16686"}
+	portForward, closeChan := CreatePortForward(namespace, "all-in-one-with-ui-config", "jaegertracing/all-in-one", ports, fw.KubeConfig)
 	defer portForward.Close()
 	defer close(closeChan)
 
-	url := fmt.Sprintf("http://localhost:16686/%s/search", basePath)
+	url := fmt.Sprintf("http://localhost:" + queryPort + "/%s/search", basePath)
```
pavolloffay: there is sprintf this change be changed to

```
fmt.Sprintf("http://localhost:%d"/%s/search", queryPort, basePath)
```

test/e2e/utils.go
```diff
@@ -1,8 +1,8 @@
 package e2e
 
 import (
-	"fmt"
 	goctx "context"
+	"fmt"
```
pavolloffay: @jpkrohling test package is not being lintedm `fmt` should be before `goctx`

PR #558, opened 2019-07-26
Check and update finalizers before setting APIVersion and Kind

### LINKED ISSUE #557: Invalid service account
When applying the simplest example with the current master, the following error appears:

```
ERRO[0003] failed to apply the changes                   error="ServiceAccount \"simplest\" is invalid: [metadata.ownerReferences.apiVersion: Invalid value: \"\": version must not be empty, metadata.ownerReferences.kind: Invalid value: \"\": kind must not be empty]" execution="2019-07-26 11:09:43.46854608 +0000 UTC" instance=simplest namespace=default
```

Running a `git bisect`, it appears that the problem is with this commit:
```
ca24c06682d526e8a1d76e78d6d66e7a4d4b1c33 is the first bad commit
commit ca24c06682d526e8a1d76e78d6d66e7a4d4b1c33
Author: Ruben Vargas Palma <ruben.vp8510@gmail.com>
Date:   Fri Jul 26 03:29:58 2019 -0500

    Remove sidecar when instance is deleted (#453)
    
:040000 040000 beedfe156f0fbe9b7df0e5edf07e040b2a1b3f4b 81f92598330ee1eac774c01c88537e8c07de7bda M	pkg
```

cc @rubenvp8510

rubenvp8510:
Fixes #557. This change is because handleFinalizer does an update (for add or remove the finalizer), so it overrides the APIVersion and Kind fields with empty strings.

### DIFF
--- a/pkg/controller/jaeger/jaeger_controller.go
+++ b/pkg/controller/jaeger/jaeger_controller.go
@@ -107,16 +107,15 @@ func (r *ReconcileJaeger) Reconcile(request reconcile.Request) (reconcile.Result
 		return reconcile.Result{}, err
 	}
 
-	// workaround for https://github.com/kubernetes-sigs/controller-runtime/issues/202
-	// see also: https://github.com/kubernetes-sigs/controller-runtime/pull/212
-	// once there's a version incorporating the PR above, the manual setting of the GKV can be removed
-	instance.APIVersion = fmt.Sprintf("%s/%s", v1.SchemeGroupVersion.Group, v1.SchemeGroupVersion.Version)
-	instance.Kind = "Jaeger"
 	if err = r.handleFinalizer(instance.GetDeletionTimestamp() != nil, instance); err != nil || instance.GetDeletionTimestamp() != nil {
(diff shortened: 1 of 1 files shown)


jpkrohling:
Interesting. Is that still because of https://github.com/kubernetes-sigs/controller-runtime/issues/202 ? If not, remove the comments from the code. Also, I just noticed an extra log entry when testing this:

```
INFO[0000] Versions                                      arch=amd64 jaeger-operator=v1.13.1-30-g67121958 operator-sdk=v0.8.1 os=linux version=go1.12.5
INFO[0010] adding Finalizer                              instance=simplest namespace=default
INFO[0010] Storage type not provided. Falling back to 'memory'  instance=simplest namespace=default
```

Could you please change the level of the `adding Finalizer` to debug?

rubenvp8510:
@jpkrohling I don't see that bug there, I ran the operator and saw that the values of APIVersion and Kind are there, it seems like that issue is already fixed, and I see that the issue was  closed with a fix.
 But not sure why the values disappeared after an update. I'll change the comment to point to this PR.

PR #573, opened 2019-07-31
Added htpasswd option to the OpenShift OAuth type

### LINKED ISSUE #572: Allow passthrough of `htpasswd` file location to the OAuth Proxy
Add an option for users to specify`Htpasswd` in the CR with the location of the htpasswd file to use, to be passed as-is to the OAuth Proxy.

jpkrohling:
Closes #572 by adding a new field to the oauth proxy struct.

### DIFF
--- a/deploy/examples/openshift/with-htpasswd.yaml
+++ b/deploy/examples/openshift/with-htpasswd.yaml
@@ -1,0 +1,18 @@
+# create the htpasswd with: htpasswd -cs /tmp/htpasswd jdoe
+# create the secret with: kubectl create secret generic htpasswd --from-file=htpasswd=/tmp/htpasswd
+apiVersion: jaegertracing.io/v1
+kind: Jaeger
+metadata:
+  name: with-htpasswd
+spec:
+  ingress:
+    openshift:
+      sar: '{"namespace": "default", "resource": "pods", "verb": "get"}'
+      htpasswdFile: /usr/local/data/htpasswd
+  volumeMounts:
+  - name: htpasswd-volume
+    mountPath: /usr/local/data
+  volumes:
+  - name: htpasswd-volume
+    secret:
+      secretName: htpasswd
--- a/pkg/apis/jaegertracing/v1/jaeger_types.go
+++ b/pkg/apis/jaegertracing/v1/jaeger_types.go
@@ -199,7 +199,10 @@ type JaegerIngressOpenShiftSpec struct {
 	SAR string `json:"sar,omitempty"`
 
 	// +optional
-	DelegateURLs string `json:"delegate-urls,omitempty"`
+	DelegateURLs string `json:"delegateUrls,omitempty"`
+
+	// +optional
+	HtpasswdFile string `json:"htpasswdFile,omitempty"`
 }
 
 // JaegerAllInOneSpec defines the options to be used when deploying the query
--- a/pkg/inject/oauth-proxy.go
+++ b/pkg/inject/oauth-proxy.go
@@ -44,6 +44,19 @@ func getOAuthProxyContainer(jaeger *v1.Jaeger) corev1.Container {
 		"--upstream=http://localhost:16686",
 	}
 
+	volumeMounts := []corev1.VolumeMount{{
+		MountPath: "/etc/tls/private",
+		Name:      service.GetTLSSecretNameForQueryService(jaeger),
+	}}
+
+	if len(jaeger.Spec.Ingress.OpenShift.HtpasswdFile) > 0 {
+		args = append(args, fmt.Sprintf("--htpasswd-file=%s", jaeger.Spec.Ingress.OpenShift.HtpasswdFile))
+		args = append(args, "--display-htpasswd-form=false")
+
+		// we can only get VolumeMounts from the top-level node
+		volumeMounts = append(volumeMounts, jaeger.Spec.JaegerCommonSpec.VolumeMounts...)
+	}
+
 	if len(jaeger.Spec.Ingress.OpenShift.SAR) > 0 {
 		args = append(args, fmt.Sprintf("--openshift-sar=%s", jaeger.Spec.Ingress.OpenShift.SAR))
 	}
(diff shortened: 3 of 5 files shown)


jpkrohling:
@jwendell, would you be able to confirm that this fixes the issue for you? I'm still running some manual tests on this, and we'd be ready to merge once you confirm it works, Travis approves and @objectiser reviews.

This is the image that can be used for the test: `jpkroehling/jaeger-operator:572-htpasswd-oauth-proxy`

jpkrohling:
@objectiser, I tested this and when `htpasswd` is enabled, I cannot login to the Jaeger UI with OpenShift users such as `kubeadmin`, only with users specified in the `htpasswd` file. @jwendell, is this the expected behavior?

I could also not verify whether bearer tokens were issued at all, perhaps this is something that @rubenvp8510 could check.

pkg/inject/oauth-proxy.go
```diff
@@ -56,14 +60,19 @@ func getOAuthProxyContainer(jaeger *v1.Jaeger) corev1.Container {
 
 	commonSpec := util.Merge([]v1.JaegerCommonSpec{jaeger.Spec.Ingress.JaegerCommonSpec, jaeger.Spec.JaegerCommonSpec})
 
+	volumeMounts := []corev1.VolumeMount{{
+		MountPath: "/etc/tls/private",
+		Name:      service.GetTLSSecretNameForQueryService(jaeger),
+	}}
+
+	// we can only get VolumeMounts from the top-level node
+	volumeMounts = append(volumeMounts, jaeger.Spec.JaegerCommonSpec.VolumeMounts...)
```
objectiser: Problem is that these mounts would be added to all components. The IngressSpec (or OpenShift specific node) probably needs to support its own volumes/mounts - and need to consider whether any other part of the `CommonSpec` is relevant - so should it be used or just the volume/mounts?

jpkrohling: I left it very simple at the moment, because I think we can discuss it more broadly. We should include storage-related objects in the mix as well, like cronjobs.

I changed the code slightly to mount only when needed, ie, when the `htpasswd-file` option is set.

objectiser: I was thinking more about the other components which will just mount the volumes listed at the top level :)

pkg/apis/jaegertracing/v1/jaeger_types.go
```diff
@@ -200,6 +200,9 @@ type JaegerIngressOpenShiftSpec struct {
 
 	// +optional
 	DelegateURLs string `json:"delegate-urls,omitempty"`
+
+	// +optional
+	HtpasswdFile string `json:"htpasswd-file,omitempty"`
```
objectiser: Just realised, this is using a hyphen (same for `delegate-urls`) whereas the convention we started to use was lower camel case.

jpkrohling:
I don't know how I missed the big blue button on the page, but it's indeed possible to login using `kubeadmin` (or other OpenShift users):

image

jpkrohling:
The newest image has just been published (version: `v1.13.1-40-gb8330536`): jpkroehling/jaeger-operator:572-htpasswd-oauth-proxy

deploy/examples/openshift/with-htpasswd.yaml
```diff
@@ -0,0 +1,18 @@
...
+# create the secret with: kubectl create secret generic htpasswd --from-file=htpasswd=/tmp/htpasswd
+apiVersion: jaegertracing.io/v1
+kind: Jaeger
+metadata:
+  name: with-htpasswd
+spec:
+  ingress:
+    openshift:
+      sar: '{"namespace": "default", "resource": "pods", "verb": "get"}'
+      htpasswd-file: /usr/local/data/htpasswd
```
objectiser: Need to update here aswell.

pkg/inject/oauth-proxy.go
jpkrohling: I see now what you mean. Yes, in order to have htpasswd, one would have to specify it at the top-level, which will cause all other components to mount the htpasswd secret. For this initial implementation, I think it's OK, but we do need to come up with a proper solution, also taking into consideration the other common spec values. Should we do this in a follow-up PR?

deploy/examples/openshift/with-htpasswd.yaml
jpkrohling: Done

jpkrohling:
For the record, this is how it looks like now:

image

pkg/inject/oauth-proxy.go
```diff
@@ -44,6 +44,19 @@ func getOAuthProxyContainer(jaeger *v1.Jaeger) corev1.Container {
...
 	}
 
+	volumeMounts := []corev1.VolumeMount{{
+		MountPath: "/etc/tls/private",
+		Name:      service.GetTLSSecretNameForQueryService(jaeger),
+	}}
+
+	if len(jaeger.Spec.Ingress.OpenShift.HtpasswdFile) > 0 {
+		args = append(args, fmt.Sprintf("--htpasswd-file=%s", jaeger.Spec.Ingress.OpenShift.HtpasswdFile))
+		args = append(args, "--display-htpasswd-form=false")
```
objectiser: Just to confirm - this doesn't affect the Jaeger UI login?

pkg/inject/oauth-proxy.go
objectiser: Yes a separate PR should be fine.

pkg/inject/oauth-proxy.go
jpkrohling: No, users would still need to click the "Login with OpenShift" button and will be directed to the login page hosted at OpenShift, which will redirect the users back to the Jaeger UI.

objectiser: Ok great.

pkg/inject/oauth-proxy.go
jpkrohling: #574

jwendell:
I confirm it works fine, thanks!

jpkrohling:
Alright, I just squashed the commits and updated the PR. Once Travis finishes, I'll merge it.

rubenvp8510:
> @objectiser, I tested this and when `htpasswd` is enabled, I cannot login to the Jaeger UI with OpenShift users such as `kubeadmin`, only with users specified in the `htpasswd` file. @jwendell, is this the expected behavior?
> 
> I could also not verify whether bearer tokens were issued at all, perhaps this is something that @rubenvp8510 could check.

In my tests I see that oauth-proxy store the bearer token encrypted in the cookie, but you need to pass the following  flags:

```
--pass-access-token=true
--pass-user-bearer-token=true
--pass-basic-auth=false
```
And a better secret.. (16 bytes )

You can see how I did my tests on https://github.com/rubenvp8510/jaeger/pull/1/commits/2f1ea29b40eb92db58493227a1c6c9e85beee341

I'm actually working on change it to a simpler test, but that worked for me to learn how the propagation works with openshift oauth-proxy

PR #580, opened 2019-08-05
Retry install

jpkrohling:
This attempts to get more stable build by retrying the network-bound commands, such as the ones that we use when installing the dependencies (SDK, tools, vendor).

It uses `travis_retry`, as documented here:
https://docs.travis-ci.com/user/common-build-problems/#timeouts-installing-dependencies


jpkrohling:
Apparently, the retry function works :-)

image

kevinearls:
@jpkrohling Cool!

jpkrohling:
Not perfect yet, as I did have to restart one job, but it's far better now.

PR #582, opened 2019-08-07
Initial tests for the examples

kevinearls:

This is the initial set of tests for #144   It covers everything but:

all-in-one-with-options.yaml
business-application-injected-sidecar.yaml
custom-serviceaccount.yaml
disable-ingress.yaml
openshift/custom-sar-oauth-proxy.yaml
openshift/disable-oauth-proxy.yaml
openshift/with-httpasswd.yaml

### DIFF
--- a/test/e2e/utils.go
+++ b/test/e2e/utils.go
@@ -1,8 +1,9 @@
 package e2e
 
 import (
-	"fmt"
 	goctx "context"
+	"fmt"
+	"net/http"
 	"os"
 	"runtime"
 	"strings"
@@ -18,6 +19,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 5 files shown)


kevinearls:
@pavolloffay @objectiser @jpkrohling Please review.  This is the initial set of tests, I will add more when I have a bit more free time.  I setup the tests to run on Travis, but at the moment they take quite a bit longer than the next slowest set of tests (16:46 v. 12:30 for cassandra tests.)  Do you think we need to run these on a per PR basis?

test/e2e/utils.go
```diff
@@ -213,6 +216,31 @@ func undeployJaegerInstance(jaeger *v1.Jaeger) {
...
+	jaegerInstance := &v1.Jaeger{}
+	key := types.NamespacedName{Name: name, Namespace: namespace}
+	err := fw.Client.Get(goctx.Background(), key, jaegerInstance)
+	require.NoError(t, err)
+	return jaegerInstance
+}
+
+type ValidateHttpResponseFunc func(response *http.Response) (done bool, err error)
+
+func WaitAndPollForHttpResponse(targetUrl string, condition ValidateHttpResponseFunc) (err error) {
```
jpkrohling: If you aren't specifying anything else for this type, you can just use the `func` definition as you had there as the parameter to this func. Like:

```go
 func WaitAndPollForHttpResponse(targetUrl string, condition func(response *http.Response) (done bool, err error)) (err error) {
```

jpkrohling: On a second thought, I think your way is easier to read... Don't listen to me.

kevinearls: :-)  I based it on this https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L202-L204 which is also not strictly necessary, but makes things a bit clearer.

kevinearls:
@jpkrohling can you merge this, as Gary is out?

pavolloffay:
in https://github.com/jaegertracing/jaeger-operator/pull/587 I had to restart travis job for examples as it failed.

PR #612, opened 2019-08-22
Move self provisioned ES example to Openshift subfolder

pavolloffay:


pavolloffay:
The file is not mentioned on jaeger website.

PR #617, opened 2019-08-23
Added github workflow

jpkrohling:


jpkrohling:
This will probably break Travis, due to the changes on how `minikube`/`kubectl` are made available.

jpkrohling:
Before removing Travis, I'd like to get the OK from @objectiser and @pavolloffay as well. If it all looks good, I'll switch it off first thing Monday.

jpkrohling:
Good catch, looks like it's not being executed:

> ok  	github.com/jaegertracing/jaeger-operator/test/e2e	0.184s

I'll fix this one.

jpkrohling:
@pavolloffay it's now fixed.

.github/workflows/e2e-openshift.yaml
```diff
@@ -0,0 +1,31 @@
...
+on: [push, pull_request]
+
+jobs:
+  end-to-end:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        TEST_GROUP: [es-self-provisioned]
+    steps:
+    - uses: jpkrohling/setup-kubectl@v1-release
```
pavolloffay: What is this uses? Why does it reference your account?

jpkrohling: This downloads and stores kubectl in the local GitHub cache. It's under my personal account because it doesn't really belong to Jaeger.

jpkrohling:
> Does GH have secrets to perform the release? E.g. push to dockerhub/quay?

I think it does, but I have to do some tests first.

PR #620, opened 2019-08-27
Add resource limits for spark dependencies cronjob

### LINKED ISSUE #598: Resource limits for spark dependencies cronjob
The [spark dependencies cronjob](https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/cronjob/spark_dependencies.go#L36) does not have resources populated. It would be nice to be able to specify requests and limits for the pods created by this cronjob.
Right now the pods created by this cronjob do not have any resources defined and they consume a substantial amount of CPU / memory causing resource starvation on the node and all pods on it.

rubenvp8510:
Fixes #598

### DIFF
--- a/pkg/cronjob/spark_dependencies.go
+++ b/pkg/cronjob/spark_dependencies.go
@@ -12,6 +12,7 @@ import (
 
 	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
 	"github.com/jaegertracing/jaeger-operator/pkg/storage"
+	"github.com/jaegertracing/jaeger-operator/pkg/util"
 )
 
 var supportedStorageTypes = map[string]bool{"elasticsearch": true, "cassandra": true}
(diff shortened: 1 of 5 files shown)


jpkrohling:
One of the e2e tests for Kubernetes failed, so I just restarted it. Almost immediately, I saw this in the logs:

```
git checkout --progress --force 76143e9bd1c40fc0b71fd45a987c80b9db0a9096
##[error]fatal: reference is not a tree: 76143e9bd1c40fc0b71fd45a987c80b9db0a9096
Removed matchers: 'checkout-git'
##[remove-matcher owner=checkout-git]
##[error]Git checkout failed with exit code: 128
##[error]Exit code 1 returned from process: file name '/home/runner/runners/2.157.0/bin/Runner.PluginHost', arguments 'action "GitHub.Runner.Plugins.Repository.CheckoutTask, Runner.Plugins"'.
```

Not sure what's going on, but could you try to rebase and update this PR?

jpkrohling:
Opened https://github.com/actions/checkout/issues/23 to track this failure.

pkg/cronjob/spark_dependencies.go
```diff
@@ -68,7 +68,8 @@ func CreateSparkDependencies(jaeger *v1.Jaeger) *batchv1beta1.CronJob {
 									Image: jaeger.Spec.Storage.Dependencies.Image,
 									Name:  name,
 									// let spark job use its default values
-									Env: removeEmptyVars(envVars),
+									Env:       removeEmptyVars(envVars),
+									Resources: jaeger.Spec.Storage.Dependencies.Resources,
```
pavolloffay: I think other components by default use the resources specified at the root level of the spec. We should consider doing the same here.

palmerabollo:
@rubenvp8510 could you please share an example of how a Jaeger resource would look like (yaml)? I am not sure about how the golang `ResourceRequirements` maps to the yaml. I am particularly interested in increasing the request/limits of the jaeger-spark-dependencies pods. Thank you.

PR #624, opened 2019-08-27
Update streaming test to download CRs, use version 0.12.1

kevinearls:

This also makes it possible to test with other kafka streams operators, like AMQ streams.


kevinearls:
@objectiser @jpkrohling  @pavolloffay  Please review

Makefile
```diff
@@ -182,8 +184,10 @@ storage:
 kafka:
 	@echo Creating namespace $(KAFKA_NAMESPACE)
 	@kubectl create namespace $(KAFKA_NAMESPACE) 2>&1 | grep -v "already exists" || true
-	@sed 's/namespace: .*/namespace: kafka/' ./test/kafka-operator.yml | kubectl -n $(KAFKA_NAMESPACE) apply -f -  2>&1 | grep -v "already exists" || true
-	@kubectl apply -f ./test/kafka.yml -n $(KAFKA_NAMESPACE) 2>&1 | grep -v "already exists" || true
+	@curl --location ${KAFKA_YAML} --output ./test/kafka-operator.yml
+	@curl --location ${KAFKA_EXAMPLE} --output ./test/kafka-example.yaml
```
jpkrohling: I see that those files were added to this PR. Is that intentional? If not, this should be added to the gitignore, or piped directly to kubectl, like:

```console
curl ${KAFKA_YAML} | sed ... | kubectl apply -n ... -f -
```

Perhaps this would be better suited for a new script, `./.ci/setup-kafka.sh`?

pavolloffay: +1 for a direct pipe

Makefile
```diff
@@ -15,6 +15,8 @@ JAEGER_VERSION ?= "$(shell grep -v '\#' jaeger.version)"
 OPERATOR_VERSION ?= "$(shell git describe --tags)"
 STORAGE_NAMESPACE ?= "${shell kubectl get sa default -o jsonpath='{.metadata.namespace}' || oc project -q}"
 KAFKA_NAMESPACE ?= "kafka"
+KAFKA_YAML ?= "https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.12.1/strimzi-cluster-operator-0.12.1.yaml"
```
pavolloffay: Define the version here rather than a direct yaml.

PR #628, opened 2019-08-29
Add Jaeger version to Elasticsearch job images

pavolloffay:
Resolves https://issues.jboss.org/browse/TRACING-682

### DIFF
--- a/pkg/strategy/controller.go
+++ b/pkg/strategy/controller.go
@@ -3,6 +3,7 @@ package strategy
 import (
 	"context"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


pavolloffay:
The image normalization in other components works slightly differently:

https://github.com/jaegertracing/jaeger-operator/blob/60f5ed6c2e46fba3cb32c2089e1d3158f2cf7c8b/pkg/deployment/all-in-one.go#L77-L80

In all-in-one the image field is not set, whereas this will override it in the CR.

PR #641, opened 2019-09-04
Update changelog for v1.14.0

jpkrohling:


jpkrohling:
There's a breaking change in Jaeger 1.14, but I'm not sure there's something we can do in the Jaeger Operator side. @pavolloffay, could you confirm whether `noop` is a valid upgrade step from 1.13.1 to 1.14.0? This is what I see in the Jaeger's changelog:

    Create ES index templates instead of indices (#1627, @pavolloffay)

    This can break existing Elasticsearch deployments if security policies are applied.
    For instance Jaeger X-Pack configuration now requires permission to create index templates - manage_index_templates.

The first commit in this PR includes all the changes between the versions. The second commit is my suggestion for the changelog. Let me know if anything from the first commit should be reinstated.

Issue #605, opened 2019-08-21
Remove legacy CRD


jpkrohling:
It's now been a while since we deprecated the CRD in the old namespace `io.jaegertracing`, I think it's time to remove it from the source repository and code.

jan25:
@jpkrohling Hi! would you know if someone is working on this? I can take a look otherwise

jpkrohling:
It's yours, @jan25!

PR #649, opened 2019-09-07
Breaking change - removed legacy io.jaegertracing CRD

jan25:
Resolves https://github.com/jaegertracing/jaeger-operator/issues/605


jpkrohling:
Thanks @jan25! I rebased your PR, but it looks good to me. If there's still something missing, we can remove it at a later point. I would just ask you to amend and reword your PR, to include the word "breaking change" in the title. Suggestion:

> Breaking change - removed io.jaegertracing CRD

This way, it's easy to spot in the changelog.

jan25:
Thanks! makes sense

PR #653, opened 2019-09-09
Improved the install-sdk target

jpkrohling:

### DIFF
--- a/.ci/install-sdk.sh
+++ b/.ci/install-sdk.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+DEST="${GOPATH}/bin/operator-sdk"
+
+function install_sdk() {
+    echo "Downloading the operator-sdk ${SDK_VERSION} into ${DEST}"
+    curl https://github.com/operator-framework/operator-sdk/releases/download/${SDK_VERSION}/operator-sdk-${SDK_VERSION}-x86_64-linux-gnu -sLo ${DEST}
+    chmod +x ${DEST}
+}
+
+mkdir -p ${GOPATH}/bin
+
+if [ ! -f ${DEST} ]; then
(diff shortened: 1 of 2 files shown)


jpkrohling:
This will now skip the download of the SDK if the same version already exists at the destination.

.ci/install-sdk.sh
```diff
@@ -0,0 +1,22 @@
...
+
+DEST="${GOPATH}/bin/operator-sdk"
+
+function install_sdk() {
+    echo "Downloading the operator-sdk ${SDK_VERSION} into ${DEST}"
+    curl https://github.com/operator-framework/operator-sdk/releases/download/${SDK_VERSION}/operator-sdk-${SDK_VERSION}-x86_64-linux-gnu -sLo ${DEST}
+    chmod +x ${DEST}
+}
+
+if [ ! -d ${GOPATH}/bin/ ]; then
```
pavolloffay: `mkdir -p` is noop and does not fail if the dir exists. The check does not seem to be necessary.

.ci/install-sdk.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+DEST="${GOPATH}/bin/operator-sdk"
+
```
pavolloffay: Shouldn't it also define the default for `SDK_VERSION`? This way the script would work without the caller knowing the configuration options.

jpkrohling: If we ever have someone to use this script manually, we can change it. Right now, it's easiest to just use the same target as before: `make install-sdk`

pavolloffay: fair enough. Defining all the prerequisites helps with portability and readability. But it's just a ci script.

PR #677, opened 2019-09-26
Add hostnetworking & dnspolicy to agent daemonset

### LINKED ISSUE #676: Agent daemonset should be started with hostNetwork
When using the `daemonset` stragegy for the agent, the agent pods should be created with `hostNetwork: true` to make it possible for tracing clients to reference their IP using the following container spec:
```
env
- name: JAEGER_AGENT_HOST
          valueFrom:
            fieldRef:
              fieldPath: status.hostIP
```

`hostNetwork` is enabled in the kubernetes production template too, so this looks like a small  oversight: https://github.com/jaegertracing/jaeger-kubernetes/blob/master/jaeger-production-template.yml#L217

sebbonnet:
These 2 properties are available in the jaeger production template,
but missing when the daemonset is created via the operator

Fixes #676

### DIFF
--- a/pkg/deployment/agent.go
+++ b/pkg/deployment/agent.go
@@ -151,6 +151,8 @@ func (a *Agent) Get() *appsv1.DaemonSet {
 						},
 						Resources: commonSpec.Resources,
 					}},
+					HostNetwork:        true,
+					DNSPolicy:          corev1.DNSClusterFirstWithHostNet,
 					Affinity:           commonSpec.Affinity,
 					Tolerations:        commonSpec.Tolerations,
 					SecurityContext:    commonSpec.SecurityContext,
--- a/pkg/deployment/agent_test.go
+++ b/pkg/deployment/agent_test.go
@@ -147,6 +147,24 @@ func TestDaemonSetAgentResources(t *testing.T) {
 	assert.Equal(t, *resource.NewQuantity(512, resource.DecimalSI), dep.Spec.Template.Spec.Containers[0].Resources.Requests[corev1.ResourceRequestsEphemeralStorage])
 }
 
+func TestDaemonSetAgentHostNetwork(t *testing.T) {
+	jaeger := v1.NewJaeger(types.NamespacedName{Name: "TestDaemonSetAgentHostNetwork"})
+	jaeger.Spec.Agent.Strategy = "daemonset"
+	agent := NewAgent(jaeger)
+	dep := agent.Get()
+
+	assert.Equal(t, true, dep.Spec.Template.Spec.HostNetwork)
+}
+
+func TestDaemonSetAgentDnsPolicy(t *testing.T) {
+	jaeger := v1.NewJaeger(types.NamespacedName{Name: "TestDaemonSetAgentDnsPolicy"})
+	jaeger.Spec.Agent.Strategy = "daemonset"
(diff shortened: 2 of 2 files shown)


jpkrohling:
@kevinearls, @jkandasa could one of you please test this in OpenShift? I'm not sure why we had this in the prod template for Kubernetes, but we are missing here.

jpkrohling:
@sebbonnet It looks like this would cause problems when the operator is running in OpenShift. @kevinearls ran some tests and it failed intermittently with the following error message:

```
Error creating: pods "agent-as-daemonset-agent-daemonset-" is forbidden: unable to validate against any security context constraint: [provider anyuid: .spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.containers[0].securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec
```

I'm wondering if the same error would happen in other Kubernetes platforms out there, like Google's or Amazon's. Have you tried there?

I think the best way to move forward is to have something like:

1. a Viper configuration entry like `HostNetworkEnabled`
1. an auto-detect function that sets this flag -- I recommend having it disabled by default for now. Reference: https://github.com/jaegertracing/jaeger-operator/blob/54f74a5bececad40024a501ee596824edbe43828/pkg/autodetect/main.go#L71-L86
1. in the agent daemonset, check if the flag is enabled and only then set the values from this PR

sebbonnet:
I've not tested on any managed kubernetes cluster like GKE or EKS, but I don't think they have any particular restrictions around host networking. 
I've tested the changes with both a vanilla kind cluster and a Kubernetes cluster that we manage in AWS.

I've not used OpenShift before so I would need to read about how their approach to priviledged containers differs from Kubernetes and try it out.
Having said that I'm not sure how the agent as daemonset approach would work without host networking, as client apps would have no easy way to find the pod agent to connect to. I suppose they could use a custom sidecar to detect the agent pod running on the same node.
It seems there used to be a set of templates for Openshift [here](https://github.com/jaegertracing/jaeger-openshift), which suggests agents as sidecar was the only possible option - however this site is deprecated, so I'm not sure if that is of any relevance.

jpkrohling:
> as client apps would have no easy way to find the pod agent to connect to.

Client apps can be configured via downward API to connect to a nodePort, as described in the documentation: https://www.jaegertracing.io/docs/1.14/operator/#installing-the-operator

To use `DaemonSet` on OpenShift, a special permission is required and there are some guards in the current operator code about it.

jpkrohling:
@sebbonnet, are you interested in continue working on this one?

KielChan:
I wanna continue to work on it. But, what I want commit is that Add DNSPolicy and HostNetwork in `JaegerCommonSpec`, it will affect agent, all_in_one, collector, es_* ,ingester, query and so on.

jpkrohling:
@KielChan what's your use case for that?

KielChan:
> @KielChan what's your use case for that?

It's so complex that ES acquired ip resource from cilium, has a domain register into external DNS to point to istio-gateway.  It cannot access by pod with ClusterFirst DNSPolicy, 'cause domain does not register into kube-dns or coredns.

jpkrohling:
I'm closing this one, as I'm not sure you want to continue working on it, @sebbonnet.

PR #681, opened 2019-10-02
Add support for TLS on ingress

### LINKED ISSUE #85: Add support for TLS on ingress
https://kubernetes.io/docs/concepts/services-networking/ingress/#tls

cedrickring:
Fixes #85.

I didn't add the auto lookup of an existing secret like `<instance-name>-tls-secret`. This would require the client to be accessible in `ingress/query.go`.


jpkrohling:

PR #689, opened 2019-10-10
Add a streaming test which uses TLS

kevinearls:

### DIFF
--- a/test/e2e/streaming_test.go
+++ b/test/e2e/streaming_test.go
@@ -89,6 +89,7 @@ func (suite *StreamingTestSuite) TestStreamingWithTLS() {
 	kafkaUser := getKafkaUser(kafkaUserName, kafkaNamespace)
 	err := fw.Client.Create(context.Background(), kafkaUser, &framework.CleanupOptions{TestContext: ctx, Timeout: timeout, RetryInterval: retryInterval})
 	require.NoError(t, err, "Error deploying kafkauser")
+	WaitForSecret(kafkaUserName)
 
 	defer func() {
 		if !debugMode || !t.Failed() {
--- a/test/e2e/streaming_test.go
+++ b/test/e2e/streaming_test.go
@@ -89,7 +89,7 @@ func (suite *StreamingTestSuite) TestStreamingWithTLS() {
 	kafkaUser := getKafkaUser(kafkaUserName, kafkaNamespace)
 	err := fw.Client.Create(context.Background(), kafkaUser, &framework.CleanupOptions{TestContext: ctx, Timeout: timeout, RetryInterval: retryInterval})
 	require.NoError(t, err, "Error deploying kafkauser")
-	WaitForSecret(kafkaUserName)
+	WaitForSecret(kafkaUserName, kafkaNamespace)
 
 	defer func() {
 		if !debugMode || !t.Failed() {
--- a/test/e2e/streaming_test.go
+++ b/test/e2e/streaming_test.go
@@ -87,7 +87,7 @@ func (suite *StreamingTestSuite) TestStreamingWithTLS() {
 
 	kafkaUserName := "my-user"
 	kafkaUser := getKafkaUser(kafkaUserName, kafkaNamespace)
-	err := fw.Client.Create(context.TODO(), kafkaUser, &framework.CleanupOptions{TestContext: ctx, Timeout: timeout, RetryInterval: retryInterval})
+	err := fw.Client.Create(context.Background(), kafkaUser, &framework.CleanupOptions{TestContext: ctx, Timeout: timeout, RetryInterval: retryInterval})
 	require.NoError(t, err, "Error deploying kafkauser")
 
 	defer func() {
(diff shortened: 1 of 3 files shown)


kevinearls:
@objectiser  @pavolloffay @jpkrohling  Please review

test/e2e/streaming_test.go
```diff
@@ -78,6 +77,45 @@ func (suite *StreamingTestSuite) TestStreaming() {
...
+	if !usingOLM {
+		t.Skip("This test should only run when using OLM")
+	}
+	// Make sure ES and the kafka instance are available
+	waitForElasticSearch()
+	waitForKafkaInstance()
+
+	kafkaUserName := "my-user"
+	kafkaUser := getKafkaUser(kafkaUserName, kafkaNamespace)
+	err := fw.Client.Create(context.TODO(), kafkaUser, &framework.CleanupOptions{TestContext: ctx, Timeout: timeout, RetryInterval: retryInterval})
```
jpkrohling: `TODO()` implies that it will change in the future to an appropriate value. As we don't have a context that we could use, the most appropriate here would be `context.Background()`.

test/e2e/streaming_test.go
```diff
@@ -113,3 +151,134 @@ func jaegerStreamingDefinition(namespace string, name string) *v1.Jaeger {
 	}
 	return j
 }
+
+func jaegerStreamingDefinitionWithTLS(namespace string, name, kafkaUserName string) *v1.Jaeger {
+	volumes := getTlsVolumes(kafkaUserName)
```
jpkrohling: nit: `TLSVolumes` (https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

test/e2e/streaming_test.go
```diff
@@ -113,3 +151,134 @@ func jaegerStreamingDefinition(namespace string, name string) *v1.Jaeger {
...
+					"type": "tls",
+				},
+			},
+		},
+	}
+
+	return kafkaUser
+}
+
+func getTlsVolumeMounts() []corev1.VolumeMount {
```
jpkrohling: `s/Tls/TLS/`

test/e2e/streaming_test.go
```diff
@@ -113,3 +151,134 @@ func jaegerStreamingDefinition(namespace string, name string) *v1.Jaeger {
...
+	}
+
+	volumeMounts := []corev1.VolumeMount{
+		kafkaUserVolumeMount, clusterCaVolumeMount,
+	}
+
+	return volumeMounts
+}
+
+func getTlsVolumes(kafkaUserName string) []corev1.Volume {
```
jpkrohling: `s/Tls/TLS`

test/e2e/streaming_test.go
```diff
@@ -78,6 +77,45 @@ func (suite *StreamingTestSuite) TestStreaming() {
...
+func (suite *StreamingTestSuite) TestStreamingWithTLS() {
+	if !usingOLM {
+		t.Skip("This test should only run when using OLM")
+	}
+	// Make sure ES and the kafka instance are available
+	waitForElasticSearch()
+	waitForKafkaInstance()
+
+	kafkaUserName := "my-user"
+	kafkaUser := getKafkaUser(kafkaUserName, kafkaNamespace)
```
jpkrohling: Does it still work if you move this to before the call to `waitForKafkaInstance()` ? Reason being that there's a (small) chance that the `KafkaUser` resource might not have been created (or reconciled) yet by the time it's secret is used by the Jaeger instance. If you can move it to before the Kafka Instance, then the chances are really, really small, as it has all the time it takes for a Kafka instance to boot to create the resource and secrets.

kevinearls: It would most likely work, since this is the second test, and the kafka instance would have to be ready for the first test to have even run.  :-)  So I could move it, but I doubt that would make much difference.

Is there anyway to confirm programmatically that all of the user resources are ready?  If not I'd rather just add a short sleep after creating the user.

jpkrohling: When you create a KafkaUser, Strimzi will create the secret that we require on the next reconciliation. So, you can just check whether the secret has been created already.

Or, you can just do "nothing" and wait for the deployment to eventually succeed, as it will not be created until all the requirements (like secrets) are available. The disadvantage of that is, if the test fails, you'll get a better error message if you are checking for the secret's existence.

test/e2e/wait_util.go
```diff
@@ -192,3 +193,20 @@ func WaitForDeployment(t *testing.T, kubeclient kubernetes.Interface, namespace,
 	logrus.Infof("Deployment of %s in namespace %s took %s\n", name, namespace, elapsed)
 	return err
 }
+
+// WaitForSecret waits for a secret to be available
+func WaitForSecret(secretName string) {
+	err := wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		secretList, err := fw.KubeClient.CoreV1().Secrets(kafkaNamespace).List(metav1.ListOptions{IncludeUninitialized: false})
```
jpkrohling: `kafkaNamespace`? Is it holding the namespace for this test, or is it really the `kafkaNamespace`?

jpkrohling: Also, I think you can use a `Get` instead of a `List`, no? You know the object's name after all...

PR #695, opened 2019-10-14
Updated Operator SDK to v0.11.0

### LINKED ISSUE #692: Update Operator SDK to v0.11.0
The Operator SDK 0.11.0 was released, fixing some bugs that were affecting us. This issue is to track the update to this new version.

jpkrohling:
Closes #692.


build/bin/entrypoint
```diff
@@ -5,7 +5,7 @@
 
 if ! whoami &>/dev/null; then
   if [ -w /etc/passwd ]; then
-    echo "${USER_NAME:-qaclana}:x:$(id -u):$(id -g):${USER_NAME:-qaclana} user:${HOME}:/sbin/nologin" >> /etc/passwd
+    echo "${USER_NAME:-jaeger-operator}:x:$(id -u):$(id -g):${USER_NAME:-jaeger-operator} user:${HOME}:/sbin/nologin" >> /etc/passwd
```
jpkrohling: This is there since the very first commit, I suppose. In my initial PoC of the operator, I used a pet project (`qaclana`) to learn about the operator, and it looks like this was left behind when I started this operator here for real.

README.md
```diff
@@ -10,7 +10,7 @@ The Jaeger Operator is an implementation of a [Kubernetes Operator](https://kube
 To install the operator, run:
 ```
 kubectl create namespace observability
-kubectl create -f https://raw.githubusercontent.com/jaegertracing/jaeger-operator/master/deploy/crds/jaegertracing_v1_jaeger_crd.yaml
+kubectl create -f https://raw.githubusercontent.com/jaegertracing/jaeger-operator/master/deploy/crds/jaegertracing.io_jaegers_crd.yaml
```
pavolloffay: This will probably break clients/tutorials.  Why is this needed?

pkg/autodetect/main.go
```diff
@@ -216,3 +216,18 @@ func isElasticsearchOperatorAvailable(apiList *metav1.APIGroupList) bool {
 	}
 	return false
 }
+
+type matchingLabelKeys map[string]string
+
+func (m matchingLabelKeys) ApplyToList(opts *client.ListOptions) {
```
pavolloffay: Isn't cleaner to return `labelSelector` than modifying the struct?

README.md
jpkrohling: We could probably keep the old one there. Unfortunately, this is a breaking change from the Operator SDK itself and some commands from it expect the file to be in this new location.

jpkrohling:
>  The CI should use makefile to install the SDK to keep it consistent.

It's build this way to make use of the caching mechanism from the CI.

pkg/autodetect/main.go
jpkrohling: Would have been, but they changed the API and I couldn't find a cleaner way to do it. See https://github.com/operator-framework/operator-sdk/issues/2060

jpkrohling:
PR updated:

1. keep the old CRD file (with a comment)
1. fix the Operator SDK version in the Makefile

deploy/crds/jaegertracing_v1_jaeger_crd.yaml
```diff
@@ -1,3 +1,6 @@
+# ATTENTION: this file is kept here for backwards compatibility reasons
+# use the new location: jaegertracing.io_jaegers_crd.yaml . If there are
+# inconsistencies between the files, this one is considered wrong.
```
pavolloffay: +1 it would be cool to print deprecation message when it is used. Not sure if that is possible though.

pkg/autodetect/main_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/googleapis/gnostic/OpenAPIv2"
+	openapi_v2 "github.com/googleapis/gnostic/OpenAPIv2"
```
pavolloffay: nit: I think we don't use underscores in import names

jpkrohling: I think this came from the `make fmt`, which puts the package name from the source file instead of the path.

https://github.com/googleapis/gnostic/blob/master/OpenAPIv2/OpenAPIv2.go#L17

PR #704, opened 2019-10-17
New DeploymentStrategy type for JaegerSpec.Strategy 

### LINKED ISSUE #137: Make JaegerSpec.Strategy a type
Strategy is defined as string, it makes it harder to find out what values are correct. It can be defined as a separate type e.g. `type DeploymentStrategy string` and valid values can be defined as constants.

volmedo:
This PR adds a new type, `DeploymentStrategy`, to be used in `JaegerSpec.Strategy` instead of just a `string`. Constants based on the new type are also defined to define valid values.

When merged, this PR will close #137. 

### To do
- [x] make sure all unserialized values are converted to lower case
- [x] a test ensuring that current CRs with a mixed-case strategy are correctly parsed
- [x] the constant in the strategy package, mentioned in #137

### DIFF
--- a/pkg/apis/jaegertracing/v1/deployment_strategy.go
+++ b/pkg/apis/jaegertracing/v1/deployment_strategy.go
@@ -1,0 +1,49 @@
+package v1
+
+import (
+	"errors"
+	"strings"
+)
+
+// DeploymentStrategy represents the possible values for deployment strategies
+// +k8s:openapi-gen=true
+type DeploymentStrategy string
+
+const (
+	// DeploymentStrategyDeprecatedAllInOne represents the (deprecated) 'all-in-one' deployment strategy
+	// +k8s:openapi-gen=true
+	DeploymentStrategyDeprecatedAllInOne DeploymentStrategy = "all-in-one"
+
+	// DeploymentStrategyAllInOne represents the 'allInOne' deployment strategy (default)
+	// +k8s:openapi-gen=true
+	DeploymentStrategyAllInOne DeploymentStrategy = "allinone"
+
+	// DeploymentStrategyStreaming represents the 'streaming' deployment strategy
+	// +k8s:openapi-gen=true
+	DeploymentStrategyStreaming DeploymentStrategy = "streaming"
+
+	// DeploymentStrategyProduction represents the 'production' deployment strategy
+	// +k8s:openapi-gen=true
+	DeploymentStrategyProduction DeploymentStrategy = "production"
+)
+
+// UnmarshalText implements encoding.TextUnmarshaler to ensure that JSON values in the
+// strategy field of JSON jaeger specs are interpreted in a case-insensitive manner
+func (ds *DeploymentStrategy) UnmarshalText(text []byte) error {
+	if ds == nil {
+		return errors.New("DeploymentStrategy: UnmarshalText on nil pointer")
+	}
+
+	switch strings.ToLower(string(text)) {
+	default:
+		*ds = DeploymentStrategyAllInOne
+	case string(DeploymentStrategyDeprecatedAllInOne):
+		*ds = DeploymentStrategyDeprecatedAllInOne
+	case string(DeploymentStrategyStreaming):
+		*ds = DeploymentStrategyStreaming
+	case string(DeploymentStrategyProduction):
+		*ds = DeploymentStrategyProduction
+	}
+
+	return nil
+}
--- a/pkg/apis/jaegertracing/v1/jaeger_types.go
+++ b/pkg/apis/jaegertracing/v1/jaeger_types.go
@@ -53,6 +53,24 @@ const (
 // +k8s:openapi-gen=true
 type DeploymentStrategy string
 
+const (
+	// DeploymentStrategyDeprecatedAllInOne
+	// +k8s:openapi-gen=true
+	DeploymentStrategyDeprecatedAllInOne DeploymentStrategy = "all-in-one"
+
+	// DeploymentStrategyAllInOne
+	// +k8s:openapi-gen=true
+	DeploymentStrategyAllInOne DeploymentStrategy = "allinone"
+
+	// DeploymentStrategyStreaming
+	// +k8s:openapi-gen=true
+	DeploymentStrategyStreaming DeploymentStrategy = "streaming"
+
+	// DeploymentStrategyProduction
+	// +k8s:openapi-gen=true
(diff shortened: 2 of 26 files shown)


pkg/deployment/collector.go
```diff
@@ -76,9 +75,9 @@ func (c *Collector) Get() *appsv1.Deployment {
 	}
 
 	storageType := c.jaeger.Spec.Storage.Type
-	// If strategy is "streaming", then change storage type
+	// If strategy is DeploymentStrategyStreaming, then change storage type
 	// to Kafka, and the storage options will be used in the Ingester instead
-	if strings.EqualFold(c.jaeger.Spec.Strategy, "streaming") {
+	if c.jaeger.Spec.Strategy == v1.DeploymentStrategyStreaming {
```
jpkrohling: What would happen if we have a CR with a strategy set to `Streaming`?

pkg/apis/jaegertracing/v1/jaeger_types.go
```diff
@@ -49,11 +49,33 @@ const (
 	IngressSecurityOAuthProxy IngressSecurityType = "oauth-proxy"
 )
 
+// DeploymentStrategy represents the possible values for deployment strategies
+// +k8s:openapi-gen=true
+type DeploymentStrategy string
```
jpkrohling: You could implement the UnmarshalJSON method to the type, to ensure the values read from JSON are converted to lowercase.

pkg/deployment/collector.go
volmedo: I was assuming that the value was being normalized somewhere else. Sorry, my bad.

volmedo:
> 1. make sure all unserialized values are converted to lower case

Where is the unmarshalling being done? I suppose it is done by some library from k8s or the operator framework. If that is the case, is it enough to make the new type implement the `Unmarshaler` interface?

> 2. possibly a test ensuring that current CRs with a mixed-case strategy are correctly parsed

Where should this test live? Are there any other similar tests?

jpkrohling:
> Where is the unmarshalling being done?

It's done by the Operator SDK. It's sufficient to have a couple of methods, like these from the `Options` object:

https://github.com/jaegertracing/jaeger-operator/blob/d11973f0dbe9b7b602da1f1214e21d40c4a5e8dd/pkg/apis/jaegertracing/v1/options.go#L41-L62

I *think* you only need to implement the `UnmarshalJSON`, but needs confirmation. 

We have some tests for the same object that could be used as reference:

https://github.com/jaegertracing/jaeger-operator/blob/d11973f0dbe9b7b602da1f1214e21d40c4a5e8dd/pkg/apis/jaegertracing/v1/options_test.go#L11-L16

pkg/strategy/strategy.go
```diff
@@ -9,12 +9,13 @@ import (
 	"k8s.io/api/extensions/v1beta1"
 	rbac "k8s.io/api/rbac/v1"
 
+	jv1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
```
jpkrohling: We typically use `v1` for Jaeger, and `corev1` for `k8s.io/api/core/v1`

volmedo: Makes a lot of sense!

I saw that `v1` was already in use and invented a different alias for Jaeger. Didn't think about the possibility of renaming the existing one :). Will do it now.

volmedo:
> Sorry, let me take back my approval :-) Just realized that it's still missing the serialization tests, to ensure backwards compatibility (which is probably why this is still WIP).

Yes, it's WIP because of that. I added three items to a todo list for the PR and I began with the last one instead of the first one :).

volmedo:
I added the unmarshalling code and the corresponding tests.

I implemented `encoding.UnmarshalText` for the type instead of `json.UnmarshalJSON` since we are dealing with a single `string` value. The [docs say](https://golang.org/pkg/encoding/json/?source=post_page---------------------------#Unmarshal) that `json.Unmarshal` uses `encoding.UnmarshalText` if `json.UnmarshalJSON` is not available.

Aside from the tests, I also did some manual tests applying several modified yaml files and everything worked as expected. BTW, is there any end to end test that involves applying malformed deployment yaml files?

I didn't include code for marshalling. Since the underlying type is `string`, marshalling comes for free. I wrote tests to check this and to add protection for custom marshalling code that could be added in the future.

I've also removed the `WIP` label from the PR as it should be ready now :)

pkg/apis/jaegertracing/v1/deployment_strategy.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// DeploymentStrategyProduction represents the 'production' deployment strategy
+	// +k8s:openapi-gen=true
+	DeploymentStrategyProduction DeploymentStrategy = "production"
+)
+
+// UnmarshalText implements encoding.TextUnmarshaler to ensure that JSON values in the
+// strategy field of JSON jaeger specs are interpreted in a case-insensitive manner
+func (ds *DeploymentStrategy) UnmarshalText(text []byte) error {
+	if ds == nil {
```
jpkrohling: This is interesting. I'm probably missing something obvious, but how would this be triggered? Wouldn't a nil object fail with a nil reference before reaching this code? Can you come up with a test case for this?

volmedo: It's not obvious. It is possible for the function to be called on a nil receiver when interfaces are involved (as is the case here).

I was digging into the docs and I had a look at how [json.RawMessage.UnmarshalJSON](https://golang.org/src/encoding/json/stream.go?s=6952:7005#L257) is implemented. The check for nil felt strange, so I did a bit of research and came up with this https://tour.golang.org/methods/12

Wonderful Go! :)

volmedo: BTW, I guess this kind of check should be implemented in other types implementing interfaces for additional safety. `Options` come to my mind, implementing both `json.Marshaler` and `json.Unmarshaler` interfaces, but I'm sure there are several others.

jpkrohling: TIL! I'm merging this one. Would you mind creating an issue to do the same nil check in the other relevant types?

jpkrohling:
Thanks for your contribution!

pkg/apis/jaegertracing/v1/deployment_strategy.go
volmedo: Will do!

PR #718, opened 2019-10-21
Added workflows publishing the 'master' container image

### LINKED ISSUE #715: Publish latest/master images for every master build
Created from https://github.com/jaegertracing/jaeger-operator/pull/705#issuecomment-543809611

Each merge to master should publish a new image - either latest/master. The use-case is to allow users to try latest operator features. We do the same in Jaeger - every build publishes a new latest image.

One thing to consider, the latest image could also by default use the latest Jaeger components.

jpkrohling:
The script that is used for this workflow is the same we had before with Travis. It works locally when setting the username/password vars and I did add the same secrets to this repository. I _believe_ this will work, but we'll only know when this gets merged.

Resolves #715

### DIFF
--- a/.ci/publish-images.sh
+++ b/.ci/publish-images.sh
@@ -5,7 +5,7 @@ OPERATOR_VERSION=${OPERATOR_VERSION:-$(git describe --tags)}
 
 ## if we are on a release tag, let's extract the version number
 ## the other possible value, currently, is 'master' (or another branch name)
-## if we are not running in travis, it fallsback to the `git describe` above
+## if we are not running in the CI, it fallsback to the `git describe` above
 if [[ $OPERATOR_VERSION == release* ]]; then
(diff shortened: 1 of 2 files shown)


jpkrohling:
This uses the branch's name instead of `latest`, in line with how we used to do before moving to GitHub Workflows.

https://github.com/jaegertracing/jaeger-operator/blob/7664ef3a22954b64229abc448c2bf607f29faf75/.travis.yml#L102-L107

We could change the new ` .github/workflows/publish-images.yaml` to *also* publish the master images under the `latest` tag, but I'd do it in a follow-up PR, I'm not 100% sure this PR here will work on the first attempt.

jpkrohling:
@pavolloffay would be the right reviewer, as he's the author of the linked issue, but not sure he is available to review as he is at EclipseCon this week. @kevinearls, would you like to review this one instead?

kevinearls:
@jpkrohling  I'll check this afternoon, although I don't really have much background in this.

PR #744, opened 2019-11-05
Use ingress instead of port mapping for sidecar test

kevinearls:

Most of our tests use port mapping to connect to Jaeger query api or to create traces.  Unfortunately this is sometimes a bit flaky, so as an experiment I am updating the sidecar test (which seems to have this issue most frequently) to use ingress instead.  If this resolves that problem, I will add it to additional tests.


kevinearls:
@objectiser @jpkrohling @pavolloffay Please review

PR #749, opened 2019-11-08
Allow specifying ingress host

### LINKED ISSUE #135: Allow specifying ingress host
There seems to be no way to specify hostname for query ingress while creating a jaeger resource. Allowing for that as well as TLS settings would make it easy to handle by popular ingress controllers such as nginx.

I would approach it by adding fields to `jaeger` CRD similar to:
```
kind: Jaeger
metadata:
  name: my-jaeger
spec:
  ingress:
    enabled: true
    hosts:
    - myjaeger.example.com
    tls:
    - hosts:
       - sslexample.foo.com
       secretName: secret-tls
...
```

I could implement it as such if everyone is ok with this approach.

pfyod:
This PR resolves #135. 

Not mentioned in #135 but covered by this PR: `ingress.secretName` field is deprecated and is superseded by `ingress.TLS` which provides more flexibility. 

Signed-off-by: Pavels Fjodorovs <me@pfyod.com>


pkg/apis/jaegertracing/v1/jaeger_types.go
```diff
@@ -208,12 +208,25 @@ type JaegerIngressSpec struct {
 	OpenShift JaegerIngressOpenShiftSpec `json:"openshift,omitempty"`
 
 	// +optional
-	SecretName string `json:"secretName,omitempty"`
```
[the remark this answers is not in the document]
pfyod: Not entirely clear to me how to handle situations when both, the top level `secretName` and `secretName(s)` inside the `tls` block are specified. Should the top level `secretName` just be ignored or also used in cases when `tls[*].hosts` is specified but `tls[*].secretKey` is not?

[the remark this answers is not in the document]
objectiser: If the `tls` block has been specified, then they should not be using the deprecated `secretName` field - so I think it is safe to ignore but just log an extra warning - i.e. one warning that the deprecated field has been used (and what its replace is) and the other warning to indicate that the deprecated field will be ignored in favour of the new value?

pkg/ingress/query.go
```diff
@@ -82,22 +77,47 @@ func (i *QueryIngress) Get() *extv1beta1.Ingress {
 	}
 }
 
+func (i *QueryIngress) addRulesSpec(spec *extv1beta1.IngressSpec, backend *extv1beta1.IngressBackend) {
+	if path, ok := i.jaeger.Spec.AllInOne.Options.Map()["query.base-path"]; ok && i.jaeger.Spec.Strategy == v1.DeploymentStrategyAllInOne {
+		spec.Rules = append(spec.Rules, getRules(path, i.jaeger.Spec.Ingress.Hosts, backend)...)
+	} else if path, ok := i.jaeger.Spec.Query.Options.Map()["query.base-path"]; ok && i.jaeger.Spec.Strategy == v1.DeploymentStrategyProduction {
+		spec.Rules = append(spec.Rules, getRules(path, i.jaeger.Spec.Ingress.Hosts, backend)...)
+	} else if i.jaeger.Spec.Ingress.Hosts != nil {
```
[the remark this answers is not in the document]
pfyod: IMO, it makes sense to follow the k8s Ingress spec here. Whatever is possible with k8s Ingress should be possible to specify with the Jaeger resource's ingress section.  

Essentially, there are two variables at play here:

1. Are there ingress hosts specified (A)?
2. Is query path set via `query.base-path` (B)?

A and B (`TestQueryIngressQueryBasePathWithHosts`) →
```
spec:
  rules:
    - host: <ingress.hosts[*]> #repeated for each host
      http:
        paths:
          - path: <query.base-path>
            backend:
              serviceName: <generated query service name>
```
A and not B (`TestQueryIngressWithHosts`) →
```
spec:
  rules:
    - host: <ingress.hosts[*]> #repeated for each host
      http:
        paths:
          - backend:
              serviceName: <generated query service name>
```
not A and B (`TestQueryIngressQueryBasePath`, already works like this without this PR) →
```
spec:
  rules:
    - http:
        paths:
          - path: <query.base-path>
            backend:
              serviceName: <generated query service name>
```
not A and not B (`TestQueryIngressWithoutHosts`, already works like this without this PR) →
```
spec:
  backend:
    serviceName: <generated query service name>
```

I will refactor the code to make all these cases a bit more evident. Something like this:
```
path = get query.base-path from spec or "" if not exists
if jaegerSpec.hosts is not empty or path is not ""
  spec.Rules = getRules(path, jaegerSpec.hosts, backend)
else
  spec.Backend = backend // this is case of "not A and not B"
```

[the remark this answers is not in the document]
objectiser: +1, that would be clearer.

pkg/ingress/query_test.go
```diff
@@ -120,12 +120,142 @@ func TestQueryIngressLabels(t *testing.T) {
...
+	jaeger.Spec.Ingress.Hosts = []string{"test-host-1"}
+
+	ingress := NewQueryIngress(jaeger)
+
+	dep := ingress.Get()
+
+	assert.NotNil(t, dep)
+	assert.Nil(t, dep.Spec.Backend)
+	assert.Len(t, dep.Spec.Rules, 1)
+	assert.Empty(t, dep.Spec.Rules[0].HTTP.Paths[0].Path)
```
objectiser: Shouldn't `Backend` also be tested here and in the next test?

pfyod: @objectiser you are right, added the missing assertions

deploy/examples/with-ingress-tls.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: jaegertracing.io/v1
```
jpkrohling: I think only one example file with both properties is enough.

pkg/apis/jaegertracing/v1/jaeger_types.go
```diff
@@ -207,13 +207,30 @@ type JaegerIngressSpec struct {
 	// +optional
 	OpenShift JaegerIngressOpenShiftSpec `json:"openshift,omitempty"`
 
+	// +optional
+	Hosts []string `json:"hosts,omitempty"`
+
+	// +optional
+	TLS []JaegerIngressTLSSpec `json:"tls,omitempty"`
```
jpkrohling: Wouldn't it be better to just use `Ingress` instead?

deploy/examples/with-ingress-tls.yaml
pfyod: What would you prefer:

1. Keep the `with-ingress-host-and-tls` example, remove `with-ingress-host` and `with-ingress-tls`
2. Add ingress options (basically `with-ingress-host-and-tls`): 
    ```
      ingress:
        hosts:
          - jaeger.example.com
        tls:
          - hosts:
              - jaeger.example.com
            secretName: jaeger-example-com-tls
    ```
    to `all-in-one-with-options` and remove all `with-ingress-*` examples (`with-ingress-host-and-tls`, `with-ingress-host`, `with-ingress-tls`

pkg/apis/jaegertracing/v1/jaeger_types.go
pfyod: Didn't quite get what you mean. Could you elaborate?

deploy/examples/with-ingress-tls.yaml
jpkrohling: Keeping `with-ingress-host-and-tls` would be fine

pkg/apis/jaegertracing/v1/jaeger_types.go
jpkrohling: Sorry, I meant that we could consider adding a new field holding `extv1beta1.Ingress` instead. We would still decorate the ingress with our own annotations, and we would probably adjust the backend values, but the ingress itself would be whatever the user is defining.

Note that I'm still not sure this is a good idea :-)

pfyod: I don't think that it's a good idea. That would:
1. Allow the user to specify values that will **always** be overridden / ignored (`rules[*].http.paths[*].path`, `rules[*].http.paths[*].backend.serviceName` and `backend.serviceName`);
2. Introduce additional error-prone matching / validation logic;
3. Potentially allow the user to expose non-jaeger services via jaeger operator's managed ingress.

Consider the following example:
```
metadata:
  name: "my-jaeger"
spec:
  strategy: allInOne
  allInOne:
    options:
      query:
        base-path: /jaeger # base-path is set here
  ingress: 
    rules:
      - host: jaeger.example.com
        http:
          paths:
            # should this be overridden based on the value above? just validated?
            - path: /query 
              backend:
                # do we validate this or overwrite with the generated service name?
                serviceName: my-jaeger-query
            # what about this?
            - path: /also-query 
              backend:
                serviceName: my-jaeger-query
      - host: also-jaeger.example.com
        http:
          paths:
            - backend:
                # should this even be allowed?
                serviceName: some-other-non-jaeger-service
```

jpkrohling: You are right, let's keep it simple for now.

deploy/examples/with-ingress-tls.yaml
pfyod: Done

jpkrohling:

mumrau:
I am late to the party, this is an awesome feature (and I was waiting for 1.16 to have this included), but I think we could use annotations as well? As this is a very common way to manage networking in k8s cluster (like whitelisting, redundancy or other stuff).
I hope I haven't missed out on the documentation for such a feature.

What do you think? I know this is harder than it looks like, because if we put in on the ingress, why not on the other resources as well? I'd say it's because it is a truly external component, and more importantly, externally handled, but...

pfyod:
@mumrau you can already specify ingress annotations. Here's the test: https://github.com/jaegertracing/jaeger-operator/blob/fcf9d40129c3fe9e3e6a0118118072356d21555b/pkg/ingress/query_test.go#L89

PR #754, opened 2019-11-09
Docker image version fix (1.15.0 -> v1.15.0)

exelban:
Hi.

I noticed that the operator configuration file has a wrong docker image version.
[Docker Hub](https://hub.docker.com/r/jaegertracing/jaeger-operator/tags) - according to this there is no version of jaeger-operator `1.15.0` but `v1.15.0`.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: jaeger-operator
       containers:
         - name: jaeger-operator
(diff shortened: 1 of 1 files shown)


jpkrohling:
Apparently, our release script is publishing tagged images with `v` in front. For now, I just tagged the same image with `1.15.0`, to give us time to discuss which format we want from now on.

exelban:
@jpkrohling thanks for the reply.
I think this PR can be closed.

Issue #771, opened 2019-11-15
Sidecar exposes all ports as TCP, but some are expected to be UDP


garyfritz:
Analogous to #577, when [injecting agent containers](https://www.jaegertracing.io/docs/1.14/operator/#auto-injecting-jaeger-agent-sidecars), the container ports of injected sidecar containers are exposed as TCP, but three out of four should be UDP.

This is from an agent pod from a daemon set (which is correct according to #577):
```
  jaeger-agent-daemonset:
    Image:         jaegertracing/jaeger-agent:1.15.0
    Ports:         5775/UDP, 5778/TCP, 6831/UDP, 6832/UDP, 14271/TCP
    Host Ports:    5775/UDP, 5778/TCP, 6831/UDP, 6832/UDP, 14271/TCP
    Args:
      --reporter.grpc.host-port=dns:///izjaeger-collector-headless.default:14250
      --reporter.type=grpc
```
While this is from an injected agent container in some of my pods:
```
  jaeger-agent:
    Image:         jaegertracing/jaeger-agent:1.15.0
    Ports:         5775/TCP, 5778/TCP, 6831/TCP, 6832/TCP
    Host Ports:    0/TCP, 0/TCP, 0/TCP, 0/TCP
    Args:
      --reporter.grpc.host-port=dns:///izjaeger-collector-headless.default:14250
      --reporter.type=grpc
```
I'd suggest adding the missing protocol specifications to https://github.com/jaegertracing/jaeger-operator/blob/058f4fa3f99df8dad922c27c286c1eac3799da2f/pkg/inject/sidecar.go#L158-L175

garyfritz:
Closing since MR #773 is now merged.

PR #773, opened 2019-11-15
Correctly expose injected UDP container ports

garyfritz:
This MR should fix the missing `Protocol` properties for three of the created container ports of injected sidecar containers (#771). A test case should cover this - at least for default port numbers. Also, I updated the test case for standalone agent containers to cover the protocols.

### DIFF
--- a/pkg/service/agent_test.go
+++ b/pkg/service/agent_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
@@ -26,10 +27,17 @@ func TestAgentServiceNameAndPorts(t *testing.T) {
 
 	for _, port := range svc.Spec.Ports {
 		ports[port.Port] = true
+		switch port.Port {
+		case
+			5775,
+			6831,
+			6832:
+			assert.Equal(t, corev1.ProtocolUDP, port.Protocol, "Expected port %v to be UDP, but wasn't", port)
+		}
 	}
 
 	for k, v := range ports {
-		assert.Equal(t, v, true, "Expected port %v to be specified, but wasn't", k)
+		assert.Equal(t, true, v, "Expected port %v to be specified, but wasn't", k)
 	}
 
 }
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -159,6 +159,7 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 			{
 				ContainerPort: zkCompactTrft,
 				Name:          "zk-compact-trft",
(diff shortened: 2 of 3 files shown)


jpkrohling:
There seems to be a formatting issue with the `pkg/inject/sidecar_test.go`:

```
Go fmt, license check, or import ordering failures, run 'make format'
pkg/inject/sidecar_test.go
```

garyfritz:
@jpkrohling My bad, first time hands-on w/ Go. Formatting issue solved 👍(after downgrading Go).

garyfritz:
Changed rebased+squashed.

pkg/service/agent_test.go
```diff
@@ -26,6 +27,13 @@ func TestAgentServiceNameAndPorts(t *testing.T) {
 
 	for _, port := range svc.Spec.Ports {
 		ports[port.Port] = true
+		switch port.Port {
+		case
+			5775,
+			6831,
+			6832:
+			assert.Equal(t, port.Protocol, corev1.ProtocolUDP, "Expected port %v to be UDP, but wasn't", port)
```
jpkrohling: The second argument to `Equal` is the expected value, followed by the actual value:

```
assert.Equal(t, corev1.ProtocolUDP, port.Protocol, ...)
```

Of course, this is not blocking the PR :)

garyfritz: Well, right! But then, for the same reason, the next assert arguments should also be swapped as well 😉 

After rebasing and seeing the test cases fail, I ran into the same failures locally. 
And they do, indeed, complain ports or port forwarding. So I'm not sure did maybe sth change from 1.14 to 1.15? In that case, it is good that the test cases fail.
```
E1118 15:55:18.902332   45722 portforward.go:400] an error occurred forwarding 57500 -> 14268: error forwarding port 14268 to pod 65359bd8523f97b65fb2350d0639d7bc5de8e68e7c752614c2c63e61d76755d0, uid : exit status 1: 2019/11/18 14:55:18 socat[24686] E connect(5, AF=2 127.0.0.1:14268, 16): Connection refused
E1118 15:57:18.950977   45722 portforward.go:408] error closing listener: close tcp4 127.0.0.1:57499: use of closed network connection
E1118 15:57:18.951003   45722 portforward.go:408] error closing listener: close tcp6 [::1]:57499: use of closed network connection
E1118 15:57:18.952080   45722 portforward.go:408] error closing listener: close tcp4 127.0.0.1:57500: use of closed network connection
E1118 15:57:18.953160   45722 portforward.go:408] error closing listener: close tcp6 [::1]:57500: use of closed network connection
```
To my understanding, these errors occur in https://github.com/jaegertracing/jaeger-operator/blob/3020ae448e003c72b7cc6d16144e68dae9305415/test/e2e/smoketest.go#L69 but are actually originating from the method https://github.com/jaegertracing/jaeger-client-go/blob/ab94682f276c8c307e7fcd3f551c48ad459eab8f/tracer.go#L366-L377 not checking are the resources closable at all. So the actual error seems to be an upstream issue.

garyfritz:
Well, errors are gone. Maybe it was a timeout issue. Looks good now 😉

PR #777, opened 2019-11-20
Cleaned up tools.go

### LINKED ISSUE #711: Build with Golang 1.13.x
Build the repository with Golang 1.13.x. At the moment it produces checksum errors https://github.com/jaegertracing/jaeger-operator/issues/709

jpkrohling:
Fixes #711 

Master is currently failing for me (Fedora 31, go1.13.4) because of changes to `latest` of `sigs.k8s.io/controller-tools/pkg/crd/generator`. Turns out, we don't actually need this at runtime, only at build time. In fact, there's nothing in the `tools.go` that we need at runtime.


go.mod
```diff
@@ -3,15 +3,15 @@ module github.com/jaegertracing/jaeger-operator
 go 1.12
```
pavolloffay: This should have been also updated

pavolloffay:
Proper way to fix this is to move the tools.go to an internal package with a separate go module.

PR #792, opened 2019-11-28
Reuse inject function from deployment controller into autodetect

### LINKED ISSUE #774: Reuse inject update loop from deployment controller into autodetect.
This Issue was created to address remainding comments of PR: https://github.com/jaegertracing/jaeger-operator/pull/454

rubenvp8510:

Fixes #774

### DIFF
--- a/pkg/autodetect/main_test.go
+++ b/pkg/autodetect/main_test.go
@@ -23,6 +23,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
+	"github.com/jaegertracing/jaeger-operator/pkg/service"
+
 	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
 	"github.com/jaegertracing/jaeger-operator/pkg/inject"
 )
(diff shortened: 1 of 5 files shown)


jpkrohling:
Looks like the unit tests are failing.

pkg/inject/sidecar.go
```diff
@@ -251,3 +254,34 @@ func CleanSidecar(deployment *appsv1.Deployment) {
 		}
 	}
 }
+
+// IfNeeded inject the sidecar on deployment only if it is required
+func IfNeeded(client client.Client, instance *appsv1.Deployment) (bool, error) {
```
jpkrohling: Can't this be tested? It would be really good to have some tests here.

rubenvp8510: Yes, it can :)

rubenvp8510: Done

jpkrohling:
This needs a rebase and will conflict due to #814

rubenvp8510:
> This needs a rebase and will conflict due to #814

:cry:

pkg/autodetect/main_test.go
```diff
@@ -496,34 +498,96 @@ func TestCleanDeployments(t *testing.T) {
...
+					Containers: []corev1.Container{
+						{
+							Name: "my-business-container",
+						},
+					},
+				},
+			},
+		},
+	}
+	dep.Name = "my-business-container"
```
jpkrohling: Why not set on the statement above?

pkg/autodetect/main_test.go
```diff
@@ -496,34 +498,96 @@ func TestCleanDeployments(t *testing.T) {
...
+	require.NoError(t, err)
+	err = cl.Create(context.TODO(), jaeger2)
+	require.NoError(t, err)
+
+	dep.Annotations = map[string]string{inject.Annotation: jaeger1.Name}
+	err = cl.Create(context.TODO(), dep)
+	require.NoError(t, err)
+
+	err = cl.Create(context.TODO(), dep2)
+
```
jpkrohling: Formatting seems a bit odd... Can you get the require.NoError close to the `err` it's testing?

pkg/autodetect/main_test.go
```diff
@@ -496,34 +498,96 @@ func TestCleanDeployments(t *testing.T) {
...
+		},
 	}
+	dep2.Name = "my-business-container-2"
+
+	err := cl.Create(context.TODO(), jaeger1)
+	require.NoError(t, err)
+	err = cl.Create(context.TODO(), jaeger2)
+	require.NoError(t, err)
+
+	dep.Annotations = map[string]string{inject.Annotation: jaeger1.Name}
```
jpkrohling: This should be part of the definition of `dep`.

pkg/autodetect/main_test.go
```diff
@@ -496,34 +498,96 @@ func TestCleanDeployments(t *testing.T) {
...
+	err = cl.Get(context.Background(), types.NamespacedName{
+		Namespace: dep.Namespace,
+		Name:      dep.Name,
+	}, persisted)
+	assert.Equal(t, 2, len(persisted.Spec.Template.Spec.Containers))
+	err = cl.Get(context.Background(), types.NamespacedName{
+		Namespace: dep2.Namespace,
+		Name:      dep2.Name,
+	}, persisted)
+	assert.NoError(t, err)
```
jpkrohling: From what I could understand, `dep2` will get `jaeger1` injected at this point. Please, add an assertion to make sure this is what's happening.

pkg/autodetect/main_test.go
```diff
@@ -496,34 +498,96 @@ func TestCleanDeployments(t *testing.T) {
...
+	err = cl.Update(context.Background(), dep)
+	require.NoError(t, err)
+
+	err = b.detectDeploymentUpdates()
+	err = cl.Get(context.Background(), types.NamespacedName{
+		Namespace: dep.Namespace,
+		Name:      dep.Name,
+	}, persisted)
+
+	reporterEndpoint := fmt.Sprintf("--reporter.grpc.host-port=dns:///%s.%s:14250", service.GetNameForHeadlessCollectorService(jaeger2), jaeger2.Namespace)
```
jpkrohling: Keep this assertion, but the official way to inform what's the Jaeger instance being used is via the annotation. So, add an assertion for the annotation that was persisted.

pkg/controller/deployment/deployment_controller.go
```diff
@@ -3,6 +3,10 @@ package deployment
 import (
 	"context"
 
+	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
```
jpkrohling: Import formatting seems odd.

pkg/controller/deployment/deployment_controller.go
```diff
@@ -80,33 +81,18 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
+	pods := &v1.JaegerList{}
+	err = r.client.List(context.Background(), pods)
```
jpkrohling: Limit it to the namespaces being watched.

pkg/controller/deployment/deployment_controller.go
```diff
@@ -80,33 +81,18 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
...
-			}).Info("Injecting Jaeger Agent sidecar")
-			instance = inject.Sidecar(jaeger, instance)
-			if err := r.client.Update(context.Background(), instance); err != nil {
-				log.WithField("deployment", instance).WithError(err).Error("failed to update")
-				return reconcile.Result{}, err
-			}
-		} else {
-			log.WithField("deployment", instance.Name).Info("No suitable Jaeger instances found to inject a sidecar")
+	if injected != nil {
+		if err := r.client.Update(context.Background(), injected); err != nil {
```
jpkrohling: It would probably be better to create a context at the beginning of the reconcile method, and use the same context everywhere down the stream. It will make it easier once we decide to add tracing to this controller, like we added to the other controller.

pkg/inject/sidecar.go
```diff
@@ -251,3 +251,22 @@ func CleanSidecar(deployment *appsv1.Deployment) {
 		}
 	}
 }
+
+// IfNeeded inject the sidecar on deployment only if it is required
+func IfNeeded(instance *appsv1.Deployment, pods *v1.JaegerList) *appsv1.Deployment {
```
jpkrohling: I don't think this function name is appropriate. I had to check it twice while doing this review, as I would expect it to return an `instance` (not nil): either the original instance, or a changed instance, but not nil. The func documentation also doesn't tell that this might return nil, and what it means.

I think it would be better to change the `IfNeeded` consumers to check `Needed` instead.

rubenvp8510:
> Did you just do a rebase, or have you changed the code significantly? Looks like new code to me :-)

I didn't change anything :) may be this is an old PR and need some updates

PR #816, opened 2019-12-05
Add wait for jaeger deployment to fix race conditions

kevinearls:

This should fix a couple of different test failures that occur intermittently with this test.  Previously the vertx app could finish deploying before the Jaeger instance did, and no traces would actually get logged.


kevinearls:
@jpkrohling @objectiser @pavolloffay Please review

PR #817, opened 2019-12-05
Retry http request after timeout to avoid intermittent test failure

kevinearls:

This test fails intermittently because of request timeouts, so here I have increased the timeout, and also changed the wait loop to retry instead of failing if we still get timeouts.

I've also made a few changes where I call require.NoError rather than returning the error, as this gives better information in the case of failures.


kevinearls:
@objectiser @jpkrohling  @pavolloffay Please review

Issue #823, opened 2019-12-10
Error cleaning orphaned deployment


jpkrohling:
While running the e2e tests in OpenShift, I noticed that some errors related to the cleaning of orphaned deployments (ie, deployments with an Agent pointing to a non-existing Jaeger Collector). I was able to reproduce it on an OpenShift 4.2 cluster using the manifests generated by a previous e2e test:

```
$ kubectl apply -f deploy/test/namespace-manifests.yaml
rolebinding.rbac.authorization.k8s.io/jaeger-operator created
role.rbac.authorization.k8s.io/jaeger-operator created
serviceaccount/jaeger-operator created
deployment.apps/jaeger-operator created
```

Once the operator's pod is up, the following can be seen:
```
time="2019-12-10T09:31:12Z" level=error msg="error cleaning orphaned deployment" error="deployments.apps is forbidden: User \"system:serviceaccount:orphaneddeployment:jaeger-operator\" cannot list resource \"deployments\" in API group \"apps\" at the cluster scope"
time="2019-12-10T09:31:12Z" level=error msg="error cleaning orphaned deployment" error="jaegers.jaegertracing.io is forbidden: User \"system:serviceaccount:orphaneddeployment:jaeger-operator\" cannot list resource \"jaegers\" in API group \"jaegertracing.io\" at the cluster scope"
```

The full logs are here: https://paste.centos.org/view/70ba245f

This might be an issue only in the `role.yaml` used by the tests, meaning that it might not be affecting real operator deployments.

cc @kevinearls, @rubenvp8510

jpkrohling:
I think we fixed this one already, right @rubenvp8510  ?

rubenvp8510:
Yes

Issue #791, opened 2019-11-27
is clusterrole required for Jaeger operator?


christine-gong:
Hi,
I am wondering if clusterrole and clusterrolebinding are required for jaeger operator  to function properly in kubernetes. (e.g. https://github.com/jaegertracing/jaeger-operator/blob/master/deploy/role.yaml and https://github.com/jaegertracing/jaeger-operator/blob/master/deploy/role_binding.yaml) 

Since everything is living under `tracing` namespace, i think role and rolebinding would be good enough.  We do not want to open cluster permission when it's unnecessary

After changing to Role and RoleBinding, my operator throws the following errors when applying jaeger components yamls.
```
1 reflector.go:126] pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/reflector.go:94: Failed to list *v1.ClusterRoleBinding: clusterrolebindings.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:tracing:jaeger-operator" cannot list resource "clusterrolebindings" in API group "rbac.authorization.k8s.io" at the cluster scope
```

jpkrohling:
Cluster permissions (role/role binding) should be required only when watching all namespaces. In all other cases, the operator has to be able to work without cluster permissions.

It should be possible to fix the code so that the operator won't break if it can't list cluster roles/role bindings. A quick workaround should be to not return an error on this code:

https://github.com/jaegertracing/jaeger-operator/blob/8eb82f0caad364914ea6f353fa46b180d5ee2b21/pkg/controller/jaeger/jaeger_controller.go#L266-L268

Ideally, though, the operator would attempt to get a list of cluster role bindings and set a new configuration entry via viper, stating whether or not it has the permission to get a list of cluster role bindings. Then, the following code should check this new config entry before attempting to list the cluster role bindings:

https://github.com/jaegertracing/jaeger-operator/blob/8eb82f0caad364914ea6f353fa46b180d5ee2b21/pkg/controller/jaeger/clusterrolebinding.go#L16-L62

Would you be willing to give it a try?

christine-gong:
@jpkrohling thanks! sure, i will try to fix it.

PR #825, opened 2019-12-11
[WIP] Detect when operator scope is not the whole cluster

rubenvp8510:
Fixes https://github.com/jaegertracing/jaeger-operator/issues/823

@jpkrohling  WDYT?

With this maybe we can fix https://github.com/jaegertracing/jaeger-operator/issues/791 too.

### DIFF
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	appsv1 "k8s.io/api/apps/v1"
@@ -29,6 +30,7 @@ type Background struct {
 
 	retryDetectKafka bool
 	retryDetectEs    bool
+	watchNamespace   string
 }
 
 // New creates a new auto-detect runner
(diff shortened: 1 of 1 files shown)


pkg/autodetect/main.go
```diff
@@ -149,6 +151,14 @@ func (b *Background) autoDetectCapabilities() {
 		b.detectKafka(apiList)
 	}
 
+	watchNamespace, err := k8sutil.GetWatchNamespace()
```
jpkrohling: Please, create a new func to hold this logic. Also, please check what's the error you get back. Just receiving an error might not be sufficient to make the right decision: on the first run of this func, this might correctly detect that the scope is `cluster`, but a future unrelated failure (network failure when talking to some backend API server) will change the scope to `namespace`.

christine-gong:
btw i applied your change, https://github.com/jaegertracing/jaeger-operator/issues/791 error is still there

rubenvp8510:
@christine-gong  I haven't validate that case, but  I mean I can do it using the same method I used here.

jpkrohling:
This PR won't fix that, but might serve as base. Once this logic is applied, we can store the outcome in viper. From there, it should be ready to accomplish what I mentioned as a comment there: https://github.com/jaegertracing/jaeger-operator/issues/791#issuecomment-559433530

PR #826, opened 2019-12-12
Removed OpenShift e2e tests

jpkrohling:
This commit removes the OpenShift e2e tests, as they are running on unsupported versions of OpenShift and are failing when testing new features that won't fail on newer versions.

Related to #819.


jpkrohling:
cc @objectiser, @kevinearls, @pavolloffay

jpkrohling:
Not that I could find, but we could probably come up with a hack with the trigger, so that it won't ever be triggered. Personally, I would prefer to check the git history in order to enable it again.

jpkrohling:
I know that @pavolloffay has concerns about disabling this, and @kevinearls reassured that there are tests being performed elsewhere that would help us get notified if the ES auto-provisioning feature breaks. Under this premise, I'm merging this, so that we can keep moving in adding features that are only available in OpenShift 4.x.

The tests will be enabled back once a proper CI for OpenShift 4.x is in place.

PR #830, opened 2019-12-17
[Issue 791] fix clusterrole hard requirement

christine-gong:
Fix for https://github.com/jaegertracing/jaeger-operator/issues/791
Followed the logic of https://github.com/jaegertracing/jaeger-operator/pull/825 to figure out watched namespace when having cluster scope error.

This is my first PR for this repo, let me know if this is sane.

### DIFF
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -230,11 +230,7 @@ func (b *Background) hasClusterRolePermission() bool {
 	err := b.clReader.List(context.Background(), list, opts...)
 	if err != nil {
 		log.WithField("errormsg", err).Debug("Failed to list cluster role bindings")
-		if strings.HasSuffix(err.Error(), "cluster scope") {
-			return false
-		}
-		// still return true if not scope issue
-		return true
+		return false
 	}
 	return true
 }
@@ -243,14 +239,6 @@ func (b *Background) detectClusterPermissions() {
 	if b.hasClusterRolePermission() {
 		viper.Set("has-cluster-permission", true)
 	} else {
-		watchNamespace, err := k8sutil.GetWatchNamespace()
-		if err != nil {
-			// fail silently
-			viper.Set("watch-ns", "")
-		} else {
-			log.Debugf("Namespace scope on namespace %s", watchNamespace)
-			viper.Set("watch-ns", watchNamespace)
-		}
 		viper.Set("has-cluster-permission", false)
 	}
 }
@@ -289,10 +277,15 @@ func (b *Background) cleanDeployments() {
 
 	jaegerOpts := []client.ListOption{}
 
-	viperWatchNs := viper.GetString("watch-ns")
-	if viperWatchNs != "" {
-		jaegerOpts = append(jaegerOpts, client.InNamespace(viperWatchNs))
-		deployOpts = append(deployOpts, client.InNamespace(viperWatchNs))
+	watchNamespace, err := k8sutil.GetWatchNamespace()
+	if err != nil {
+		// fail silently
+		watchNamespace = ""
+	}
+
+	if watchNamespace != "" {
+		jaegerOpts = append(jaegerOpts, client.InNamespace(watchNamespace))
+		deployOpts = append(deployOpts, client.InNamespace(watchNamespace))
 	}
 
 	instances := &v1.JaegerList{}
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -9,6 +9,7 @@ import (
 	"github.com/spf13/viper"
 	appsv1 "k8s.io/api/apps/v1"
 	authenticationapi "k8s.io/api/authentication/v1"
+	rbac "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
@@ -149,6 +150,12 @@ func (b *Background) autoDetectCapabilities() {
 		b.detectKafka(apiList)
 	}
 
+	if b.hasClusterRolePermission() {
+		viper.Set("has-cluster-permission", true)
+	} else {
+		viper.Set("has-cluster-permission", false)
+	}
+
 	b.detectClusterRoles()
 	b.cleanDeployments()
 	b.detectDeploymentUpdates()
@@ -162,6 +169,17 @@ func (b *Background) availableAPIs() (*metav1.APIGroupList, error) {
 	}
 
 	return apiList, nil
+}
+
+func (b *Background) hasClusterRolePermission() bool {
+	list := &rbac.ClusterRoleBindingList{}
+	opts := []client.ListOption{}
+	err := b.clReader.List(context.Background(), list, opts...)
+	if err != nil {
+		log.WithField("errormsg", err).Debug("Failed to list cluster role bindings")
+		return false
+	}
+	return true
 }
 
 func (b *Background) detectPlatform(apiList *metav1.APIGroupList) {
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	appsv1 "k8s.io/api/apps/v1"
@@ -150,12 +151,7 @@ func (b *Background) autoDetectCapabilities() {
 		b.detectKafka(apiList)
 	}
 
-	if b.hasClusterRolePermission() {
-		viper.Set("has-cluster-permission", true)
-	} else {
-		viper.Set("has-cluster-permission", false)
-	}
-
+	b.detectClusterPermissions()
 	b.detectClusterRoles()
 	b.cleanDeployments()
 	b.detectDeploymentUpdates()
@@ -169,17 +165,6 @@ func (b *Background) availableAPIs() (*metav1.APIGroupList, error) {
 	}
 
 	return apiList, nil
-}
-
-func (b *Background) hasClusterRolePermission() bool {
-	list := &rbac.ClusterRoleBindingList{}
-	opts := []client.ListOption{}
-	err := b.clReader.List(context.Background(), list, opts...)
-	if err != nil {
-		log.WithField("errormsg", err).Debug("Failed to list cluster role bindings")
-		return false
-	}
-	return true
 }
 
 func (b *Background) detectPlatform(apiList *metav1.APIGroupList) {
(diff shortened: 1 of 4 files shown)


pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -291,8 +291,11 @@ func (r *ReconcileJaeger) apply(ctx context.Context, jaeger v1.Jaeger, str strat
 		return jaeger, tracing.HandleError(err, span)
 	}
 
-	if err := r.applyClusterRoleBindingBindings(ctx, jaeger, str.ClusterRoleBindings()); err != nil {
-		return jaeger, tracing.HandleError(err, span)
+	// only apply clusterolebindings when the list is not empty
+	if len(str.ClusterRoleBindings()) > 0 {
```
christine-gong: actually this is not correct. i'll change this

christine-gong:
@jpkrohling I think the permission error is triggered by https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/controller/jaeger/clusterrolebinding.go#L26 but it does not set err properly, e.g. the error message does not follow the jaeger operator format. What's the suggested way here to precheck its permission to list clusterrolebindings?

jpkrohling:
> What's the suggested way here to precheck its permission to list clusterrolebindings?

The best would be to have a new func in the auto detect package, setting a viper config entry to the outcome of the permission. Like:

```golang

func hasClusterPermissions() bool {
  _, err := client.List....
  if err != nil {
    viper.Set("has-cluster-permissions", true) // see the other entries for naming and usage patterns
  } else {
    viper.Set("has-cluster-permissions", false)
  }
}
```

Then, in jaeger_controller.go, check the value of this config entry and skip `applyClusterRoleBindingBindings`. You can also set a debug message in case the strategy returns cluster role bindings to be created and the operator doesn't have enough permissions.

christine-gong:
@jpkrohling could you help review this? thanks!

pkg/autodetect/main.go
```diff
@@ -221,6 +224,37 @@ func (b *Background) detectKafka(apiList *metav1.APIGroupList) {
...
+		return true
+	}
+	return true
+}
+
+func (b *Background) detectClusterPermissions() {
+	if b.hasClusterRolePermission() {
+		viper.Set("has-cluster-permission", true)
+	} else {
+		watchNamespace, err := k8sutil.GetWatchNamespace()
```
jpkrohling: I believe this might return a comma-separated list of namespaces that the operator is watching. Not sure it has any impact in this PR, but the naming could be in the plural here if that's the case.

jpkrohling: In any case, I think you can just set the `has-cluster-permission` and let `k8sutil.GetWatchNamespace()` to be used by the other parts.

pkg/autodetect/main.go
```diff
@@ -254,6 +288,13 @@ func (b *Background) cleanDeployments() {
 	}
 
 	jaegerOpts := []client.ListOption{}
+
+	viperWatchNs := viper.GetString("watch-ns")
```
jpkrohling: I think there's no harm in using `k8sutil.GetWatchNamespace()` here directly.

pkg/autodetect/main.go
```diff
@@ -254,6 +288,13 @@ func (b *Background) cleanDeployments() {
...
+	if viperWatchNs != "" {
+		jaegerOpts = append(jaegerOpts, client.InNamespace(viperWatchNs))
```
jpkrohling: This might be tricky now :-) If my suspicion that the "WATCH_NAMESPACE" can return a list of namespaces, then this function will need to be called once per namespace. Perhaps change this function to accept a namespace as parameter, and create a `cleanDeploymentsWatchedNamespaces()` that iterates over the list and calls `cleanDeployments(namespace string)`?

pkg/upgrade/upgrade.go
```diff
@@ -21,11 +23,28 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
...
 	identity := viper.GetString(v1.ConfigIdentity)
-	opts := client.MatchingLabels(map[string]string{
+	opts := []client.ListOption{}
+	opts = append(opts, client.MatchingLabels(map[string]string{
 		v1.LabelOperatedBy: identity,
-	})
-	if err := reader.List(ctx, list, opts); err != nil {
-		return tracing.HandleError(err, span)
+	}))
+	if err := reader.List(ctx, list, opts...); err != nil {
```
jpkrohling: You can have a conditional here like you have in the `jaeger_controller`, checking for cluster-permissions before failing due to the lack of permissions.

pkg/autodetect/main.go
```diff
@@ -221,6 +224,37 @@ func (b *Background) detectKafka(apiList *metav1.APIGroupList) {
 	}
 }
 
+func (b *Background) hasClusterRolePermission() bool {
+	list := &rbac.ClusterRoleBindingList{}
+	opts := []client.ListOption{}
+	err := b.clReader.List(context.Background(), list, opts...)
+	if err != nil {
+		log.WithField("errormsg", err).Debug("Failed to list cluster role bindings")
+		if strings.HasSuffix(err.Error(), "cluster scope") {
```
jpkrohling: Not quite sure I agree here... I *think* I would return `false` no matter the error. What do you think, @rubenvp8510?

pkg/autodetect/main.go
christine-gong: my understanding is that WATCH_NAMESPACE (https://github.com/jaegertracing/jaeger-operator/blob/90b4053280fb230f99da4a6be5b40091d754ae43/deploy/operator.yaml#L27) is the namespace for jaeger, and it could only be one? Does that field accept a list of strings? If Jaeger and Jaeger operator sit in two different namespaces, jaeger operator would need cluster role.

pkg/upgrade/upgrade.go
christine-gong: Sure i can add that. The reason I did not add that before is that we have upgrade check before the autodect, that viper config was not set in the beginning. But no warm to add

pkg/autodetect/main.go
christine-gong: That also makes sense

pkg/autodetect/main.go
christine-gong: ok

christine-gong:
@jpkrohling @rubenvp8510 could u help review the updated changes when you get a chance? thanks

pkg/autodetect/main.go
jpkrohling: POD_NAMESPACE is the namespace where the Jaeger Operator is running. WATCH_NAMESPACE is which namespaces the Jaeger Operator is watching for new events (new/updated/deleted Jaeger CRs). I'm sure multiple values are possible, just not sure what's the format.

jpkrohling:
Sorry for taking so long to review again, but I was out for the past two weeks. Once the change to the WATCH_NAMESPACE is done, I'll review it again.

jpkrohling:
@christine-gong just wanted to check how the change to the WATCH_NAMESPACE is going. Do you need help there?

pkg/autodetect/main.go
rubenvp8510: I saw some operators where WATCH_NAMESPACE can be one namespace or a comma-separated list. I think is ok if we want to support multiple namespaces here.

pkg/autodetect/main.go
rubenvp8510: Yes, I think it makes more sense to return false here, TBH I don't like to have to check the error message string for validations, but there is no other way.

So we can assume that if there is an error here we don't have permissions or check the error, I would prefer to do the assumption and log the error just in case.

pkg/autodetect/main.go
rubenvp8510: If I'm not wrong, Is not necessary to have ClusterRole permissions to watch multiple namespaces, you only need the appropriate permissions on each namespace.  Of course watching more than one namespace complicates this PR a little bit. so I'm ok if for now we only support one namespace.

jpkrohling: I just tried and it doesn't seem to be possible to have more than one namespace there. ClusterRole *is* necessary when `WATCH_NAMESPACE` is empty, though.

pkg/autodetect/main.go
```diff
@@ -221,6 +224,25 @@ func (b *Background) detectKafka(apiList *metav1.APIGroupList) {
 	}
 }
 
+func (b *Background) hasClusterRolePermission() bool {
+	list := &rbac.ClusterRoleBindingList{}
+	opts := []client.ListOption{}
+	err := b.clReader.List(context.Background(), list, opts...)
+	if err != nil {
+		log.WithField("errormsg", err).Debug("Failed to list cluster role bindings")
```
jpkrohling: `WithErr` is more appropriate here than `WithField`

pkg/autodetect/main.go
```diff
@@ -254,6 +276,18 @@ func (b *Background) cleanDeployments() {
 	}
 
 	jaegerOpts := []client.ListOption{}
+
+	watchNamespace, err := k8sutil.GetWatchNamespace()
+	if err != nil {
+		// fail silently
+		watchNamespace = ""
```
jpkrohling: Perhaps log it? If it fails, something is really wrong...

pkg/upgrade/upgrade.go
```diff
@@ -21,11 +23,33 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
...
-	opts := client.MatchingLabels(map[string]string{
+	opts := []client.ListOption{}
+	opts = append(opts, client.MatchingLabels(map[string]string{
 		v1.LabelOperatedBy: identity,
-	})
-	if err := reader.List(ctx, list, opts); err != nil {
-		return tracing.HandleError(err, span)
+	}))
+
+	// if set and np cluster scope permission, skip
```
jpkrohling: `s/np/no`?

pkg/upgrade/upgrade.go
```diff
@@ -21,11 +23,33 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
...
+	// if not set, treat as true
+	if (viper.IsSet("has-cluster-permission") && viper.GetBool("has-cluster-permission")) || !viper.IsSet("has-cluster-permission") {
```
jpkrohling: When would it not be set? During the bootstrap?

pkg/upgrade/upgrade.go
```diff
@@ -21,11 +23,33 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
...
+		if err := reader.List(ctx, list, opts...); err != nil {
+			if strings.HasSuffix(err.Error(), "cluster scope") {
+				log.WithError(err).Warn("List indentity failed with cluster scope")
```
jpkrohling: `Listing of cluster-wide instances belonging to this operator failed.` is probably more descriptive.

jpkrohling:
@christine-gong, are you still interested in pursuing this one?

christine-gong:
@jpkrohling Yes, I am! Sorry I was out for vacation. Will address the comments tomorrow.

jpkrohling:
I'm closing this in favor of #916.

pkg/autodetect/main.go
munipravy: @jpkrohling  Can you please provide me necessary ClusterRole when WATCH_NAMESPACE is empty ?

PR #837, opened 2020-01-02
Configure keyspace in cassandra init job

### LINKED ISSUE #469: Connect to cassandra with credentials
Hello,

I'm trying to use jaeger with cassandra. Cassandra is secured by username and password. When I try to create a jaeger instance with the jaeger operator, jaeger cannot connect to cassandra. I get the following error message / logs:

`Connection error: ('Unable to connect to any servers', {'IP': AuthenticationFailed('Remote end requires authentication.',)})
Cassandra is still not up at jaeger-cassandra.cassandra-system.svc.cluster.local. Waiting 1 second.`

I use the following configuration:
```apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: cass-test
spec:
  strategy: production
  storage:
    type: cassandra
    options:
      cassandra:
        servers: jaeger-cassandra.cassandra-system.svc.cluster.local
        username: cassandra
        password: ***
        keyspace: jaeger_v1_datacenter3
  ingress:
    enabled: true´´´

How is the correct configuration with username and password for cassandra?

### LINKED ISSUE #822: Unable to set KEYPSPACE environment in cassandraCreateSchema
The create cassandra schema supports setting a KEYSPACE environment variable, it seems logical that this can be done in the `cassandraCreateSchema`. However it does not have any affect. this could be nice to have if for example datacenter name contains dashes instead of underscore

### PR DESCRIPTION
**Pull request by arpitjindal97**
Resolves #822 
Resolves #469 

Added the missing `KEYSPACE` variable in schema creation

### DIFF
--- a/pkg/storage/cassandra_dependencies.go
+++ b/pkg/storage/cassandra_dependencies.go
@@ -42,6 +42,12 @@ func cassandraDeps(jaeger *v1.Jaeger) []batchv1.Job {
 		host = "cassandra" // this is the default in the image
 	}
 
+	keyspace := jaeger.Spec.Storage.Options.Map()["cassandra.keyspace"]
+	if keyspace == "" {
+		jaeger.Logger().Info("Cassandra keyspace not specified. Using 'jaeger_v1_test' for the cassandra-create-schema job.")
+		keyspace = "jaeger_v1_test" // this is default in the image
+	}
+
 	username := jaeger.Spec.Storage.Options.Map()["cassandra.username"]
 	password := jaeger.Spec.Storage.Options.Map()["cassandra.password"]
 
@@ -116,6 +122,9 @@ func cassandraDeps(jaeger *v1.Jaeger) []batchv1.Job {
 							}, {
 								Name:  "DATACENTER",
 								Value: jaeger.Spec.Storage.CassandraCreateSchema.Datacenter,
+							}, {
+								Name:  "KEYSPACE",
+								Value: keyspace,
 							}, {
 								Name:  "CASSANDRA_USERNAME",
 								Value: username,

### CODE REVIEW

File: pkg/storage/cassandra_dependencies.go [hunk 1]
```diff
@@ -42,6 +42,12 @@ func cassandraDeps(jaeger *v1.Jaeger) []batchv1.Job {
 		host = "cassandra" // this is the default in the image
 	}
 
+	keyspace := jaeger.Spec.Storage.Options.Map()["cassandra.keyspace"]
+	if keyspace == "" {
+		jaeger.Logger().Info("Cassandra keyspace not specified. Using 'jaeger_v1_test' for the cassandra-create-schema job.")
+		keyspace = "jaeger_v1_test" // this is default in the image
```
pavolloffay: Keyspace should contain datacenter if it is specified 

https://github.com/jaegertracing/jaeger/blob/b2a159a67b3d937eb29fe2a98f5b05708d34f56a/plugin/storage/cassandra/schema/create.sh#L44

pavolloffay: Maybe we can rely on the script to provide the defaults. Have a look at `util.RemoveEmptyVars` function.

arpitjindal97: If we look at Jaeger from other way. One cassandra can be used by different application based on `KEYSPACE`. So, user should be given flexibility to define it.

pavolloffay:
Could you also modify the integration tests and define datacenter and keyspace?

pkg/storage/cassandra_dependencies.go
pavolloffay: You are right, I am fine allowing the full control on the keyspace assuming it can be fully specified.

PR #842, opened 2020-01-08
Support self provisioned ES in streaming strategy

pavolloffay:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,11 @@ e2e-tests-self-provisioned-es: prepare-e2e-tests deploy-es-operator
 	@echo Running Self provisioned Elasticsearch end-to-end tests...
 	@STORAGE_NAMESPACE=$(STORAGE_NAMESPACE) ES_OPERATOR_NAMESPACE=$(ES_OPERATOR_NAMESPACE) ES_OPERATOR_IMAGE=$(ES_OPERATOR_IMAGE) go test -tags=self_provisioned_elasticsearch ./test/e2e/... $(TEST_OPTIONS)
 
+.PHONY: e2e-tests-self-provisioned-es-kafka
+e2e-tests-self-provisioned-es-kafka: prepare-e2e-tests deploy-kafka-operator deploy-es-operator
+	@echo Running Self provisioned Elasticsearch and Kafka end-to-end tests...
+	@STORAGE_NAMESPACE=$(STORAGE_NAMESPACE) ES_OPERATOR_NAMESPACE=$(ES_OPERATOR_NAMESPACE) ES_OPERATOR_IMAGE=$(ES_OPERATOR_IMAGE) go test -tags=self_provisioned_elasticsearch_kafka ./test/e2e/... $(TEST_OPTIONS)
+
 .PHONY: e2e-tests-streaming
 e2e-tests-streaming: prepare-e2e-tests es kafka
 	@echo Running Streaming end-to-end tests...
@@ -195,8 +200,8 @@ storage:
 	@echo Creating namespace $(STORAGE_NAMESPACE)
 	@kubectl create namespace $(STORAGE_NAMESPACE) 2>&1 | grep -v "already exists" || true
 
-.PHONY: kafka
-kafka:
+.PHONY: deploy-kafka-operator
+deploy-kafka-operator:
 	@echo Creating namespace $(KAFKA_NAMESPACE)
 	@kubectl create namespace $(KAFKA_NAMESPACE) 2>&1 | grep -v "already exists" || true
 ifeq ($(OLM),true)
@@ -209,12 +214,9 @@ else
 	@sed 's/namespace: .*/namespace: $(KAFKA_NAMESPACE)/' deploy/test/kafka-operator.yaml | kubectl -n $(KAFKA_NAMESPACE) apply -f -  2>&1 | grep -v "already exists" || true
 	@kubectl set env deployment strimzi-cluster-operator -n ${KAFKA_NAMESPACE} STRIMZI_NAMESPACE="*"
 endif
-	@curl --location $(KAFKA_EXAMPLE) --output deploy/test/kafka-example.yaml
-	@kubectl -n $(KAFKA_NAMESPACE) apply -f deploy/test/kafka-example.yaml  2>&1 | grep -v "already exists" || true
 
-.PHONY: undeploy-kafka
-undeploy-kafka:
-	@kubectl delete --namespace $(KAFKA_NAMESPACE) -f deploy/test/kafka-example.yaml 2>&1 || true
+.PHONY: undeploy-kafka-operator
+undeploy-kafka-operator:
 ifeq ($(OLM),true)
 	@echo Skiping kafka-operator undeploy
 else
@@ -224,6 +226,17 @@ else
 	@kubectl delete clusterrolebinding strimzi-cluster-operator-topic-operator-delegation
 endif
 	@kubectl delete namespace $(KAFKA_NAMESPACE) 2>&1 || true
+
+.PHONY: kafka
+kafka: deploy-kafka-operator
+	@echo Creating namespace $(KAFKA_NAMESPACE)
+	@kubectl create namespace $(KAFKA_NAMESPACE) 2>&1 | grep -v "already exists" || true
+	@curl --location $(KAFKA_EXAMPLE) --output deploy/test/kafka-example.yaml
+	@kubectl -n $(KAFKA_NAMESPACE) apply -f deploy/test/kafka-example.yaml  2>&1 | grep -v "already exists" || true
+
+.PHONY: undeploy-kafka
+undeploy-kafka: undeploy-kafka-operator
+	@kubectl delete --namespace $(KAFKA_NAMESPACE) -f deploy/test/kafka-example.yaml 2>&1 || true
 
 .PHONY: clean
 clean: undeploy-kafka undeploy-es-operator
--- a/pkg/controller/jaeger/jaeger_controller.go
+++ b/pkg/controller/jaeger/jaeger_controller.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/jaegertracing/jaeger-operator/pkg/storage"
+
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -115,6 +115,11 @@ e2e-tests-self-provisioned-es: prepare-e2e-tests deploy-es-operator
 	@echo Running Self provisioned Elasticsearch end-to-end tests...
 	@STORAGE_NAMESPACE=$(STORAGE_NAMESPACE) ES_OPERATOR_NAMESPACE=$(ES_OPERATOR_NAMESPACE) ES_OPERATOR_IMAGE=$(ES_OPERATOR_IMAGE) go test -tags=self_provisioned_elasticsearch ./test/e2e/... $(TEST_OPTIONS)
 
+.PHONY: e2e-tests-self-provisioned-es-kafka
+e2e-tests-self-provisioned-es-kafka: prepare-e2e-tests deploy-kafka-operator
```
pavolloffay: @kevinearls could you please create a jira to run this in the internal build?

kevinearls: @pavolloffay  I've added a note to TRACING-806 to check once this is merged.  I think it will cover that case.

test/e2e/self_provisioned_elasticsearch_kafka_test.go
```diff
@@ -0,0 +1,160 @@
...
+	err = e2eutil.WaitForDeployment(t, fw.KubeClient, namespace, jaegerInstanceName+"-query", 1, retryInterval, timeout)
+	require.NoError(t, err, "Error waiting for query deployment")
+
+	err = WaitForDeployment(t, fw.KubeClient, namespace, jaegerInstanceName+"-ingester", 1, retryInterval, timeout)
+	require.NoError(t, err, "Error waiting for ingester deployment")
+
+	ProductionSmokeTest(jaegerInstanceName)
+}
+
+func (suite *SelfProvisionedTestSuite) TestValidateEsOperatorImage() {
```
kevinearls: @pavolloffay This may be redundant, as it is already run in self_provisioned_elasticsearch_test.  On the other hand it doesn't take very long, so leave it if you'd prefer.

pkg/strategy/streaming.go
```diff
@@ -25,7 +27,7 @@ import (
 	"github.com/jaegertracing/jaeger-operator/pkg/util"
 )
 
-func newStreamingStrategy(ctx context.Context, jaeger *v1.Jaeger) S {
+func newStreamingStrategy(ctx context.Context, jaeger *v1.Jaeger, es *storage.ElasticsearchDeployment) S {
```
jpkrohling: This doesn't seem to belong here... We can't expect to add a `cassandra`, `badger`, `in-memory` to the strategies. There has to be a better way to do this.

pkg/strategy/streaming.go
```diff
@@ -105,16 +100,66 @@ func newStreamingStrategy(ctx context.Context, jaeger *v1.Jaeger) S {
...
+	cDep := collector.Get()
+	queryDep := inject.Sidecar(jaeger, inject.OAuthProxy(jaeger, query.Get()))
+	var ingesterDep *appsv1.Deployment
+	if d := ingester.Get(); d != nil {
+		ingesterDep = d
+	}
 	manifest.dependencies = storage.Dependencies(jaeger)
 
+	// assembles the pieces for an elasticsearch self-provisioned deployment via the elasticsearch operator
+	if storage.ShouldDeployElasticsearch(jaeger.Spec.Storage) {
```
jpkrohling: Most of this new code seems to be duplicated from the production strategy. Can't we find a way to reuse it? Duplicating a couple of lines is fine, but this is a bit more than just a couple...

https://github.com/jaegertracing/jaeger-operator/blob/5999ea6f1057cc63d30888272e9f101a697da8f6/pkg/strategy/production.go#L106-L126

pkg/strategy/streaming.go
pavolloffay: The main motivation is to set path to cert generation scrip. Tests use a different path than binary.

pkg/strategy/streaming.go
pavolloffay: I can extract this to a func, but anyways the whole streaming strategy is mostly copy of production strategy with added ingester.

pkg/strategy/streaming.go
jpkrohling: Can this be done via viper, or added to the context? Sounds a bit outside of the scope to change the strategy signature just to be able to get some information internal to the auto-provisioning of one of the storages...

pkg/strategy/streaming.go
jpkrohling: I can understand the streaming strategy code being very similar to the production when it comes to the assembled objects (deployments, services, ...), as the final set of objects is quite similar, but the logic around the self-provisioning of ES could be abstracted away. For instance, here's how the auto-provisioning of Kafka happens:

https://github.com/jaegertracing/jaeger-operator/blob/5999ea6f1057cc63d30888272e9f101a697da8f6/pkg/strategy/streaming.go#L61-L66

The strategy itself needs to know only whether the provisioning should happen. The auto-provision procedure will accept a manifest, and return a modified manifest, which might or not contain the Kafka objects required to do the provisioning.

pkg/strategy/streaming.go
pavolloffay: So when creating the self-provisioned ES we need two things: path to cert generation script and more importantly secretes from the environment. 

The path could be done via viper (although there are downsides of using viper in the impl packages, it's cleaner to pass params directly if we are talking about a package down in the stack).

Anyways the production strategy is using exactly the same construction.

pkg/strategy/streaming.go
pavolloffay: I have moved it to a shared func.

pkg/strategy/streaming.go
pavolloffay: So when creating the self-provisioned ES we need two things: path to cert generation script and more importantly secretes from the environment.

The path could be done via viper (although there are downsides of using viper in the impl packages, it's cleaner to pass params directly if we are talking about a package down in the stack).

Anyways the production strategy is using exactly the same construction.

jpkrohling: It sounds reasonable to me to require access to a configuration value (via viper or not) if the path to the cert generation is required. In fact, we are using viper already in this source for other purposes:

https://github.com/jaegertracing/jaeger-operator/blob/5999ea6f1057cc63d30888272e9f101a697da8f6/pkg/strategy/streaming.go#L90-L98

pavolloffay: Given the same approach is used in the production strategy I would like to keep it consistent and if required do refactoring in a separate PR.

Anyways could you please review other parts of the PR?

jpkrohling: That part was also contentious when it was added. I'd rather do it right this time and fix the production strategy later, than to have it wrong in both places.

pavolloffay: I am happy to hear any suggestions on how to get around this.

At the moment we create this object which is passed to prod and streaming strategy:

```
es := &storage.ElasticsearchDeployment{Jaeger: jaeger, CertScript: esCertGenerationScript, Secrets: secrets}
```

`es` object holds three params jaeger cr, path to cert generation script and secrets from the environment. Jaeger object could be removed because it's also passed to the strategy.
To properly handle ES secrets we need the path to cert generation and existing secrets.

To completely remove this object we would have to use viper for the cert generation scrip path and pass k8s client to the strategy (or create a new client, I am not sure if SKD provides API to a globally cached client).

> It sounds reasonable to me to require access to a configuration value (via viper or not) if the path to the cert generation is required. In fact, we are using viper already in this source for other purposes:

Using viper in high-level packages might be fine, although I do not like to depend on it in low-level components. These should have clear API contracts. Then it's easier to test and understand external package requirements.

jpkrohling: The cert path seems to be something that is installation-dependent, rather than CR-dependent, so, it's value belongs to a config store instead of parameter to method calls related to the reconciliation loop.

So, the only thing that needs a solution is about the secrets. I assume you need it to get access to the username/password to use for ES, right? Is it the case, that there can only be one provisioned ES per cluster? If so, you could create a secret with the username/pass in the format that the deployments could consume, like described in #846.

pavolloffay: > I assume you need it to get access to the username/password to use for ES, right?

No, we don't use user/pass in self-provisioned ES. The secrets contain certificates. The logic is bit more complicated. IIRC the certs are copied to operator filesystem in case of operator restart (or rescheduling to another node?)

jpkrohling: Can't they be mounted in the file system directly from the secrets, like it's done in the Kafka auto-provisioning? 

https://github.com/jaegertracing/jaeger-operator/blob/5999ea6f1057cc63d30888272e9f101a697da8f6/pkg/strategy/streaming.go#L145-L173

pavolloffay: They are mounted to Jaeger filesystem. But for the use case I am mentioned we need certs in the operator filesystem and not Jaeger instance (see the operator in my previous comment). The cert generation script then assures certs are valid and not expired...

jpkrohling: > But for the use case I am mentioned we need certs in the operator filesystem and not Jaeger instance (see the operator in my previous comment). The cert generation script then assures certs are valid and not expired...

This is another signal that there's a mismatch between things that should be operator-scoped and operand-scoped. Renewal of the certs, as well as creating/managing the secrets are operations that should be made by the operator independent of the reconciliation logic. The reconciliation logic should only be about the operand.

pavolloffay: > Renewal of the certs, as well as creating/managing the secrets are operations that should be made by the operator independent of the reconciliation logic.

Why handling objects/dependencie/whatever is required for the operand part of the reconciliation? My understanding of reconciliation logic is to prepare objects/dependencie/whatever for the operand.

jpkrohling: I don't know, you said it yourself, but I probably misunderstood what you meant...

> IIRC the certs are copied to operator filesystem in case of operator restart (or rescheduling to another node?)

pavolloffay:
@jpkrohling I have moved the ES cert generation to jaeger controller as discussed on the call. Please have a look. If that is fine I will run e2e tests on OCP before merging.

pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -248,6 +237,31 @@ func (r *ReconcileJaeger) apply(ctx context.Context, jaeger v1.Jaeger, str strat
 		return jaeger, tracing.HandleError(err, span)
 	}
 
+	// ES cert handling requires secretes from environment
```
jpkrohling: `s/secretes/secrets`

pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/jaegertracing/jaeger-operator/pkg/storage"
```
jpkrohling: Was it placed here by `make format`?

pavolloffay: yes 

It either places it after core imports or after 3rd party imports in a separate block. However when I put it into operator imports it leaves it there.

PR #860, opened 2020-01-17
Skip this test when running on OpenShift OCP 4.1

kevinearls:

### DIFF
--- a/test/e2e/examples2_test.go
+++ b/test/e2e/examples2_test.go
@@ -3,6 +3,8 @@
 package e2e
 
 import (
+	"os"
+	"strings"
 	"testing"
 
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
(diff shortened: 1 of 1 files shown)


test/e2e/examples2_test.go
```diff
@@ -53,6 +55,13 @@ func (suite *ExamplesTestSuite2) TestSimplestExample() {
 }
 
 func (suite *ExamplesTestSuite2) TestWithBadgerExample() {
+	if isOpenShift(t) {
+		ocpVersion := os.Getenv("OCP_VERSION")
+		if len(ocpVersion) > 0 && strings.HasPrefix(ocpVersion, "4.1") {
```
pavolloffay: Can you explain why it is skipped only on OCP 4.1?

kevinearls: At the moment it's only failing on 4.1 and we need to decide how much time we want to invest in investigating this, as we probably won't be supporting OCP 4.3 for much longer

pavolloffay: so It was passing before on 4.1 but not it is failing?

kevinearls:
I've opened issue #870 to track this problem.

test/e2e/examples2_test.go
kevinearls: Yes, that's correct.

kevinearls:
No longer applicable

Issue #863, opened 2020-01-19
Jaeger tags for injected agent are only added if a single container is present


TBBle:
I'm specifically looking at `func container` in https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/inject/sidecar.go.

The flow *appears* to be intended to inject a bunch of tags from the pod info, and *if a single container is present*, include a tag about the container.

However, the entire set of tags is only added in a single-container setup.

This looks like a trivial mistake in #648.

TBBle:
I don't have an immediate fix for it, but I suspect the 'only one container' logic will also fail if called from `UpdateSidecar`, as it the jaeger-agent side-car will be seen as a second container.

jpkrohling:
Nice catch!

PR #864, opened 2020-01-19
Auto-inject agent tags in multi-container pods

TBBle:
Fixes: #863

### DIFF
--- a/pkg/inject/sidecar_test.go
+++ b/pkg/inject/sidecar_test.go
@@ -439,6 +439,17 @@ func TestSidecarWithPrometheusAnnotations(t *testing.T) {
 	assert.Equal(t, dep.Annotations["prometheus.io/port"], "9090")
 }
 
+func TestSidecarAgentTagsWithMultipleContainers(t *testing.T) {
+	jaeger := v1.NewJaeger(types.NamespacedName{Name: "TestSidecarAgentTagsWithMultipleContainers"})
+	dep := Sidecar(jaeger, depWithTwoContainers(map[string]string{Annotation: jaeger.Name}, map[string]string{}))
+
+	assert.Len(t, dep.Spec.Template.Spec.Containers, 3, "Expected 3 containers")
+	assert.Equal(t, "jaeger-agent", dep.Spec.Template.Spec.Containers[2].Name)
+	containsOptionWithPrefix(t, dep.Spec.Template.Spec.Containers[2].Args, "--jaeger.tags")
+	agentTags := agentTags(dep.Spec.Template.Spec.Containers[2].Args)
+	assert.Equal(t, "", util.FindItem("container.name=", agentTags))
+}
+
 func dep(annotations map[string]string, labels map[string]string) *appsv1.Deployment {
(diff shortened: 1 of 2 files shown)


pkg/inject/sidecar_test.go
```diff
@@ -474,3 +498,42 @@ func containsOptionWithPrefix(t *testing.T, args []string, prefix string) bool {
 	assert.Fail(t, "list of arguments didn't have an option starting with '%s'", prefix)
 	return false
 }
+
+func containsAgentTag(t *testing.T, args []string, tag string, value string) bool {
```
jpkrohling: We have a `util.FindItem(prefix string, args []string)` already that you could make use of to get the agent tag. With that, it's probably sufficient to use something like `assert.Contains(t, tag, value)`.

pkg/inject/sidecar_test.go
```diff
@@ -308,6 +309,7 @@ func TestSidecarOrderOfArguments(t *testing.T) {
 	containsOptionWithPrefix(t, dep.Spec.Template.Spec.Containers[1].Args, "--jaeger.tags")
 	containsOptionWithPrefix(t, dep.Spec.Template.Spec.Containers[1].Args, "--reporter.grpc.host-port")
 	containsOptionWithPrefix(t, dep.Spec.Template.Spec.Containers[1].Args, "--reporter.type")
+	containsAgentTag(t, dep.Spec.Template.Spec.Containers[1].Args, "container.name", "only_container")
```
jpkrohling: I think you can replace most of the new code in this test with this:

```golang
assert.Contains(t, util.FindItem("--jaeger.tags=", dep.Spec.Template.Spec.Containers[1].Args), "container.name=only_container")
```

pkg/inject/sidecar_test.go
```diff
@@ -474,3 +498,42 @@ func containsOptionWithPrefix(t *testing.T, args []string, prefix string) bool {
...
+			assert.Fail(t, "agent tag '%s=%s' not found", tag, value)
+			return false
+		}
+	}
+	assert.Fail(t, "list of arguments didn't have an option starting with '%s'", prefix)
+	return false
+}
+
+func doesNotContainAgentTag(t *testing.T, args []string, tag string, value string) bool {
+	prefix := "--jaeger.tags="
```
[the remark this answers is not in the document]
TBBle: Won't the `t.Fail` calls then occur in the "pass" flow?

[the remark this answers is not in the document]
jpkrohling: True. Instead of having the assert.Fail inside the `containsAgent...` function, you can keep the assertion only at the top-level functions, like:

```golang
func TestBla(t *testing.T)  {
  assert.True(t, containsAgentTag(...))
}
```

pkg/inject/sidecar_test.go
TBBle: Ah, drat. I saw the first two comments, and then went ahead and refactored it, and ended up with this but slightly long-hand.

jpkrohling:
Thanks for your PR!

Issue #855, opened 2020-01-16
Duplicated "Injecting sidecar" message


jpkrohling:
When deploying an application that should have a sidecar injected, the following message is duplicated:

```
INFO[0270] Injecting Jaeger Agent sidecar                deployment=tracegen jaeger=simplest jaeger-namespace=default namespace=default
INFO[0270] Injecting Jaeger Agent sidecar                deploymentName=tracegen deploymentNamespace=default instance=simplest namespace=default
```

Using  the `simple-prod` CR as example, the following deployment can be used to reproduce:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: tracegen
  annotations:
    "sidecar.jaegertracing.io/inject": "true"
spec:
  selector:
    matchLabels:
      app: tracegen
  template:
    metadata:
      labels:
        app: tracegen
    spec:
      containers:
      - name: tracegen
        image: quay.io/jpkroehling/jaeger-tracegen:latest
        args: ["-duration", "10s", "-workers", "1"]
```

jpkrohling:
@rubenvp8510 could you please take a look?

rubenvp8510:
Sure, I'll take a look

rubenvp8510:
This is due to https://github.com/jaegertracing/jaeger-operator/pull/454 

Reading the original [issue](https://github.com/jaegertracing/jaeger-operator/issues/442) it seems to me like we shouldn't implement this using the auto-detect feature, instead I think we should scan for updates only when operator starts and let the reconciliation loop do the rest.

@jpkrohling let me know what do you think, I can do the change and send a PR.

Issue #442, opened 2019-05-28
Scan deployments for agent injection


jpkrohling:
When a Jaeger instance is installed via a given CR and the operator is then stopped, new deployments with the auto-injection annotation are ignored until the deployment has an update.

To solve this, the operator should check for all deployments when it first starts, injecting an agent when needed.

jpkrohling:
From #427

jpkrohling:
@rubenvp8510 I think this case here is covered, right?

rubenvp8510:
@jpkrohling yes!

PR #869, opened 2020-01-20
Remove deployment updates from autodetect loop

rubenvp8510:

In https://github.com/jaegertracing/jaeger-operator/pull/454 we introduced the detection/update of deployments to address https://github.com/jaegertracing/jaeger-operator/issues/442, But after seen this closely, I think this is not the right approach. We should run the scan only at start.

More than that, In my latest tests debugging for fix https://github.com/jaegertracing/jaeger-operator/issues/855.  I see that we don't need to run the scan any more (operator-sdk 0.12.0) , because for some reason (I'm still not sure why) the reconciliation run on each deployment at bootstrap time (see logs). Not sure when this behaviour changed.

[bootstrap-operator.txt](https://github.com/jaegertracing/jaeger-operator/files/4088376/bootstrap-operator.txt)

### DIFF
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -78,64 +78,6 @@ func (b *Background) Stop() {
 	b.ticker.Stop()
 }
 
-func (b *Background) requireUpdates(deps *appsv1.DeploymentList) []*appsv1.Deployment {
-	instances := &v1.JaegerList{}
-	if err := b.clReader.List(context.Background(), instances); err != nil {
-		log.WithError(err).Info("failed to retrieve the list of Jaeger instances")
-		return nil
-	}
-
-	requireUpdates := make([]*appsv1.Deployment, 0)
-	for i := 0; i < len(deps.Items); i++ {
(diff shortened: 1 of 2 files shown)


kevinearls:
@objectiser @jpkrohling @rubenvp8510 What is the status of this?  E2E tests are broken on OpenShift without this.

jpkrohling:
@rubenvp8510 please, find evidence that this new behavior is the intended one, rather than a bug that has been introduced in the latest Operator SDK versions.

rubenvp8510:
I asked on kubernetes-operator slack channel.

Ruben Vargas Today at 7:38 AM
Hi, question, I'm developing an operator that watches deployments and inject sidecars on deployments containing certain annotations. If the operator is not running and a new deployment is created, is an expected behavior when the operator runs the first time it scan all deployments when it start watching for it?
1 replies
joelanford  3 hours ago
Yep, that’s correct.Also btw, this sounds like a use case for a mutating admission webhook. If you haven’t already, you may want to consider that as an option as well.

PR #871, opened 2020-01-22
Auto-inject the IP tag for operator-injected agent

### LINKED ISSUE #865: Auto-inject the IP tag for Operator-injected Agents
The Operator provides a few [default tags for Jaeger Agents](https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/inject/sidecar.go#L134) when they are deployed as auto-injected side-cars.

Since a Pod is limited to a single host, it'd be useful to inject the 'ip' tag the same way. This would make life much simpler for non-Jaeger clients (such as the OpenCensus exporter) and saves having to include code to discover the hostIP in every k8s-hosted client, since the default IP for the networking stack is almost certainly for the container's network namespace, not the actual host IP.

This would help alleviate jaegertracing/jaeger#1459 by ensuring that at-least such Pods do not get false results from the Clock Skew Adjuster because they are seen as different hosts, even though they are different pods on the same hardware.

This probably isn't ~trivial~ hard, I believe this requires that the Jaeger Agent container have an env-var populated from its status.hostIP, and then that can be passed through via the command-line.

rubenvp8510:
Fixes #865 

### DIFF
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -34,6 +34,7 @@ const (
 	envVarServiceName = "JAEGER_SERVICE_NAME"
 	envVarPropagation = "JAEGER_PROPAGATION"
 	envVarPodName     = "POD_NAME"
+	envVarHostIP      = "HOST_IP"
 )
 
 // Sidecar adds a new container to the deployment, connecting to the given jaeger instance
@@ -132,11 +133,12 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 	jgBinaryTrft := util.GetPort("--processor.jaeger-binary.server-host-port=", args, 6832)
 
 	if len(util.FindItem("--jaeger.tags=", args)) == 0 {
-		agentTags := fmt.Sprintf("%s=%s,%s=%s,%s=%s,%s=%s",
+		agentTags := fmt.Sprintf("%s=%s,%s=%s,%s=%s,%s=%s,%s=%s",
 			"cluster", "undefined", // this value isn't currently available
 			"deployment.name", dep.Name,
 			"pod.namespace", dep.Namespace,
 			"pod.name", fmt.Sprintf("${%s:}", envVarPodName),
+			"host.ip", fmt.Sprintf("${%s:}", envVarHostIP),
 		)
 
 		if len(dep.Spec.Template.Spec.Containers) == 1 {
@@ -158,14 +160,24 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 		Image: util.ImageName(jaeger.Spec.Agent.Image, "jaeger-agent-image"),
 		Name:  "jaeger-agent",
 		Args:  args,
-		Env: []corev1.EnvVar{{
-			Name: envVarPodName,
-			ValueFrom: &corev1.EnvVarSource{
-				FieldRef: &corev1.ObjectFieldSelector{
-					FieldPath: "metadata.name",
+		Env: []corev1.EnvVar{
+			{
+				Name: envVarPodName,
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{
+						FieldPath: "metadata.name",
+					},
 				},
 			},
-		}},
+			{
+				Name: envVarHostIP,
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{
+						FieldPath: "status.hostIP",
+					},
+				},
+			},
+		},
 		Ports: []corev1.ContainerPort{
 			{
 				ContainerPort: zkCompactTrft,
--- a/pkg/inject/sidecar_test.go
+++ b/pkg/inject/sidecar_test.go
@@ -58,6 +58,7 @@ func TestInjectSidecarWithEnvVars(t *testing.T) {
 	assert.Len(t, dep.Spec.Template.Spec.Containers, 2)
 	assert.Contains(t, dep.Spec.Template.Spec.Containers[1].Image, "jaeger-agent")
 	containsEnvVarNamed(t, dep.Spec.Template.Spec.Containers[1].Env, envVarPodName)
+	containsEnvVarNamed(t, dep.Spec.Template.Spec.Containers[1].Env, envVarHostIP)
 
 	assert.Len(t, dep.Spec.Template.Spec.Containers[0].Env, 2)
 	assert.Contains(t, dep.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: envVarPropagation, Value: "jaeger,b3"})


pkg/inject/sidecar.go
```diff
@@ -158,14 +160,24 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
...
+						FieldPath: "metadata.name",
+					},
 				},
 			},
-		}},
+			{
+				Name: envVarPodIp,
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{
+						FieldPath: "status.podIP",
```
jpkrohling: I think the original request was to have the _host_ IP, not the pod. IIRC, the pod IP will already be reported as part of the process tags.

cc @TBBle

rubenvp8510: Haaa you are totally right! Changed.

pkg/inject/sidecar.go
```diff
@@ -132,11 +133,12 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 	jgBinaryTrft := util.GetPort("--processor.jaeger-binary.server-host-port=", args, 6832)
 
 	if len(util.FindItem("--jaeger.tags=", args)) == 0 {
-		agentTags := fmt.Sprintf("%s=%s,%s=%s,%s=%s,%s=%s",
+		agentTags := fmt.Sprintf("%s=%s,%s=%s,%s=%s,%s=%s,%s=%s",
 			"cluster", "undefined", // this value isn't currently available
 			"deployment.name", dep.Name,
 			"pod.namespace", dep.Namespace,
 			"pod.name", fmt.Sprintf("${%s:}", envVarPodName),
+			"host.ip", fmt.Sprintf("${%s:}", envVarHostIP),
```
TBBle: For #865 this should be the `ip` tag, specifically so that the Clock Skew Adjustor would recognise [different spans from the same host as the same](https://github.com/jaegertracing/jaeger/blob/master/model/adjuster/clockskew.go#L78), and not completely mess-up the spans if two k8s nodes were out of clock-sync.

rubenvp8510: changed.

jpkrohling: This needs confirmation, but I don't think that's how it works. The `ip` tag that you see is under `process` and is gathered by the tracer and I don't think we have any influence on that in most cases...

TBBle: According to [Agent level tags](https://www.jaegertracing.io/docs/1.16/deployment/#agent-level-tags), tags passed to `--jaeger.tags` are
> added to the *process* tags of all spans passing through the agent.
(Emphasis mine)

TBBle: Given
```yaml
kind: "Jaeger"
metadata:
  name: "hampy-jaeger"
  namespace: "observability"
spec:
  strategy: allInOne
  ingress:
    hosts:
      - hampy-jaeger.dev.kepler.wgsyd
  agent:
    options:
        jaeger.tags: injected_agent_tag=true
```

image

Note: The `ip` process tag is set in our code, to thwart the Clock Skew Adjustor, pending this change.

jpkrohling: Decided to give it a try myself, and confirmed that this isn't going to work:

image

Note how the `ip` tag is duplicated, as the `ip` tag is added by the client automatically. We had a similar case of duplicate tags in the collector, but we can't de-dup it there, as the tags have different values, we can't know which one is right.

IMO, if we want to have the `hostIP` for a given span, its name should be qualified so that it's not ambiguous. Something like @rubenvp8510 did before, `host.ip`. This won't fix the clock skew problem, but neither will the current state of this PR.

Note that I had to apply the fix from #864 to get this tested.

TBBle: Drat. I had implied from the docs that the values would be overwritten by the agent, but did not test that flow in my earlier test.

I don't have this issue here because I'm using OpenCensus, which does not add the `ip` tag automatically.

`host.ip` is consistent, but sadly doesn't resolve #865, which is perhaps not-doable, as demonstrated.

It's a shame because the Client library is providing the wrong IP for the Clock Skew Adjustor's algorithm.

I guess the best hope of progress would be if the Clock Skew Adjustor supported a configurable tag. Then I could specify `host.ip` knowing it was always available in agent-injected pods.

jpkrohling: As consumer of the client libraries, what you can do is initialize the tracer [with a set of tags](https://github.com/jaegertracing/jaeger-client-java/blob/74adce4c4b4c507ccfb010ffb06c95973cc97c6a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java#L653-L658), which might include the IP with the appropriate hostIP. If that tag is already set by the time the tracer is initialized, that value will be used instead:

https://github.com/jaegertracing/jaeger-client-java/blob/74adce4c4b4c507ccfb010ffb06c95973cc97c6a/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java#L109-L125

TBBle: That's what I'm doing now, visible in my screenshot above. The effort is simply that we aren't otherwise exposing the Host IP to the pods, so we'll have to plumb that through into each pod to pass into the trace initialisation, which means a half-dozen two-file updates to various services.

Thinking about it, since in my case I don't have the conflict over the `ip` tag, once this change goes in I could put in my Jaeger something like
```yaml
spec:
  agent:
    options:
        jaeger.tags: ip=${HOST_IP:}
```

although because the jaeger.tags option's presence hides all the auto-added tags, I then lose the other auto-added process tags, which would be disappointing.

I think I'll stick with the constant string to defang the Clock Skew Adjuster, and my next effort to get the Clock Skew Adjuster doing the right thing will be to get it to honour the new `host.ip` tag instead of an `ip` tag for its machine-key.

Still, even without the Clock Skew Adjuster use-case I opened #865 for, `host.ip` is a valuable process tag to add.

pkg/inject/sidecar.go
```diff
@@ -132,11 +133,12 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 	jgBinaryTrft := util.GetPort("--processor.jaeger-binary.server-host-port=", args, 6832)
 
 	if len(util.FindItem("--jaeger.tags=", args)) == 0 {
-		agentTags := fmt.Sprintf("%s=%s,%s=%s,%s=%s,%s=%s",
+		agentTags := fmt.Sprintf("%s=%s,%s=%s,%s=%s,%s=%s,%s=%s",
 			"cluster", "undefined", // this value isn't currently available
 			"deployment.name", dep.Name,
 			"pod.namespace", dep.Namespace,
 			"pod.name", fmt.Sprintf("${%s:}", envVarPodName),
+			"ip", fmt.Sprintf("${%s:}", envVarHostIP),
```
TBBle: So it's not lost in the other thread, this needs to go back to `host.ip`, as when a tracer library submits traces with a process tag `ip`, the resulting trace will end up with two such tags, rather than one suppressing the other as I had assumed.

rubenvp8510:
After read all the discussion I reverted the last change , back to `host.ip` tag name.

TBBle:
I've be fine with having the env-var and manually using it, but it'd be nice if this was either the case for *all* the parameters, or the agent args in the `Jaeger` CRD were *additive*, so I don't lose the rest of the magic info. I really don't want to have to teach my processes about their own deployment name, namespace, etc, but I don't mind a template `Jaeger` resource I can give users that enables those values.

My preference is to keep 'em all, or trim ones that aren't valuable though. Less configuration (and hence less teaching of configuration) makes me a happier developer. ^_^

objectiser:
@TBBle Agree it might be nice to add new `jaeger-tags` rather than replacing the existing set when the option is explicitly used in the CRD. Only concern is that then there is no way for a user to remove those tags if they are not wanted.

jpkrohling:
I'd be OK with merging this PR as is, and see if we have a need for further tags in the future. This might be the last one we can automatically add anyway...

objectiser:
The other thing to keep an eye on is OpenTelemetry: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-resource-semantic-conventions.md

IP attribute is not covered currently, but `host.ip` would fit in with convention used in [this section](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-resource-semantic-conventions.md#host).

TBBle:
That document suggests to me that we should be (later...) exposing all these attributes as env-vars, and then `--jaeger.tags` could be user-provided to choose the tag names, but wouldn't need access to the data.

That said, I understand OpenTelemetry has plans for their own agents, so it might never be a super-interesting use-case for this part of the Jaeger Operator, but remains academically interesting.

jpkrohling:
@TBBle, @objectiser : what's the verdict then?

objectiser:
I think the agreement was to merge as-is, unless you have changed your mind?

jpkrohling:
No, just wanted to confirm before merging :-)

PR #874, opened 2020-01-27
Fix addition of agent tags in deployment of multiple containers

### LINKED ISSUE #863: Jaeger tags for injected agent are only added if a single container is present

annanay25:
Added tests. Fixes #863.

Signed-off-by: Annanay <annanayagarwal@gmail.com>


jpkrohling:
This is already being addressed by #864. I'm closing it for now in that favor, as I think that one will get merged soon.

PR #881, opened 2020-02-03
Delete PVCs from auto provisioned Jaeger at end of test

kevinearls:

Delete PVCs at the end of tests which create Kafka instances with persistent storage.   This should not strictly be necessary, as the PVCs should be deleted when the namespace is deleted.  

However, we have had times where a subsequent test gets a panic, particularly because of port-forwarding problems, and then the namespace doesn't get deleted.  This can cause problems with some of our clusters which do not have a lot of resources for volumes.

### DIFF
--- a/test/e2e/utils.go
+++ b/test/e2e/utils.go
@@ -414,3 +414,14 @@ func createSecret(secretName, secretNamespace string, secretData map[string][]by
 	WaitForSecret(secretName, secretNamespace)
 	return createdSecret
 }
+
+func deletePersistentVolumeClaims(namespace string) {
+	pvcs, err := fw.KubeClient.CoreV1().PersistentVolumeClaims(kafkaNamespace).List(metav1.ListOptions{})
+	require.NoError(t, err)
+
+	emptyDeleteOptions := metav1.DeleteOptions{}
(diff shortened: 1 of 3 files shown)


test/e2e/utils.go
```diff
@@ -414,3 +414,14 @@ func createSecret(secretName, secretNamespace string, secretData map[string][]by
 	WaitForSecret(secretName, secretNamespace)
 	return createdSecret
 }
+
+func deletePersistentVolumeClaims(namespace string) {
+	pvcs, err := fw.KubeClient.CoreV1().PersistentVolumeClaims(kafkaNamespace).List(metav1.ListOptions{})
```
jpkrohling: What happens to the volumes? Are they deleted automatically?

kevinearls: Normally they are deleted when the namespace is deleted, but not when the Kafka instance is deleted.  Deleting them in this function just covers some edge cases where the namespace for a test may not get deleted because a subsequent test panics, or something of that sort.  

This has caused us a few problems on internal CI.

kevinearls: @jpkrohling Yes, if you delete the pvc the pv is deleted along with it.

PR #883, opened 2020-02-04
Preserve generated cookie secret on the reconciliation process

rubenvp8510:
This is another patch for prevent multiple creations of deployments  when a jaeger resource is created.

// cc: @kevinearls 

I only tested with all-in-one strategy, I'll test tomorrow with-cassandra yaml file

### DIFF
--- a/pkg/inject/oauth_proxy.go
+++ b/pkg/inject/oauth_proxy.go
@@ -109,11 +109,11 @@ func PropagateOAuthCookieSecret(specSrc, specDst appsv1.DeploymentSpec) appsv1.D
 			break
 		}
 	}
-	// Found the cooke secretArg parameter, replace argument.
+	// Found the cookie secretArg parameter, replace argument.
 	if secretArg != "" {
 		for i, container := range spec.Template.Spec.Containers {
 			if container.Name == "oauth-proxy" {
-				util.ReplaceArgument("--cookie-secret", secretArg, spec.Template.Spec.Containers[i].Args, true)
+				util.ReplaceArgument("--cookie-secret", secretArg, spec.Template.Spec.Containers[i].Args)
 			}
 		}
 	}
--- a/pkg/inject/oauth_proxy.go
+++ b/pkg/inject/oauth_proxy.go
@@ -99,7 +99,7 @@ func getOAuthProxyContainer(jaeger *v1.Jaeger) corev1.Container {
 }
 
 //PropagateOAuthCookieSecret preserve the generated oauth cookie across multiple reconciliations
-func PropagateOAuthCookieSecret(specSrc, specDst *appsv1.DeploymentSpec) appsv1.DeploymentSpec {
+func PropagateOAuthCookieSecret(specSrc, specDst appsv1.DeploymentSpec) appsv1.DeploymentSpec {
 	spec := specDst.DeepCopy()
 	secretArg := ""
 	// Find secretArg from old object
@@ -113,7 +113,7 @@ func PropagateOAuthCookieSecret(specSrc, specDst *appsv1.DeploymentSpec) appsv1.
 	if secretArg != "" {
 		for i, container := range spec.Template.Spec.Containers {
 			if container.Name == "oauth-proxy" {
-				util.ReplaceArgument("--cookie-secret", secretArg, spec.Template.Spec.Containers[i].Args)
+				util.ReplaceArgument("--cookie-secret", secretArg, spec.Template.Spec.Containers[i].Args, true)
 			}
 		}
 	}
--- a/pkg/inject/oauth_proxy.go
+++ b/pkg/inject/oauth_proxy.go
@@ -3,6 +3,7 @@ package inject
 import (
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/spf13/viper"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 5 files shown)


pkg/inject/oauth_proxy.go
```diff
@@ -97,3 +98,35 @@ func getOAuthProxyContainer(jaeger *v1.Jaeger) corev1.Container {
 		Resources: commonSpec.Resources,
 	}
 }
+
+//PreserveOauthCookieSecret preserve the generated oauth cookie across multiple reconciliations
+func PreserveOauthCookieSecret(specSrc, specDst *appsv1.DeploymentSpec) {
```
jpkrohling: This could have a better name. Perhaps `PropagateOAuthCookieSecret`? Or just `HandleOAuthCookieSecret`?

jpkrohling: In recent changes, I've been trying to avoid those pointers, returning the resulting spec. You could also do it, so that your caller would look like this:

```
tp.Spec = inject.PropagateOAuthCookieSecret(&t.Spec, &tp.Spec)
```

rubenvp8510: Sounds reasonable. I would prefer to not modify funcion parameters , make it pure function.

About the names, I'll rename it. I just put a the first name came to my mind just to test if this  fix the issue.

Thanks.

jpkrohling:
By the way, the fix itself looks correct, as @kevinearls confirmed during his tests. I could also confirm that the several deployment versions were caused by this:

```
$ kubectl rollout history deployment/with-cassandra --revision=1 > /tmp/deployment-r1.yaml
$ kubectl rollout history deployment/with-cassandra --revision=2 > /tmp/deployment-r2.yaml
$ diff /tmp/deployment-r*
1c1
< deployment.extensions/with-cassandra with revision #1
---
> deployment.extensions/with-cassandra with revision #2
9c9
< 	pod-template-hash=59c7df5c79
---
> 	pod-template-hash=57865b948b
38c38
<       --cookie-secret=rcoiF6NixkQzKbhJFTJcEw==
---
>       --cookie-secret=yCD2nT9xw64DYqFKGxUYJw==
```

pkg/inject/oauth_proxy.go
```diff
@@ -97,3 +98,35 @@ func getOAuthProxyContainer(jaeger *v1.Jaeger) corev1.Container {
...
+			}
+		}
+	}
+
+	// Found the cooke secret parameter.
+	if secret != "" {
+	destinationLoop:
+		for index, container := range specDst.Template.Spec.Containers {
+			if container.Name == "oauth-proxy" {
+				for argIndex, arg := range container.Args {
```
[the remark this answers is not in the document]
rubenvp8510: I use` util.FindItem` above but here `util.FindItem` only return the value if found it, in this case I'm interested on modify the value. I don't want to modify `util.FindItem` to return index, as is used in other parts of the code.

kevinearls:
I ran the full set of E2E tests on this on OCP 4.2 and they all passed.

rubenvp8510:
I've already address comments. If no other comment or observation I think this is ready to merge.

pkg/inventory/deployment.go
```diff
@@ -28,6 +29,7 @@ func ForDeployments(existing []appsv1.Deployment, desired []appsv1.Deployment) D
 
 			// we can't blindly DeepCopyInto, so, we select what we bring from the new to the old object
 			tp.Spec = v.Spec
+			tp.Spec = inject.PropagateOAuthCookieSecret(&t.Spec, &tp.Spec)
```
jpkrohling: You can have both lines 31 and 32 done on the same line:

```
tp.Spec = inject.PropagateOAuthCookieSecret(&t.Spec, v.Spec)
```

You'll need to change your arguments to accept values instead of pointers, which is in line with what we discussed in a previous review :)

pkg/util/util.go
```diff
@@ -176,6 +176,16 @@ func FindItem(prefix string, args []string) string {
 	return ""
 }
 
+// ReplaceArgument replace argument value with given value.
+func ReplaceArgument(prefix string, newValue string, args []string) {
+	for argIndex, arg := range args {
```
jpkrohling: What happens if `prefix` isn't found? And when multiple prefixes exist? You'll need tests to assert those cases.

This method could return a `bool`, signaling whether there was a replacement.

pkg/util/util.go
```diff
@@ -176,6 +176,21 @@ func FindItem(prefix string, args []string) string {
 	return ""
 }
 
+// ReplaceArgument replace argument value with given value.
+func ReplaceArgument(prefix string, newValue string, args []string, firstOccurrence bool) int {
```
jpkrohling: I appreciate the idea behind the `firstOccurrence`, but there's no need for it, at least for the moment. If you decide to keep this type of feature, you can follow the idea behind Go's `strings.Replace`, where the last parameter is the max number of times the replace should be applied, with a negative meaning "replace all occurrences".

https://golang.org/src/strings/strings.go?s=23551:23597#L918

pkg/inject/oauth_proxy_test.go
```diff
@@ -5,13 +5,15 @@ import (
 	"sort"
 	"testing"
 
+	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
```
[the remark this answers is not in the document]
rubenvp8510: mmm.. it is.

Not sure why but this happens when I do a `make format`.

pkg/util/util.go
rubenvp8510: I'll remove the feature, as you said is not necessary for now.

rubenvp8510:
Done with observations.
Thanks

pkg/inject/oauth_proxy_test.go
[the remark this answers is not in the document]
rubenvp8510: I changed this line and the` make check` don't pass. :(

PR #897, opened 2020-02-07
Per namespace agent injection

### LINKED ISSUE #373: Per-namespace inject annotation
Looking at the istio documentation, I learned that we can annotate namespaces (seems obvious in hindsight). The operator could then support looking at the namespace's annotations to determine whether auto-injection should happen or not.

https://istio.io/docs/setup/kubernetes/install/kubernetes/

pavolloffay:
Resolves #373 

This adds a capability to auto inject Jaeger agent to namespaces annotated with `sidecar.jaegertracing.io/inject` - the same annotation which is used for deployments. 

It brings a complication when the annotation is present on both namespace and deployment. Then  the deployment has higher precedence, however if the specified instance on deployment is not found it does not fallback to the instance specified on the namespace.

## Other notable changes
* added `namespaces` to cluster role binding
* tests now use cluster role binding from `deploy/` directory

There was a discussion that Jaeger should not require access to the cluster resources and this PR adds the requirement on accessing `Namespaces`. At the moment the operator requires access to `ClusterRoleBinding`. To make the operator work without cluster-wide RBAC it will need more refactoring in a separate PR - related to https://github.com/jaegertracing/jaeger-operator/issues/791.

Once merged

- [ ] Documentation

My test notes:
```
oc new-project project33
oc create deployment vertx --image=jaegertracing/vertx-create-span:operator-e2e-tests

oc annotate namespace project33 sidecar.jaegertracing.io/inject=true
oc annotate deployment vertx sidecar.jaegertracing.io/inject=true 

kubectl create namespace project33
kubectl config set-context minikube --namespace=project33  
```


Makefile
```diff
@@ -82,18 +82,22 @@ unit-tests:
...
 	@echo "---" >> deploy/test/namespace-manifests.yaml
 
-	@cat test/role.yaml >> deploy/test/namespace-manifests.yaml
-	@echo "---" >> deploy/test/namespace-manifests.yaml
+	@cat test/operator.yaml | sed "s~image: jaegertracing\/jaeger-operator\:.*~image: $(BUILD_IMAGE)~gi" >> deploy/test/namespace-manifests.yaml
 
-	@cat test/service_account.yaml >> deploy/test/namespace-manifests.yaml
-	@echo "---" >> deploy/test/namespace-manifests.yaml
+.PHONY: e2e-tetsts-prepare-global-manifests
+e2e-tests-prepare-global-manifests: crd build docker push
```
jpkrohling: If I run `make e2e-tests-prepare-global-manifest`, should I expect `crd build docker push` to also be executed?

deploy/role.yaml
```diff
@@ -17,6 +16,7 @@ rules:
   - configmaps
   - secrets
   - serviceaccounts
+  - namespaces
```
jpkrohling: It might be better to have `namespaces` in a separate group, with permissions to only `list` and read`.

deploy/role_binding.yaml
```diff
@@ -2,7 +2,6 @@ kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: jaeger-operator
-  namespace: observability
```
jpkrohling: If the namespace in the role_binding from the deploy directory is being removed, it should be removed from the other files in the directory as well (operator.yaml, service_account.yaml). The readme has to be changed as well, to include `-n observability` to the `kubectl` commands.

cc @objectiser : I think this came from a change of yours. Do you remember why you needed this?

pkg/inject/sidecar_test.go
```diff
@@ -195,77 +196,145 @@ func TestSkipInjectSidecar(t *testing.T) {
...
-			},
+func TestSelect(t *testing.T) {
+	jTest := v1.NewJaeger(types.NamespacedName{Name: "test"})
+	jProd := v1.NewJaeger(types.NamespacedName{Name: "prod"})
+	tests := []struct {
+		dep      *appsv1.Deployment
+		ns       *corev1.Namespace
+		jaegers  *v1.JaegerList
+		expected *v1.Jaeger
+		cap      string
```
jpkrohling: What does `cap` stands for?

pkg/inject/sidecar_test.go
```diff
@@ -195,77 +196,145 @@ func TestSkipInjectSidecar(t *testing.T) {
...
+			jaegers:  &v1.JaegerList{Items: []v1.Jaeger{*jProd, *jTest}},
+			expected: jProd,
+			cap:      "dep explicit, ns explicit",
+		},
+		{
+			dep:      dep(map[string]string{Annotation: "doesNotExist"}, map[string]string{}),
+			ns:       ns(map[string]string{Annotation: "test"}),
+			jaegers:  &v1.JaegerList{Items: []v1.Jaeger{*jProd, *jTest}},
+			expected: nil,
+			cap:      "dep explicit not exists, ns explicit",
```
jpkrohling: `s/not exists/does not exist/`

test/e2e/sidecar_namespace_test.go
```diff
@@ -0,0 +1,106 @@
...
+	j := getJaegerAgentAsSidecarDefinition(jaegerInstanceName, namespace)
+	err := fw.Client.Create(goctx.TODO(), j, cleanupOptions)
+	require.NoError(t, err, "Failed to create jaeger instance")
+	defer undeployJaegerInstance(j)
+
+	err = e2eutil.WaitForDeployment(t, fw.KubeClient, namespace, jaegerInstanceName, 1, retryInterval, timeout)
+	require.NoError(t, err, "Error waiting for Jaeger instance deployment")
+
+	dep := getVertxDefinition(map[string]string{})
+	err = fw.Client.Create(goctx.TODO(), dep, cleanupOptions)
```
jpkrohling: Could the test perhaps wait for the deployment to stabilize and check that no sidecar exists? This way, it asserts that the sidecar was injected because of the namespace annotation.

test/e2e/sidecar_test.go
```diff
@@ -168,6 +168,11 @@ func getJaegerAgentAsSidecarDefinition(name, namespace string) *v1.Jaeger {
 		},
 		Spec: v1.JaegerSpec{
 			Strategy: v1.DeploymentStrategyAllInOne,
+			JaegerCommonSpec: v1.JaegerCommonSpec{
+				// add inject annotation to jaeger deployment to non existing Jaeger instance
+				// we don't want to inject agent to jaeger deployment because the port collisions
+				Annotations: map[string]string{inject.Annotation: "doesNotExists"},
```
jpkrohling: Not sure I get it. Are you saying that if a Jaeger CR for all-in-one is created, it will get a sidecar injected if the namespace is annotated, unless the deployment is annotated with a non-existing Jaeger instance?

I think the operator should handle this special case: we can detect that the deployment is a Jaeger deployment, skip the deployment.

deploy/role_binding.yaml
pavolloffay: `ClusteRole` and `ClusterRoleBindings` are cluster-wide objects the namespace is ignored. This patch does not change the way how the operator is being installed.

pkg/inject/sidecar_test.go
pavolloffay: it's a caption. It's passed to `test.Run(cap, ..`. It helps to identify the failing test - if there is any.

Makefile
pavolloffay: No, I will actually merge the global-manifests target with the prepare e2e test. It's not being reused in other places so a single target is enough.

test/e2e/sidecar_test.go
pavolloffay: > Not sure I get it. Are you saying that if a Jaeger CR for all-in-one is created, it will get a sidecar injected if the namespace is annotated, unless the deployment is annotated with a non-existing Jaeger instance? 

Yes, it will get sidecar injected.

pavolloffay: I will create an issue for it

pavolloffay: https://github.com/jaegertracing/jaeger-operator/issues/900

test/e2e/utils.go
```diff
@@ -102,31 +102,29 @@ func prepare(t *testing.T) (*framework.TestCtx, error) {
...
 	}
 
 	crb := &rbac.ClusterRoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: namespace + "-jaeger-operator-cluster-admin",
+			Name: namespace + "jaeger-operator",
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					Name:       ns.Name,
+					Kind:       "Namespace",
```
pavolloffay: `ns.Kind` and `ns.APIVersion` were always empty. This seems to work fine.

pavolloffay:
@jpkrohling I have applied requested changes. It should be ready from my side.

test/e2e/sidecar_test.go
jpkrohling: Would it be possible to fix it in this PR? This currently isn't a problem, as the Jaeger deployment has to be explicitly annotated to reach this situation, whereas current deployment will fail once this PR is merged and the namespace gets annotated. It should be relatively simple, as the `Needed` could just check for label `app: jaeger`, or `app.kubernetes.io/part-of: jaeger`.

Makefile
```diff
@@ -82,18 +82,18 @@ unit-tests:
 e2e-tests: prepare-e2e-tests e2e-tests-smoke e2e-tests-cassandra e2e-tests-es e2e-tests-self-provisioned-es e2e-tests-streaming e2e-tests-examples1 e2e-tests-examples2
 
 .PHONY: prepare-e2e-tests
-prepare-e2e-tests: crd build docker push
```
jpkrohling: Why has this been removed? The `crd` target is idempotent and it does make sense to be part of the preparation for the e2e tests.

pkg/inject/sidecar_test.go
jpkrohling: Got it. I was confused by the name, even though I got the purpose.

Makefile
pavolloffay: CRD is added to `global-manifests` there is no need to have it in two places.

test/e2e/sidecar_test.go
pavolloffay: I would prefer a separate PR this is already a medium size.

jpkrohling: Alright, let's just keep in mind that we probably don't want to release 1.17 with this bug.

PR #901, opened 2020-02-11
Handle test namespace creation and deletion as operator-sdk names are…

kevinearls:
… too long.

The update to operator-sdk 0.15.1 included a change where namespace names became much longer.  https://github.com/operator-framework/operator-sdk/blob/master/pkg/test/context.go#L53  This resulted in a number of tests failing on OpenShift because of route names that were too long and other issues.

With this change tests run on OpenShift will manage the creation and deletion of namespaces instead of using the one created by the operator sdk.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,10 +23,12 @@ PROMETHEUS_OPERATOR_TAG ?= v0.34.0
 ES_OPERATOR_IMAGE ?= quay.io/openshift/origin-elasticsearch-operator:4.3
 SDK_VERSION=v0.15.1
 GOPATH ?= "$(HOME)/go"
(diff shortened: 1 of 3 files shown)


jpkrohling:
Question: what exactly is failing? If it's the route that we talked before, then the bug is actually in the operator, as it shouldn't create a route with an invalid name.

By the way, this is going to conflict with #897.

kevinearls:
@jpkrohling In one case it was definitely the route name.  The others were not so obvious, so I was hoping to use this as a way to fix tests, and then investigate if necessary.

I can investigate the other failures before proceeding.

kevinearls:
This is no longer needed as the problem was fixed with #904 and #915

PR #902, opened 2020-02-12
No modify annotation when injecting

### LINKED ISSUE #899: Inject sidecar agent with true breaks if Jaeger name changes
Jaeger operator uses annotation `sidecar.jaegertracing.io/inject: true`  to inject an agent to a deployment. After the first loop (if the Jaeger instance is found) it changes `true` to `<jaeger-instance-name>`. Therefore if the jaeger instance name is changed the sidecars will not be injected.

rubenvp8510:

Resolves #899 

In the actual code, when Select is called to select a jaeger instance, it modifies the annotation and set the value to the jaeger instance name.

I think this can lead to the scenario when you rename (or recreate) the Jaeger instance and there is no way to map the old name instance to the new one. I would prefer to preserve the value.

### DIFF
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -243,7 +243,10 @@ func (b *Background) cleanDeployments(ctx context.Context) {
 
 	// check deployments to see which one needs to be cleaned.
 	for _, dep := range deployments.Items {
-		if instanceName, ok := dep.Annotations[inject.Annotation]; ok {
+		if dep.Labels == nil {
+			continue
+		}
+		if instanceName, ok := dep.Labels[inject.Label]; ok {
 			_, instanceExists := instancesMap[instanceName]
 			if !instanceExists { // Jaeger instance not exist anymore, we need to clean this up.
 				inject.CleanSidecar(&dep)
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -243,9 +243,6 @@ func (b *Background) cleanDeployments(ctx context.Context) {
 
 	// check deployments to see which one needs to be cleaned.
 	for _, dep := range deployments.Items {
-		if dep.Labels == nil {
-			continue
-		}
 		if instanceName, ok := dep.Labels[inject.Label]; ok {
 			_, instanceExists := instancesMap[instanceName]
 			if !instanceExists { // Jaeger instance not exist anymore, we need to clean this up.
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -41,20 +41,14 @@ const (
 func Sidecar(jaeger *v1.Jaeger, dep *appsv1.Deployment) *appsv1.Deployment {
 	deployment.NewAgent(jaeger) // we need some initialization from that, but we don't actually need the agent's instance here
 	logFields := jaeger.Logger().WithField("deployment", dep.Name)
-	if jaeger == nil || (dep.Annotations[Annotation] != jaeger.Name && dep.Annotations[AnnotationLegacy] != jaeger.Name) {
+	labelValue, hasLabel := dep.Labels[Label]
+	if jaeger == nil || (hasLabel && labelValue != jaeger.Name) {
 		logFields.Trace("skipping sidecar injection")
 	} else {
 		decorate(dep)
-
 		logFields.Debug("injecting sidecar")
 		dep.Spec.Template.Spec.Containers = append(dep.Spec.Template.Spec.Containers, container(jaeger, dep))
-
-		jaegerName := util.Truncate(jaeger.Name, 63)
-		if dep.Labels == nil {
-			dep.Labels = map[string]string{Label: jaegerName}
-		} else {
-			dep.Labels[Label] = jaegerName
-		}
+		setInjectedInstanceLabel(dep, jaeger)
 	}
 
 	return dep
@@ -77,6 +71,15 @@ func Needed(dep *appsv1.Deployment, ns *corev1.Namespace) bool {
 		return false
 	}
 	return !HasJaegerAgent(dep)
+}
+
+func setInjectedInstanceLabel(target *appsv1.Deployment, jaeger *v1.Jaeger) {
+	jaegerName := util.Truncate(jaeger.Name, 63)
+	if target.Labels == nil {
+		target.Labels = map[string]string{Label: jaegerName}
+	} else {
+		target.Labels[Label] = jaegerName
+	}
 }
 
 // Select a suitable Jaeger from the JaegerList for the given Pod, or nil of none is suitable
@@ -102,10 +105,6 @@ func Select(target *appsv1.Deployment, ns *corev1.Namespace, availableJaegerPods
 		// jaeger instance to use!
 		// first, we make sure we normalize the name:
 		jaeger := &availableJaegerPods.Items[0]
-		if target.Annotations == nil {
-			target.Annotations = map[string]string{}
-		}
-		target.Annotations[Annotation] = jaeger.Name
 		return jaeger
 	}
 	return nil
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -41,7 +41,6 @@ const (
 func Sidecar(jaeger *v1.Jaeger, dep *appsv1.Deployment) *appsv1.Deployment {
 	deployment.NewAgent(jaeger) // we need some initialization from that, but we don't actually need the agent's instance here
 	logFields := jaeger.Logger().WithField("deployment", dep.Name)
-
 	if jaeger == nil || (dep.Annotations[Annotation] != jaeger.Name && dep.Annotations[AnnotationLegacy] != jaeger.Name) {
 		logFields.Trace("skipping sidecar injection")
 	} else {
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -41,14 +41,26 @@ const (
 func Sidecar(jaeger *v1.Jaeger, dep *appsv1.Deployment) *appsv1.Deployment {
 	deployment.NewAgent(jaeger) // we need some initialization from that, but we don't actually need the agent's instance here
 	logFields := jaeger.Logger().WithField("deployment", dep.Name)
-	labelValue, hasLabel := dep.Labels[Label]
-	if jaeger == nil || (hasLabel && labelValue != jaeger.Name) {
-		logFields.Trace("skipping sidecar injection")
+
+	if jaeger == nil {
+		logFields.Trace("no Jaeger instance found, skipping sidecar injection")
+		return dep
+	}
+
+	if val, ok := dep.Labels[Label]; ok && val != jaeger.Name {
+		logFields.Trace("deployment is assigned to a different Jaeger instance, skipping sidecar injection")
+		return dep
+	}
+
+	decorate(dep)
+	logFields.Debug("injecting sidecar")
+	dep.Spec.Template.Spec.Containers = append(dep.Spec.Template.Spec.Containers, container(jaeger, dep))
+	jaegerName := util.Truncate(jaeger.Name, 63)
+
+	if dep.Labels == nil {
+		dep.Labels = map[string]string{Label: jaegerName}
 	} else {
-		decorate(dep)
-		logFields.Debug("injecting sidecar")
-		dep.Spec.Template.Spec.Containers = append(dep.Spec.Template.Spec.Containers, container(jaeger, dep))
-		setInjectedInstanceLabel(dep, jaeger)
+		dep.Labels[Label] = jaegerName
 	}
 
 	return dep
@@ -71,15 +83,6 @@ func Needed(dep *appsv1.Deployment, ns *corev1.Namespace) bool {
 		return false
 	}
 	return !HasJaegerAgent(dep)
-}
-
-func setInjectedInstanceLabel(target *appsv1.Deployment, jaeger *v1.Jaeger) {
-	jaegerName := util.Truncate(jaeger.Name, 63)
-	if target.Labels == nil {
-		target.Labels = map[string]string{Label: jaegerName}
-	} else {
-		target.Labels[Label] = jaegerName
-	}
 }
 
 // Select a suitable Jaeger from the JaegerList for the given Pod, or nil of none is suitable
--- a/pkg/inject/sidecar_test.go
+++ b/pkg/inject/sidecar_test.go
@@ -32,17 +32,9 @@ func reset() {
 
 func TestInjectSidecar(t *testing.T) {
 	jaeger := v1.NewJaeger(types.NamespacedName{Name: "TestInjectSidecar"})
-	dep := dep(map[string]string{Annotation: jaeger.Name}, map[string]string{})
+	dep := dep(map[string]string{}, map[string]string{})
 	dep = Sidecar(jaeger, dep)
-	assert.Len(t, dep.Spec.Template.Spec.Containers, 2)
-	assert.Contains(t, dep.Spec.Template.Spec.Containers[1].Image, "jaeger-agent")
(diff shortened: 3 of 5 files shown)


pavolloffay:
Let's consider also a scenario when:
1. deploy a single Jaeger instance in cluster
2. add app deployment with inject annotation = true
3. deploy another Jaeger instance.

In this case we should not break existing injected agents.

rubenvp8510:
@pavolloffay good point, I'll check that case.

But, if you add a four step:
4. delete the first instance

Then the deployment with inject annotation = true, should be injected with the new side-car, right? 
Which essentially is the problem this PR is addressing.

pavolloffay:
A good point. @jpkrohling what do you think? It really depends on how we define the behavior.

jpkrohling:
I used to think that `true` means `select a suitable Jaeger instance for me` restricted for the first time it found a suitable instance, but I think that we could do better. In the case that @pavolloffay mentioned, the deployment should still be using the first Jaeger instance. And once the first instance is deleted, the deployment should automatically be switched over to the new instance.

We might have all the bits necessary for this already: we have the annotation `sidecar.jaegertracing.io/inject` and we have a label `sidecar.jaegertracing.io/injected`. So, we could just leave the inject as the user has defined it originally (true, false, instance-name), and update the label to state what's the instance currently in use. In the step 3 from @pavolloffay's, this value can serve to disambiguate.

rubenvp8510:
@jpkrohling  yes that is what I was thinking, using the label  for disambiguate, I think this is the best approach as we avoid touching values defined by the user, (the label is completely handled by the operator).

jpkrohling:
@pavolloffay do you agree with the expected behavior? If so, @rubenvp8510 could you then adapt this PR to what we discussed above, adding tests to confirm that this new behavior is in place?

pavolloffay:
:+1:  It sounds good to me

pkg/inject/sidecar.go
```diff
@@ -41,8 +41,13 @@ const (
...
 	logFields := jaeger.Logger().WithField("deployment", dep.Name)
-
-	if jaeger == nil || (dep.Annotations[Annotation] != jaeger.Name && dep.Annotations[AnnotationLegacy] != jaeger.Name) {
+	// Three conditions for inject:
+	// 	- jaeger not nil
+	// 	- jaeger instance name equal to annotation or annotation true (including legacy annotation).
+	//  - was not already injected (does not contain label)
+	labelValue, hasLabel := dep.Labels[Label]
+	hasInvalidAnnotation := dep.Annotations[Annotation] != jaeger.Name && dep.Annotations[Annotation] != "true" && dep.Annotations[AnnotationLegacy] != jaeger.Name
+	if jaeger == nil || hasInvalidAnnotation || (hasLabel && labelValue != jaeger.Name) {
```
pavolloffay: `(hasLabel && labelValue != jaeger.Name)` perhaps it can be `labelValue != jaeger.Name`

pkg/controller/namespace/namespace_controller.go
```diff
@@ -111,7 +111,9 @@ func (r *ReconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Res
 				return reconcile.Result{}, err
 			}
 
+			// Add inject annotation
 			jaeger := inject.Select(dep, ns, jaegers)
+			dep.Annotations[inject.Annotation] = jaeger.Name
```
pavolloffay: Is this intentional? I thought we are not going to change the annotation value and instead use label to store jaeger instance name.

pkg/inject/sidecar.go
rubenvp8510: not the same, the first time this function is called (and the deployment does not have an injected side-car), the label does not exist, in successive calls it should exist, but is only re-inject the side-car if is the same Jaeger instance. May be the second part is not entirely necessary.

pkg/controller/namespace/namespace_controller.go
rubenvp8510: This was a mistake, this tried to address the case when the deployment doesn't have any  annotation, but the namespace have one. In that case the` inject.sidecar` function won't inject anything due the modification I did to `inject.select`  for not to add the annotation. Now reading your comment I realized that this is wrong way to address the things and I changed.

rubenvp8510: I think `Inject.Select` function should not have any side effect nor modify any resource (deployment or jaeger instance), as it's only concern is to select the appropriate Jaeger instance to inject. 

On the other hand, the `Inject.Sidecar `should do all the rest of the work.

rubenvp8510:
this is ready for another review.

pkg/controller/namespace/namespace_controller.go
pavolloffay: +1 on not having any effects on the deployment objects

pkg/autodetect/main.go
```diff
@@ -243,7 +243,10 @@ func (b *Background) cleanDeployments(ctx context.Context) {
 
 	// check deployments to see which one needs to be cleaned.
 	for _, dep := range deployments.Items {
-		if instanceName, ok := dep.Annotations[inject.Annotation]; ok {
+		if dep.Labels == nil {
```
pavolloffay: why the check? Was it causing NPE somewhere?

test/e2e/sidecar_test.go
```diff
@@ -47,7 +48,7 @@ func (suite *SidecarTestSuite) SetupSuite() {
 }
 
 func (suite *SidecarTestSuite) TearDownSuite() {
-	handleSuiteTearDown()
+	// handleSuiteTearDown()
```
pavolloffay: foget to remove?

test/e2e/sidecar_test.go
```diff
@@ -64,13 +65,20 @@ func (suite *SidecarTestSuite) AfterTest(suiteName, testName string) {
...
 func (suite *SidecarTestSuite) TestSidecar() {
+
+	agentAsSideCarDeleted := false
+
 	cleanupOptions := &framework.CleanupOptions{TestContext: ctx, Timeout: timeout, RetryInterval: retryInterval}
 
 	jaegerInstanceName := "agent-as-sidecar"
 	j := getJaegerAgentAsSidecarDefinition(jaegerInstanceName, namespace)
+	defer func() {
+		if !agentAsSideCarDeleted {
```
pavolloffay: Why is this check needed? The variable is intialized and set in the middle of the tests. Isn't the deterministic at the end of the test?

pkg/inject/sidecar.go
```diff
@@ -41,21 +41,14 @@ const (
 func Sidecar(jaeger *v1.Jaeger, dep *appsv1.Deployment) *appsv1.Deployment {
 	deployment.NewAgent(jaeger) // we need some initialization from that, but we don't actually need the agent's instance here
 	logFields := jaeger.Logger().WithField("deployment", dep.Name)
-
-	if jaeger == nil || (dep.Annotations[Annotation] != jaeger.Name && dep.Annotations[AnnotationLegacy] != jaeger.Name) {
+	labelValue, hasLabel := dep.Labels[Label]
+	if jaeger == nil || (hasLabel && labelValue != jaeger.Name) {
```
pavolloffay: Can be the `hasLabel` omitted? e.g.  `if jaeger == nil || (labelValue != jaeger.Name)`

test/e2e/sidecar_test.go
rubenvp8510: I extended this test to test the case addressed by this PR.

At some point in this test I create a new instance and delete the "agent-as-sidecar" Jaeger instance. If I don't do this check the test fails because it tries to delete an non-exist instance at the end.

test/e2e/sidecar_test.go
rubenvp8510: Yes :flushed:

pkg/autodetect/main.go
rubenvp8510: At some point it throws me a NPE on the tests, but it seems like this is not necessary in the latest changes.

pkg/inject/sidecar.go
rubenvp8510: I don't think it can be omitted. 

The logic is this: if deployment doesn't have a label, that means that it needs to be injected.  If it has a label, we need to check if the label is equal to jaeger instance. The latest only applies in the case of an update, while the former is for new deployments. Remember, now `inject.Select` function doesn't have side effects, nor modify the deployment.

If we don't check for `hasLabel` some sidecar tests will fail. For instance, `TestInjectSidecar` will fail because it will try to inject a sidecar in a deployment that doesn't have any label (the label is added when the sidecar is injected). If we check only `labelValue != jaeger.Name` the deployment will never be injected. The first time, when it doesn't have a label, `labelValue` will be an empty string.

On the other hand, if we check for `hasLabel` we will inject the side car the first time.  Of course I can do a `labelValue != "",` but I would prefer to use the value returned by go maps for key existence.


Does that makes sense?

rubenvp8510:
@jpkrohling please review :)

pkg/inject/sidecar.go
jpkrohling: I'm not a huge fan of this conditional. How about something like this?

```go
	if jaeger == nil {
		logFields.Trace("no Jaeger instance found, skipping sidecar injection")
		return dep
	}

	if val, ok := dep.Labels[Label]; ok && val != jaeger.Name {
		logFields.Trace("deployment is assigned to a different Jager instance, skipping sidecar injection")
		return dep
	}

	decorate(dep)
	dep.Spec.Template.Spec.Containers = append(dep.Spec.Template.Spec.Containers, container(jaeger, dep))
	setInjectedInstanceLabel(dep, jaeger)
```

pkg/inject/sidecar.go
```diff
@@ -80,6 +73,15 @@ func Needed(dep *appsv1.Deployment, ns *corev1.Namespace) bool {
 	return !HasJaegerAgent(dep)
 }
 
+func setInjectedInstanceLabel(target *appsv1.Deployment, jaeger *v1.Jaeger) {
```
jpkrohling: In new code, we are trying to avoid changing the parameters we receive, returning new values instead. Like:

```go
func setInjectedInstanceLabel(target *appsv1.Deployment, jaeger *v1.Jaeger) *appsv1.Deployment {
  ret := *target
  ...
  return &ret
}
```

In this particular case, I would actually not split it in a separate function, as it this is very simple and, apparently, needed only in the `Sidecar` function.

test/e2e/sidecar_test.go
jpkrohling: I also got confused by this, but I think I understand what's going on: if the test fails midway of the execution, this instance will be lingering around, which is why you need the conditional undeploy.

It might be better to change the `undeployJaegerInstance` to return a `bool` indicating whether it was able to find and remove the instance. In most cases, the caller wouldn't care about whether the Jaeger has been undeployed correctly, I suppose, but in the cases where it does care, it could just call `require.True(t, undeployJaegerInstance(j))`.

cc @kevinearls

pkg/inject/sidecar.go
jpkrohling: One more thing I wanted to check: this changes the behavior of the `Sidecar` function, so that it won't check the `Annotation` at all anymore. Therefore, this test would be outdated:

https://github.com/jaegertracing/jaeger-operator/blob/4424e8ac6f600a469a2ba9d05abbd895d43a29c4/pkg/inject/sidecar_test.go#L33-L40

Removing the `Annotation` would still get this test passing, which proves the change in behavior. If this is desired, then please change the existing tests, including the one above, but make sure that the contract with the consumers isn't broken: people annotating their deployments have to get the same behavior as before.

pkg/inject/sidecar_test.go
```diff
@@ -187,7 +187,7 @@ func TestSidecarDefaultPorts(t *testing.T) {
 func TestSkipInjectSidecar(t *testing.T) {
 	// prepare
 	jaeger := v1.NewJaeger(types.NamespacedName{Name: "TestSkipInjectSidecar"})
-	dep := dep(map[string]string{Annotation: "non-existing-operator"}, map[string]string{})
+	dep := dep(map[string]string{}, map[string]string{Label: "non-existing-operator"})
```
jpkrohling: Depending on the outcome of the previous comment around the test `TestInjectSidecar`, this test might now be ambiguous: is this testing that the sidecar is being skipped because there's no `Annotation` set, or is it because the instance isn't existing?

test/e2e/sidecar_test.go
kevinearls: @jpkrohling agreed, I think we can change that to return a bool and let the caller decide if they care about failure.

pkg/inject/sidecar.go
rubenvp8510: True, I need to update the above test, (and I need to review if I need to update more). The contract with the consumer isn't broken, the only difference is that now the annotation is only used for select the instance (in the select function) and not for inject. but If I'm not wrong or miss something, that is transparent for the consumer.

rubenvp8510: > I'm not a huge fan of this conditional. How about something like this?
> 
> ```go
> 	if jaeger == nil {
> 		logFields.Trace("no Jaeger instance found, skipping sidecar injection")
> 		return dep
> 	}
> 
> 	if val, ok := dep.Labels[Label]; ok && val != jaeger.Name {
> 		logFields.Trace("deployment is assigned to a different Jager instance, skipping sidecar injection")
> 		return dep
> 	}
> 
> 	decorate(dep)
> 	logFields.Debug("injecting sidecar")
> 	dep.Spec.Template.Spec.Containers = append(dep.Spec.Template.Spec.Containers, container(jaeger, dep))
> 	setInjectedInstanceLabel(dep, jaeger)
> ```

Is essentially the same, but more separated, and with better log messages :)

pkg/inject/sidecar.go
rubenvp8510: I don't remember why I split this.. I think it was due some codeclimate issues. I'll check but seems to not be necessary

test/e2e/sidecar_test.go
rubenvp8510: +1

pkg/inject/sidecar.go
jpkrohling: I find it easier to read and understand, but if you disagree, feel free to keep the current version.

test/e2e/sidecar_test.go
rubenvp8510: @jpkrohling @kevinearls @pavolloffay I'll do the suggested change but Is it ok to maintain this in a single test? or better to split?

pkg/inject/sidecar.go
rubenvp8510: I indeed think your suggestion is a good idea, improves a lot the readability.

rubenvp8510:
@jpkrohling 

I updated the sidecar tests, but as you mentioned on one of your comments, the annotation is irrelevant for those tests (except the `inject.Select` test), and this is intentional.

This is because  this PR is not validating for annotation on `inject.Sidecar`, now  it only  injects the  Jaeger instance passed as an argument without checking for annotations,  of course, it does another checks, like if the instance not `nil` and if the deploy doesn't have one sidecar already. We can assume a previously selected jaeger instance  by` inject.Select` , which is still using the annotator for the selection. Does that makes sense?


 Summary:  This PR uses the annotations only for select the correct Jaeger instance. and use label for check if the deployment already have an agent and which Jaeger instance is the owner.

From the consumer/user perspective it will be the same, the deployments with annotations will be injected.

test/e2e/sidecar_test.go
kevinearls: @rubenvp8510 I think it would be better to split.

kevinearls:
@jpkrohling @rubenvp8510 I'd be ok splitting the tests in a separate PR.

jpkrohling:
I'm merging this, but @rubenvp8510, please create an issue to track the splitting of the test.

PR #903, opened 2020-02-12
Disable agent injection to jaeger instances and when false value is used

### LINKED ISSUE #900: Jaeger agent can be injected into jaeger instance
Jaeger agent can be injected into Jaeger agent or all-in-one deployment which causes deployment failures due to port collisions.

This is not a problem when the inject annotation (`sidecar.jaegertracing.io/inject=true`) is added on deployments, but it can be problematic when the inject annotation is applied on a namespace (https://github.com/jaegertracing/jaeger-operator/pull/897) and the namespace contains Jaeger instance.

pavolloffay:
Resolves #900 

Disables agent injection on:
* deployments with label `app=jaeger`
* deployment with the inject annotation value `false`

### DIFF
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -72,6 +72,11 @@ func Needed(dep *appsv1.Deployment, ns *corev1.Namespace) bool {
 		}).Debug("annotation not present, not injecting")
 		return false
 	}
+	// do not inject jaeger due to port collision
+	// do not inject if deployment's Annotation value is false
+	if dep.Labels["app"] == "jaeger" || dep.Annotations[Annotation] == "false" {
+		return false
+	}
 	return !HasJaegerAgent(dep)
 }
 
@@ -98,6 +103,9 @@ func Select(target *appsv1.Deployment, ns *corev1.Namespace, availableJaegerPods
 		// jaeger instance to use!
 		// first, we make sure we normalize the name:
 		jaeger := &availableJaegerPods.Items[0]
+		if target.Annotations == nil {
+			target.Annotations = map[string]string{}
+		}
 		target.Annotations[Annotation] = jaeger.Name
 		return jaeger
 	}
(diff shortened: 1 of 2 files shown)


pkg/inject/sidecar.go
```diff
@@ -98,6 +103,9 @@ func Select(target *appsv1.Deployment, ns *corev1.Namespace, availableJaegerPods
 		// jaeger instance to use!
 		// first, we make sure we normalize the name:
 		jaeger := &availableJaegerPods.Items[0]
+		if target.Annotations == nil {
```
pavolloffay: not related to this PR but I got a couple of times `"assignment to entry in nil map" (assignment to entry in nil map)` when testing this feature.

pkg/inject/sidecar.go
```diff
@@ -72,6 +72,11 @@ func Needed(dep *appsv1.Deployment, ns *corev1.Namespace) bool {
 		}).Debug("annotation not present, not injecting")
 		return false
 	}
+	// do not inject jaeger due to port collision
+	// do not inject if deployment's Annotation value is false
+	if dep.Labels["app"] == "jaeger" || dep.Annotations[Annotation] == "false" {
```
jpkrohling: Actually, I think `false` is already covered. If this value is set to anything other than `false` or an existing instance name, it will just skip the injection, as there won't be any suitable Jaeger instance:

https://github.com/jaegertracing/jaeger-operator/blob/16592e02c4089738e2156ce20349cb52037348cd/pkg/inject/sidecar_test.go#L186-L197

pavolloffay: The referenced test is correct. It does not inject the sidecar but there will be a log saying that Jaeger instance wasn't resolved for injection. 

We can pursue this direction, it seems more like a workaround solution rather than something what is supported.

pavolloffay: https://github.com/jaegertracing/jaeger-operator/issues/910

PR #908, opened 2020-02-14
Reduced logging verbosity when debug is enabled

jpkrohling:
While debugging a problem with #904, I noticed that the logs are too verbose to be useful. This PR reduces it drastically, by:

* Changing the "waiting for..." messages to be shown only once
* Setting the levels for background process to "trace" for messages that are repeated
* Fixing the role.yaml to include the ability to 'watch' namespaces

Unrelated change included in this PR: the reconciliation for namespaces and deployments had three different contexts each. This PR creates one context and reuses it across the reconciliation. It should make it easier when we decide to instrument those routines.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ prepare-e2e-tests: build docker push
 
 	@cat test/operator.yaml | sed "s~image: jaegertracing\/jaeger-operator\:.*~image: $(BUILD_IMAGE)~gi" >> deploy/test/namespace-manifests.yaml
 
-	# ClusterRoleBinding is created in test codebase because we don't know service account namespace
+	@# ClusterRoleBinding is created in test codebase because we don't know service account namespace
 	@cp deploy/role.yaml deploy/test/global-manifests.yaml
 	@echo "---" >> deploy/test/global-manifests.yaml
 
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -25,6 +25,7 @@ rules:
   verbs:
   - 'get'
   - 'list'
+  - 'watch'
 - apiGroups:
   - apps
   resources:
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -3,6 +3,7 @@ package autodetect
 import (
 	"context"
 	"strings"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -27,6 +28,7 @@ type Background struct {
 	dcl      discovery.DiscoveryInterface
 	ticker   *time.Ticker
 
+	firstRun         *sync.Once
 	retryDetectKafka bool
 	retryDetectEs    bool
 }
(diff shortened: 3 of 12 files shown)


jpkrohling:
@kevinearls, I think you'll appreciate this change.

Makefile
```diff
@@ -89,7 +89,7 @@ prepare-e2e-tests: build docker push
 
 	@cat test/operator.yaml | sed "s~image: jaegertracing\/jaeger-operator\:.*~image: $(BUILD_IMAGE)~gi" >> deploy/test/namespace-manifests.yaml
 
-	# ClusterRoleBinding is created in test codebase because we don't know service account namespace
+	@# ClusterRoleBinding is created in test codebase because we don't know service account namespace
```
pavolloffay: Is `@` needed? It's just a comment

deploy/role.yaml
```diff
@@ -25,6 +25,7 @@ rules:
   verbs:
   - 'get'
   - 'list'
+  - 'watch'
```
pavolloffay: good catch! It is related to the log entry we spoke about yesterday?

pkg/controller/deployment/deployment_controller.go
```diff
@@ -65,14 +65,16 @@ type ReconcileDeployment struct {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	ctx := context.Background()
```
pavolloffay: Isn't there a context in the request parameter?

Makefile
jpkrohling: Otherwise it is seen in the console logs:

```
Writing manifest to image destination
Storing signatures
# ClusterRoleBinding is created in test codebase because we don't know service account namespace
Creating namespace default
```

deploy/role.yaml
jpkrohling: Yes

pkg/controller/deployment/deployment_controller.go
jpkrohling: Unfortunately, no.

kevinearls:
@jpkrohling :thumbsup:

pkg/autodetect/main.go
```diff
@@ -81,12 +90,16 @@ func (b *Background) Stop() {
...
+		log.WithError(err).Info("failed to determine the platform capabilities, auto-detected properties will fallback to their default values.")
 		viper.Set("platform", v1.FlagPlatformKubernetes)
 		viper.Set("es-provision", v1.FlagProvisionElasticsearchNo)
 	} else {
-		// we could run all the detect* functions in parallel, but let's keep it simple for now
-		b.detectPlatform(apiList)
+
+		b.firstRun.Do(func() {
+			// the platform won't change during the execution of the operator, need to run it only once
+			b.detectPlatform(apiList)
```
pavolloffay: Does this affect the detection of Kafka and ES operators? 

Kafka or ES operators can be installed after Jaeger operator. We even removed OLM dependency on ES operator so we can expect people installing ES operator after Jaeger operator.

jpkrohling: No, the only part within the `firstRun.Do()` is the detect platform, which detects whether the operator is running in Kubernetes or OpenShift.

Kafka and ES are detected two lines below:

```golang
		b.detectElasticsearch(apiList)
		b.detectKafka(apiList)
```

PR #915, opened 2020-02-19
Fixes required after update to operator-sdk 0.15.1

kevinearls:
Signed-off-by: Kevin Earls <kearls@redhat.com>

This is a followup to #904  as a few changes were still required on the testing side.

PR #917, opened 2020-02-19
Fix route name for OCP

pavolloffay:
Resolves https://issues.redhat.com/browse/TRACING-965

Fixes smoke test failures on OCP `make e2e-tests-smoke` caused probably by https://github.com/jaegertracing/jaeger-operator/pull/891. Other related fixes https://github.com/jaegertracing/jaeger-operator/pull/904 and https://github.com/jaegertracing/jaeger-operator/pull/915

### DIFF
--- a/pkg/strategy/all_in_one_test.go
+++ b/pkg/strategy/all_in_one_test.go
@@ -12,6 +12,7 @@ import (
 
 	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
 	"github.com/jaegertracing/jaeger-operator/pkg/storage"
+	"github.com/jaegertracing/jaeger-operator/pkg/util"
 )
 
 func init() {
@@ -110,7 +111,7 @@ func assertDeploymentsAndServicesForAllInOne(t *testing.T, name string, s S, has
 	ingresses := map[string]bool{}
 	routes := map[string]bool{}
 	if viper.GetString("platform") == v1.FlagPlatformOpenShift {
-		routes[fmt.Sprintf("%s", name)] = false
+		routes[fmt.Sprintf("%s", util.DNSName(name))] = false
 	} else {
 		ingresses[fmt.Sprintf("%s-query", name)] = false
 	}
--- a/pkg/strategy/production_test.go
+++ b/pkg/strategy/production_test.go
@@ -14,6 +14,7 @@ import (
 
 	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
 	"github.com/jaegertracing/jaeger-operator/pkg/storage"
+	"github.com/jaegertracing/jaeger-operator/pkg/util"
 )
 
 func init() {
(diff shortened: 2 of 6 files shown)


pavolloffay:
This is the failing log 

```
Get https://my-jaeger-with-ingress-osdk-e2e-22fc72a5-db25-4707-876c-173229d49b42.apps.02-18-1582017833.devcluster.openshift.com/api/services: dial tcp: lookup my-jaeger-with-ingress-osdk-e2e-22fc72a5-db25-4707-876c-173229d49b42.apps.02-18-1582017833.devcluster.openshift.com: no such host
```

The instance name is the shortest part of the URL. Truncate should be called on the final URL but we do not have access to it.

This PR fixes the issue though.

jpkrohling:
> The instance name is the shortest part of the URL. Truncate should be called on the final URL but we do not have access to it.

The hostname is made of name and namespace of the route, and we can control the size of the name. This is what the actual code does to prevent creating an invalid route:
https://github.com/jaegertracing/jaeger-operator/blob/6b81f9d23b38396ba4401cf9f10a745672164da0/pkg/route/query.go#L37-L45

rubenvp8510:
I had the same problem as I mentioned in https://github.com/jaegertracing/jaeger-operator/pull/904 

This PR seems to fix the issue.

pavolloffay:
To @jpkrohling note, the test code correctly gets the route and uses the hostname from the route to construct the URL for smoke test.

The problem might by that shortened URL is ending with `-` which is causing the failure.

jpkrohling:
> The problem might by that shortened URL is ending with - which is causing the failure.

There's a `util.DNSName()` that could be applied to the name after the `util.Truncate()` operation.

pkg/util/dns_name.go
```diff
@@ -9,28 +9,22 @@ import (
 var regex = regexp.MustCompile(`[a-z0-9]`)
 
 // DNSName returns a dns-safe string for the given name.
-// Any char that is not [a-z0-9] is replaced by "-".
-// If the final name starts with "-", "a" is added as prefix. Similarly, if it ends with "-", "z" is added.
+// Any char that is not [a-z0-9] is replaced by "-" or "a".
+// Replacement character "a" is used only at the beginning or at the end of the name.
+// The function does not change length of the string.
 func DNSName(name string) string {
```
pavolloffay: I have changed the behavior, the issue was that the function was changing the length of the `name` and in our codebase the `DNSName` was applied after the truncation.

There is no need to use `a` and `z` as replacement chars. The original name cannot be reconstructed.

pkg/strategy/all_in_one_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/jaegertracing/jaeger-operator/pkg/util"
```
jpkrohling: Import is out of order :/

pkg/strategy/production_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/jaegertracing/jaeger-operator/pkg/util"
```
jpkrohling: This one too

pkg/strategy/all_in_one_test.go
pavolloffay: `goimports` seems to be buggy. I will fix it manually.

PR #922, opened 2020-02-24
Sidecar injection: Include the admin-http port

jrockway:
Judging from the code, at some point the jaeger-agent accepted metrics scrapes on the other HTTP port (config-rest), but that is no longer the case.  So this change exposes the admin-http port, which does accept metric scrapes.

(It appears that the 1.12.0 release changed this port, see https://github.com/jaegertracing/jaeger/pull/1442 )

I am actually confused as to why any of the other ports are exposed; the idea of the sidecar is that these ports only show up to other containers in the pod on 127.0.0.1 and it would be somewhat harmful if other applications started sending traces to the pod address.  But they're there, so I didn't touch them.

### DIFF
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -25,7 +25,7 @@ var (
 	// PrometheusDefaultAnnotations is a map containing annotations for prometheus to be inserted at sidecar in case it doesn't have any
 	PrometheusDefaultAnnotations = map[string]string{
 		"prometheus.io/scrape": "true",
-		"prometheus.io/port":   "5778",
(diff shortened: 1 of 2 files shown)


jpkrohling:
Looks like GitHub CI is having some problems at the moment. I tried to re-run the jobs, but it kept failing. Would you be able to run a `git commit --amend` and update this PR? Hopefully this would bypass GitHub's issues...

jrockway:
I did that, let's see how it goes!

jpkrohling:
Thanks for your PR!

PR #940, opened 2020-03-02
Use non-cached CR on reconciliation

jpkrohling:
When receiving a reconciliation request, we try to obtain the CR that triggered the event from Kubernetes repository. When using the regular client, we might end up retrieving a cached version that isn't up to date, meaning that we might see errors like the following from time to time:

```
ERRO[0350] failed to store back the current CustomResource  error="Operation cannot be fulfilled on jaegers.jaegertracing.io \"simple-prod\": the object has been modified; please apply your changes to the latest version and try again" execution="2020-03-02 13:55:31.292270472 +0000 UTC" instance=simple-prod namespace=observability
```

This error isn't a big problem, as it will trigger another reconciliation loop and things will eventually stabilize, but given that we might have a couple of reconciliation loops for the same object when it is first created, we want to bypass the cache and reduce the risk of facing this error.

### DIFF
--- a/pkg/controller/jaeger/jaeger_controller_test.go
+++ b/pkg/controller/jaeger/jaeger_controller_test.go
@@ -71,7 +71,7 @@ func TestDeletedInstance(t *testing.T) {
 	// no known objects
 	cl := fake.NewFakeClient()
 
-	r := &ReconcileJaeger{client: cl, scheme: s}
+	r := &ReconcileJaeger{client: cl, scheme: s, rClient: cl}
 
 	req := reconcile.Request{
 		NamespacedName: types.NamespacedName{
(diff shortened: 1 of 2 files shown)


jpkrohling:
@rubenvp8510, @annanay25, would one of you please review this PR:

pkg/controller/jaeger/jaeger_controller_test.go
```diff
@@ -71,7 +71,7 @@ func TestDeletedInstance(t *testing.T) {
 	// no known objects
 	cl := fake.NewFakeClient()
 
-	r := &ReconcileJaeger{client: cl, scheme: s}
+	r := &ReconcileJaeger{client: cl, scheme: s, rClient: cl}
```
annanay25: Should we use a different `fakeClient` as the `rClient` to ensure that it is invoking the non-cached reader?

jpkrohling: The `fakeClient` isn't cached, it holds everything in memory already.

annanay25: Okay right. I meant should we ensure its calling the `rClient` and not the `client`

jpkrohling: Great point. Yes, I think it's a good test to have.

PR #952, opened 2020-03-09
Remove unneeded dependency and increase timeout

### LINKED ISSUE #947: e2e-tests-examples2 fails for multiple reasons
This tests has been failing frequently for at least 2 reasons:

1. The jaeger-operator deployment at the beginning times out.  When I have run this manually and changed the timeout the operator eventually deploys, but sometimes takes well over 5 minutes.  This is sometimes seen in other test suites, but strangely occurs most frequently here, even though it's using the same code as the other test suites.

2.  The TestSimpleProdExample sometimes fails when the Jaeger instance doesn't fully deploy.  

This may be related to #945 although I think it is probably unrelated.

kevinearls:

This should fix #947 by removing an unneeded dependency in the e2e-tests-examples2 target.  It also adds increases some timeouts to avoid some other intermittent failures we've seen, and adds logging to help better diagnose failures.


kevinearls:
@jpkrohling  @objectiser @pavolloffay  @rubenvp8510  Please review

PR #962, opened 2020-03-12
Restore two minute timeout

kevinearls:

Five minutes is not needed for timeouts.  If two minutes is not adequate for something we should override the default for that test only.

### DIFF
--- a/test/e2e/utils.go
+++ b/test/e2e/utils.go
@@ -38,7 +38,7 @@ import (
 
 var (
 	retryInterval        = time.Second * 5
-	timeout              = time.Minute * 5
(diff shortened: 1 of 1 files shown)


test/e2e/utils.go
```diff
@@ -38,7 +38,7 @@ import (
 
 var (
 	retryInterval        = time.Second * 5
-	timeout              = time.Minute * 5
+	timeout              = time.Minute * 2
```
jpkrohling: I wonder if it would be helpful to use an env var, and fallback to a default of 2 minutes...

Issue #974, opened 2020-03-18
End to end test for examples1 does not pass on CI


pavolloffay:
PR showing failing CI https://github.com/jaegertracing/jaeger-operator/pull/973

@kevinearls would you like to have a look?

kevinearls:
@pavolloffay Sure, please assign this to me

pavolloffay:
Other failures can be seen here https://github.com/jaegertracing/jaeger-operator/actions?query=workflow%3A%22Kubernetes+end-to-end+tests%22

kevinearls:
@pavolloffay This should now be fixed with PR #995

PR #976, opened 2020-03-18
Only use 2 cassandra replica to avoid CI failures on minikube

kevinearls:

This is the first part of fixes for #974   I've seen a number of test failures where the three cassandra replicas either never come up, or they start but shortly thereafter go into an error state. In testing on my laptop changing to 2 replicas has eliminated that problem.

### DIFF
--- a/test/cassandra.yml
+++ b/test/cassandra.yml
@@ -36,7 +36,7 @@ items:
       matchLabels:
         app: cassandra
     serviceName: cassandra
-    replicas: 3
+    replicas: 1
     template:
       metadata:
         labels:
--- a/test/cassandra.yml
+++ b/test/cassandra.yml
@@ -36,7 +36,7 @@ items:
       matchLabels:
         app: cassandra
     serviceName: cassandra
-    replicas: 1
+    replicas: 2
     template:
       metadata:
         labels:
--- a/test/e2e/examples1_test.go
+++ b/test/e2e/examples1_test.go
@@ -103,6 +103,9 @@ func (suite *ExamplesTestSuite) TestWithCassandra() {
 }
 
 func (suite *ExamplesTestSuite) TestBusinessApp() {
+	if !isOpenShift(t) {
+		t.Skip("Skipping until issue #974 is fixed")
+	}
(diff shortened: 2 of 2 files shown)


pavolloffay:
Cassandra e2e failed

kevinearls:
@pavolloffay Please review.  I will continue working on TestBusinessApp, but so far am having a difficult time reproducing it outside of Github.

test/cassandra.yml
```diff
@@ -36,7 +36,7 @@ items:
       matchLabels:
         app: cassandra
     serviceName: cassandra
-    replicas: 3
+    replicas: 2
```
pavolloffay: I am not sure if 2 replicas is really a good idea. Can we switch rather to a single node cassandra cluster?

test/e2e/examples1_test.go
```diff
@@ -103,6 +103,9 @@ func (suite *ExamplesTestSuite) TestWithCassandra() {
 }
 
 func (suite *ExamplesTestSuite) TestBusinessApp() {
+	if !isOpenShift(t) {
+		t.Skip("Skipping until issue #974 is fixed")
```
pavolloffay: Add action item to #974 for removing this

test/cassandra.yml
kevinearls: My first commit to this PR was to change to a single replica, but that resulted in the cassandra_tests/TestCassandraSparkDependencies test failing and a message somewhere (I don't remember exactly, but can find it if necessary) about 2 cassandra instances being required for a quorum.  

Do I need to do something other than change the replica count to change to a single node cluster?

pavolloffay: The schema create job configuration will have to change probably.

e.g. mode test https://github.com/jaegertracing/jaeger/blob/b2a159a67b3d937eb29fe2a98f5b05708d34f56a/plugin/storage/cassandra/schema/create.sh#L36

Given that will require changing the example let's go with 2 replicas in tests.

kevinearls:
@jpkrohling @objectiser Please review or @pavolloffay please merge if you think this is ok

PR #978, opened 2020-03-18
Change 'make generate' to write only a single CRD

### LINKED ISSUE #977: CRD generator creates a list of CRDs instead of a single CRD
In previous versions, the Jaeger Operator used the operator-sdk to generate CRDs. As part of #932, we changed it to use the `controller-gen`, which writes a list of CRDs to the output, instead of a single CRD as it was before with the operator-sdk. Although the YAML file is still valid, it causes problems with other Operator SDK tooling, such as OLM.

jpkrohling:
Fixes #977 by using `tail` to chop off the first two lines of the CRD file.

### DIFF
--- a/.ci/generate.sh
+++ b/.ci/generate.sh
@@ -35,6 +35,15 @@ fi
 # move the generated CRD to the same location the operator-sdk places
 mv deploy/crds/jaegertracing.io_jaegers.yaml deploy/crds/jaegertracing.io_jaegers_crd.yaml
 
+# the controller-gen will generate a list of CRDs, but the operator-sdk tooling expects
+# a single item
+# the proper solutions are, in order:
+# 1) find a controller-gen switch that makes it write only one CRD. Such a switch doesn't exist yet: https://git.io/JvX5D
+# 2) use a YAML command line tool to get the first item from the file
+# 3) chop off the first two lines of the file
+# the last option is the easiest to implement for now, also because `tail` is found everywhere
+echo "$(tail -n +3 deploy/crds/jaegertracing.io_jaegers_crd.yaml)" > deploy/crds/jaegertracing.io_jaegers_crd.yaml
+
 # generate the schema validation (openapi) stubs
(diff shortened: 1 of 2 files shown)


.ci/generate.sh
```diff
@@ -35,6 +35,15 @@ fi
 # move the generated CRD to the same location the operator-sdk places
 mv deploy/crds/jaegertracing.io_jaegers.yaml deploy/crds/jaegertracing.io_jaegers_crd.yaml
 
+# the controller-gen will generate a list of CRDs, but the operator-sdk tooling expects
+# a single item
+# the proper solutions are, in order:
+# 1) find a controller-gen switch that makes it write only one CRD. Such a switch doesn't exist yet: https://git.io/JvX5D
```
pavolloffay: It might be worth opening an issue in the controller-gen repo to let them know that this is causing problems.

.ci/generate.sh
```diff
@@ -35,6 +35,15 @@ fi
...
+# 2) use a YAML command line tool to get the first item from the file
+# 3) chop off the first two lines of the file
+# the last option is the easiest to implement for now, also because `tail` is found everywhere
+echo "$(tail -n +3 deploy/crds/jaegertracing.io_jaegers_crd.yaml)" > deploy/crds/jaegertracing.io_jaegers_crd.yaml
```
pavolloffay: Actually maybe more robust would be to remove lines with `---`. Reason being they can remove empty line and this will chop off the first lines of the CRD

jpkrohling: Your suggestion gave me another idea: how about I positively check that the first line matches with what we expect it to be? This way, changes to the controller-gen will be caught right away.

pavolloffay: +1

jpkrohling: PR updated with this suggestion.

jpkrohling:
@pavolloffay I updated the PR after you reviewed. Would you mind taking a look again, even if only at the latest commit?

PR #983, opened 2020-03-23
Skip cassandra tests on openshift

kevinearls:

### DIFF
--- a/test/e2e/cassandra_test.go
+++ b/test/e2e/cassandra_test.go
@@ -22,7 +22,7 @@ type CassandraTestSuite struct {
 
 func (suite *CassandraTestSuite) SetupSuite() {
 	t = suite.T()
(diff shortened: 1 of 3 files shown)


jpkrohling:
What's the reason for this? Are we facing unsolvable problems for Cassandra?

kevinearls:
@jpkrohling I wouldn't say unsolvable, but intermittent problems, and we don't officially support Cassandra on OpenShift anyway.

jpkrohling:
I think we should try to ensure that it works there, even though it's not supported. If you need to temporarily disable it there, I'm fine with it, but disabling it altogether means that we cannot ever tell whether it works in OpenShift at all.

jpkrohling:
I'm merging with the assumption that this will be used only in special situations.

PR #987, opened 2020-03-25
Restored the displayName in the CSV

jpkrohling:
For some reason, the generated CSV for 1.17.1 is missing the `displayName`. When installing a bundle from the current master, the operator fails to install because of the missing field.

### DIFF
--- a/deploy/olm-catalog/jaeger-operator/1.17.1/jaeger-operator.v1.17.1.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/jaeger-operator/1.17.1/jaeger-operator.v1.17.1.clusterserviceversion.yaml
@@ -77,6 +77,7 @@ spec:
   customresourcedefinitions:
     owned:
     - description: Jaeger is the Schema for the jaegers API
+      displayName: Jaeger
       kind: Jaeger
       name: jaegers.jaegertracing.io
       version: v1


jpkrohling:
cc @jkandasa

deploy/olm-catalog/jaeger-operator/1.17.1/jaeger-operator.v1.17.1.clusterserviceversion.yaml
```diff
@@ -77,6 +77,7 @@ spec:
   customresourcedefinitions:
     owned:
     - description: Jaeger is the Schema for the jaegers API
+      displayName: Jaeger
```
pavolloffay: If this has to be done manually after every `generate` then it should be documented somewhere

jpkrohling: I don't know if there's an annotation missing somewhere, or if it's a bug in the CSV generator. Something to be investigated as part of another task.

pavolloffay: The issue is that we will probably forget to change this during the next release.

jpkrohling: Agree with you. I do suggest creating another task to investigate this, either to fix the problem or to document it in the release instructions.

PR #992, opened 2020-03-26
Retry after unauthorized errors

kevinearls:


kevinearls:
I'm no longer able to reproduce this, it may have been caused by a problem with the cluster I was testing on.

PR #1019, opened 2020-04-14
Change Ubuntu and Minikube versions to fix CI builds

kevinearls:

I have downgraded the CI builds to use Ubuntu version 16.04 instead of 18.04 because of https://github.com/kubernetes/kubernetes/issues/66067.  (Getting minikube to work on 18.04 will require some DNS configuration changes which I don't yet understand.)  

I have also taken this opportunity to update to the most recent version of minikube that I can get to run on Ubuntu 16.04, as we were previously using 1.3.1, which is quite old.

When possible we should got back to using ubuntu-latest and a newer version of minikube.  I have opened issue #1021 to track this.


kevinearls:
@objectiser  Please review.

PR #1032, opened 2020-04-21
Disable Elasticsearch feature to monitor available disk space 

kevinearls:

This is to avoid test failures when disk usage goes above 95% on the images we are using for tests.


kevinearls:

PR #1036, opened 2020-04-23
create Jaeger resource in the watched namespace

therealmitchconnors:
As is, the readme doesn't seem to work, and creates a jaeger whose status will never be updated...


README.md
```diff
@@ -33,6 +33,7 @@ apiVersion: jaegertracing.io/v1
 kind: Jaeger
 metadata:
   name: simplest
+  namespace: observability
```
pavolloffay, line 36: We should be consistent with other manifests and do not harcode the namespace.  `apply -n observability` should do the same.

jpkrohling:
@therealmitchconnors thanks for the PR! Looks like this is failing the [DCO check](https://github.com/jaegertracing/jaeger-operator/pull/1036/checks?check_run_id=612912526). As @pavolloffay mentioned, it might be better to use `-n observability` instead of specifying it in the YAML, for consistency.

PR #1046, opened 2020-04-29
Add CLI command jaeger-operator generate to generate manifests

### LINKED ISSUE #375: Operator to generate k8s deployment manifests
The operator could be used as a command line tool to generate k8s resource files based on supplied CR. So instead of posting objects to k8s API it would flush them to a file or standard output. 

The main motivation behind this is to deprecate https://github.com/jaegertracing/jaeger-kubernetes. During the release process we could generate and publish the template for target deployments - all-in-one. 

Example execution:
```
./jaeger-operator generate --cr cr.yml --output resources.yml
```

chlunde:
Defaults to stdin/stdout

Examples:

    jaeger-operator generate \
        --jaeger-all-in-one-image localimage \
        --cr ./deploy/examples/all-in-one-with-options.yaml \
        | kubectl apply -f -

    cat jaeger.yaml | podman run jaeger-operator generate > manifest.yaml

Fixes #375


chlunde:
POC for discussion, primarily because I wanted when looking at the code :)

pkg/strategy/strategy.go
```diff
@@ -17,7 +18,8 @@ import (
 
 // S knows what type of deployments to build based on a given spec
 type S struct {
-	typ                      v1.DeploymentStrategy
+	typ v1.DeploymentStrategy
+	// When adding a new type head, remember to update All() too
```
chlunde, line 22:
```suggestion
	// When adding a new type here, remember to update All() too
```

pavolloffay:
Could you please document this feature in readme how to use it as a docker command?

pkg/cmd/generate/main.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// TODO: jaeger-version -- now deprecated/nop. How to handle that. OK if running from a container with the config file? TODO test that
+	// TODO: Test with podman run
+
+	// --- 8< ---
+	// TODO: These options must be shared with start/bootstrap.go Where (which package) do we put them?
+
+	// TODO: Did I get all the relevant options? Too many?
+
+	cmd.Flags().String("jaeger-agent-image", "jaegertracing/jaeger-agent", "The Docker image for the Jaeger Agent")
```
pavolloffay, line 44: This is no-go to have double copy of all the flags. It should be unified with start/main.go

pkg/cmd/generate/main.go
```diff
@@ -0,0 +1,164 @@
...
+		Long: `Generate YAML manifests from Jaeger CRD.
+
+Defaults to reading Jaeger CRD from standard input and writing it to standard output, override with --cr <filename> and --output <filename>.`,
+		RunE: generate,
+	}
+
+	cmd.Flags().String("cr", "/dev/stdin", "Input Jaeger CRD")
+	viper.BindPFlag("cr", cmd.Flags().Lookup("cr"))
+
+	cmd.Flags().String("output", "/dev/stdout", "Where to print the generated YAML documents")
```
pavolloffay, line 33: note that the operator logs its operations. Stdout will be mixed with logs and the result.

pavolloffay:
Could you add a e2e test that would generate all-in-one deployment from a generated file?

pkg/cmd/generate/main.go
chlunde: Actually they log to stderr, so I think this is OK

pkg/strategy/strategy.go
```diff
@@ -209,3 +211,69 @@ func (s S) Secrets() []corev1.Secret {
 func (s S) Dependencies() []batchv1.Job {
 	return s.dependencies
 }
+
+// All returns the list of all objects for this strategy
+func (s S) All() []runtime.Object {
```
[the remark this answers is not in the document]
chlunde: It should not fail, unless the object also has a namespace, but it might cause additional retries. I found a nice resource with an order of objects here: https://github.com/kubernetes-sigs/kustomize/blob/master/api/resid/gvk.go#L77-L103 so we should use that.

pkg/account/main.go
```diff
@@ -25,6 +25,10 @@ func Get(jaeger *v1.Jaeger) []*corev1.ServiceAccount {
 func getMain(jaeger *v1.Jaeger) *corev1.ServiceAccount {
 	trueVar := true
 	return &corev1.ServiceAccount{
+		TypeMeta: metav1.TypeMeta{
```
pavolloffay, line 28: Why has this changed?

pkg/cmd/generate/main.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+// NewGenerateCommand starts the Jaeger Operator
+func NewGenerateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "generate",
+		Short: "Generate YAML manifests from Jaeger CRD",
+		Long: `Generate YAML manifests from Jaeger CRD.
+
+Defaults to reading Jaeger CRD from standard input and writing it to standard output, override with --cr <filename> and --output <filename>.`,
```
pavolloffay, line 28: and writing the manifest file to standard output

pkg/account/main.go
chlunde: This was just missing. All other types had a correct TypeMeta, but ServiceAccount did not. I guess the k8s API or client adds defaults which is why it was not a problem.

The e2e tests will catch if something else misses this, as long as it is a part of allInOne.

pkg/cmd/generate/main.go
chlunde: Now shared, if you have an idea for a better name/package, please say where :)

README.md
```diff
@@ -48,6 +48,15 @@ In this example, the Jaeger UI is available at http://192.168.122.34.
 
 The official documentation for the Jaeger Operator, including all its customization options, are available under the main [Jaeger Documentation](https://www.jaegertracing.io/docs/latest/operator/).
 
+
+## Command line invocation
+
+To dry-run, experiment or run adapt the generated output with kustomize, command line invocation using the `generate`subcommand is possible. In this example we apply the manifest generated by examples/simple-prod.yaml to the namespace `jaeger-test`:
```
pavolloffay, line 54: It might not be just dry-run some people still prefer to use plain manifest files. This paragraph should tell that `jaeger-operator generate` generates kubernetes manifest file from a given CR.

pkg/cmd/generate/main.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return &spec, nil
+}
+
+func generate(cmd *cobra.Command, args []string) error {
+	level, err := log.ParseLevel(viper.GetString("log-level"))
+	if err != nil {
+		log.SetLevel(log.InfoLevel)
+	} else {
+		log.SetLevel(level)
```
pavolloffay, line 66: Is this func returning an error?

chlunde: Same as https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/cmd/start/bootstrap.go#L207-L212

test/e2e/generator_test.go
```diff
@@ -0,0 +1,141 @@
...
+		if err != nil && strings.Contains(err.Error(), "Timeout exceeded") {
+			t.Logf("Retrying request after error %v", err)
+			return false, nil
+		}
+		require.NoError(t, err)
+
+		if res.StatusCode != 200 {
+			return false, fmt.Errorf("unexpected status code %d", res.StatusCode)
+		}
+
```
kevinearls, line 123: Can you replace this with:  require.Equal(t, 200, res.StatusCode)  Using require is more concise and also gives better error messages and a nice stacktrace when the test fails.

test/e2e/generator_test.go
```diff
@@ -0,0 +1,141 @@
...
+		body, err := ioutil.ReadAll(res.Body)
+		require.NoError(t, err)
+
+		if len(body) == 0 {
+			return false, fmt.Errorf("empty body")
+		}
+
```
kevinearls, line 130: Can you also replace this with: require.NotEqual(t, 0, len(body), "Empty body")

README.md
chlunde: Updated, but I'm not sure if you wanted the introduction sentence at all. Maybe you want to spell our the whole paragraph instead? :)

If anything, maybe we should have an example with kustomize or hint that people should use kustomize instead of manually patching the generated file. This would allow them to run generate with a newer version of the operator to get upgrades later, and hopefully not update/redo the local changes.

pkg/cmd/generate/main.go
chlunde: OK?

test/e2e/generator_test.go
chlunde: Thanks, fixes

pkg/cmd/generate/main.go
pavolloffay: looks good

pkg/cmd/start/main.go
```diff
@@ -70,6 +63,28 @@ func NewStartCommand() *cobra.Command {
...
+		Long:  "Starts a new Jaeger Operator",
+		Run: func(cmd *cobra.Command, args []string) {
+			start(cmd, args)
+		},
+	}
+
+	AddGeneratorFlags(cmd)
+
+	// Operator specific flags here. Any flags affecting manifest generation should be added to AddGeneratorFlags instead
 	cmd.Flags().String("log-level", "info", "The log-level for the operator. Possible values: trace, debug, info, warning, error, fatal, panic")
```
pavolloffay, line 88: Log level flag should be also added to the generic flag function

pkg/cmd/generate/main.go
```diff
@@ -0,0 +1,111 @@
...
+	var spec v1.Jaeger
+	decoder := yaml.NewYAMLOrJSONDecoder(f, 8192)
+	if err := decoder.Decode(&spec); err != nil && err != io.EOF {
+		return nil, err
+	}
+
+	return &spec, nil
+}
+
+func generate(cmd *cobra.Command, args []string) error {
```
pavolloffay, line 61: nit: `func generate(_ *cobra.Command, _ []string)` it's easier to read that the arguments are not used.

pkg/cmd/generate/main.go
```diff
@@ -0,0 +1,111 @@
...
+	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
+	"github.com/jaegertracing/jaeger-operator/pkg/cmd/start"
+	"github.com/jaegertracing/jaeger-operator/pkg/strategy"
+)
+
+// NewGenerateCommand starts the Jaeger Operator
+func NewGenerateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "generate",
+		Short: "Generate YAML manifests from Jaeger CRD",
```
pavolloffay, line 26: Add `(experimental)` at the beginning of the message.

pavolloffay:
thanks @chlunde  :rocket: !

PR #1061, opened 2020-05-13
Update tests to run with jaeger and otel collectors

kevinearls:

### DIFF
--- a/.ci/run-e2e-tests.sh
+++ b/.ci/run-e2e-tests.sh
@@ -9,10 +9,17 @@ sed -i 's/imagePullPolicy: Always/imagePullPolicy: Never/g' test/operator.yaml
 export VERBOSE='-v -timeout 20m'
 if [ "${TEST_GROUP}" = "es" ]; then
     echo "Running elasticsearch tests"
+    export USE_OTEL_COLLECTOR=false
+    make es
+    make e2e-tests-es
+elif [ "${TEST_GROUP}" = "es-otel" ]; then
+    echo "Running elasticsearch tests with OTEL collector"
+    export USE_OTEL_COLLECTOR=true
     make es
     make e2e-tests-es
 elif [ "${TEST_GROUP}" = "es-self-provisioned" ]; then
     echo "Running self provisioned elasticsearch tests"
+    export USE_OTEL_COLLECTOR=false
     make e2e-tests-self-provisioned-es
     res=$?
     if [[ ${res} -ne 0 ]]; then
@@ -31,6 +38,12 @@ then
 elif [ "${TEST_GROUP}" = "streaming" ]
 then
     echo "Running Streaming Tests"
+    export USE_OTEL_COLLECTOR=false
+    make e2e-tests-streaming
+elif [ "${TEST_GROUP}" = "streaming-otel" ]
+then
+    echo "Running Streaming Tests with OTEL collector"
+    export USE_OTEL_COLLECTOR=true
     make e2e-tests-streaming
 elif [ "${TEST_GROUP}" = "examples1" ]
 then
@@ -46,7 +59,7 @@ then
     make e2e-tests-token-propagation-es
 elif [ "${TEST_GROUP}" = "generate" ]
 then
-    echo "Running CLI manifest generatation tests"
+    echo "Running CLI manifest generation tests"
     make e2e-tests-generate
 else
     echo "Unknown TEST_GROUP [${TEST_GROUP}]"; exit 1
--- a/.ci/run-e2e-tests.sh
+++ b/.ci/run-e2e-tests.sh
@@ -36,7 +36,6 @@ then
 elif [ "${TEST_GROUP}" = "streaming" ]
 then
     echo "Running Streaming Tests"
-    export USE_OTEL_COLLECTOR=false
     make e2e-tests-streaming
 elif [ "${TEST_GROUP}" = "streaming-otel" ]
 then
--- a/.ci/run-e2e-tests.sh
+++ b/.ci/run-e2e-tests.sh
@@ -9,7 +9,6 @@ sed -i 's/imagePullPolicy: Always/imagePullPolicy: Never/g' test/operator.yaml
 export VERBOSE='-v -timeout 20m'
 if [ "${TEST_GROUP}" = "es" ]; then
     echo "Running elasticsearch tests"
-    export USE_OTEL_COLLECTOR=false
     make es
     make e2e-tests-es
 elif [ "${TEST_GROUP}" = "es-otel" ]; then
@@ -19,7 +18,6 @@ elif [ "${TEST_GROUP}" = "es-otel" ]; then
     make e2e-tests-es
 elif [ "${TEST_GROUP}" = "es-self-provisioned" ]; then
     echo "Running self provisioned elasticsearch tests"
-    export USE_OTEL_COLLECTOR=false
     make e2e-tests-self-provisioned-es
     res=$?
     if [[ ${res} -ne 0 ]]; then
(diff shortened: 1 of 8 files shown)


kevinearls:

test/e2e/utils.go
```diff
@@ -548,3 +549,53 @@ func deletePersistentVolumeClaims(namespace string) {
...
+
+	require.Failf(t, "Did not find a collector image for %s in namespace %s", jaegerInstanceName, namespace)
+	return false
+}
+
+// Returns a collector spec which uses the Open Telementry Collector
+func getOtelCollectorSpec(collectorSpec v1.JaegerCollectorSpec) v1.JaegerCollectorSpec {
+	otelOptions := make(map[string]interface{})
+
+	processors := map[string]string{"queued_retry": ""}
```
pavolloffay, line 581: I would prefer statically defining the `otelOptions` on the line `579`. When it's split to multiple lines it's hard to read.

test/e2e/utils.go
```diff
@@ -548,3 +549,53 @@ func deletePersistentVolumeClaims(namespace string) {
...
+			return strings.Contains(container.Image, "jaeger-opentelemetry-collector")
+		}
+	}
+
+	require.Failf(t, "Did not find a collector image for %s in namespace %s", jaegerInstanceName, namespace)
+	return false
+}
+
+// Returns a collector spec which uses the Open Telementry Collector
+func getOtelCollectorSpec(collectorSpec v1.JaegerCollectorSpec) v1.JaegerCollectorSpec {
```
pavolloffay, line 578: This is not ideal either change the signature to accept a pointer of the collector spec or return just otel config,  image and options as 3 return values.

test/e2e/elasticsearch_test.go
```diff
@@ -80,22 +81,38 @@ func (suite *ElasticSearchTestSuite) TestSparkDependenciesES() {
...
+
+func (suite *ElasticSearchTestSuite) TestSimpleProdWithOTEL() {
+	if !testOtelCollector {
+		t.Skip("Skipping OTEL collecotor test")
+	}
+	jaegerInstanceName := "simple-prod-with-otel"
+	runSimpleProdTest(jaegerInstanceName, true)
+}
+
+func runSimpleProdTest(name string, useOtelCollector bool) {
```
pavolloffay, line 96: Why is name being parametrized? How does it help with debugging?

test/e2e/elasticsearch_test.go
```diff
@@ -80,22 +81,38 @@ func (suite *ElasticSearchTestSuite) TestSparkDependenciesES() {
 }
 
 func (suite *ElasticSearchTestSuite) TestSimpleProd() {
+	jaegerInstanceName := "simple-prod"
+	runSimpleProdTest(jaegerInstanceName, false)
+}
+
+func (suite *ElasticSearchTestSuite) TestSimpleProdWithOTEL() {
```
pavolloffay, line 88: Do we need to have separate tests for OTEL?  Isn't it enough to have configurable env var for OTEL and then run the subset of tests with changed config?

test/e2e/utils.go
kevinearls: @pavolloffay Can you give me an example of how to do this?

test/e2e/elasticsearch_test.go
kevinearls: How would that work in CI?  I was assuming we'd want to run both sets of tests for the time being.

test/e2e/elasticsearch_test.go
kevinearls: It's parameterized so that I can use a different instance name when running with or without the otel collector.   I can't remember details off the top of my head but we've had times where reusing the same instance name has caused problems.

test/e2e/utils.go
pavolloffay: Something like this, the other objects can be added analogically. Btw we don't need the queued_retry processor here.
```
	otelCfg := map[string]interface{}{
		"processors": map[string]string{"queued_retry": ""},
	}
```

test/e2e/elasticsearch_test.go
pavolloffay: The CI job would set an env var to use OTEL distributions. The same test would run with a different config.

test/e2e/utils.go
```diff
@@ -574,28 +575,22 @@ func wasUsingOtelCollector(jaegerInstanceName, namespace string) bool {
...
+		"processors": map[string]string{
+			"queued_retry": "",
+		},
+		"extensions": map[string]interface{}{
+			"health_check": map[string]string{"port": "14269"},
+		},
+		"service": map[string]interface{}{
+			"pipelines": map[string]interface{}{
+				"traces": map[string]string{
+					"processors": "queued_retry",
```
pavolloffay, line 589: Why do we need `queued_retry` processor?

kevinearls: I have no idea.  I just copied the example from your PR.  :-)  

Which lines should I remove?

pavolloffay: The whole service can be removed. We just need to override the HC port.

.ci/run-e2e-tests.sh
```diff
@@ -9,10 +9,17 @@ sed -i 's/imagePullPolicy: Always/imagePullPolicy: Never/g' test/operator.yaml
 export VERBOSE='-v -timeout 20m'
 if [ "${TEST_GROUP}" = "es" ]; then
     echo "Running elasticsearch tests"
+    export USE_OTEL_COLLECTOR=false
```
pavolloffay, line 12: Rather than explicitly setting it to `false` it would be better to enable it only when it's set to `true`.

test/e2e/elasticsearch_test.go
```diff
@@ -179,6 +184,17 @@ func getJaegerSimpleProdWithServerUrls() *v1.Jaeger {
...
 	}
+
+	if useOtelCollector {
+		logrus.Infof("Using OTEL collector for %s", name)
+		otelOptions := getOtelCollectorOptions()
+		otelCollectorSpec := &v1.JaegerCollectorSpec{
+			Image:  otelCollectorImage,
+			Config: v1.NewFreeForm(otelOptions),
+		}
+		exampleJaeger.Spec.Collector = *otelCollectorSpec
```
pavolloffay, line 195: Some tests just set the image and config and others set the whole spec? Can we unify them to use the same approach?

test/e2e/streaming_test.go
```diff
@@ -191,10 +201,17 @@ func jaegerStreamingDefinition(namespace string, name string) *v1.Jaeger {
 			},
 		},
 	}
+
+	if useOtelCollector {
+		log.Infof("Using OTEL collector for %s", name)
+		j.Spec.Collector.Image = otelCollectorImage
```
pavolloffay, line 207: Related to the previous comment this one setting only the image and config.

test/e2e/utils.go
```diff
@@ -548,3 +550,37 @@ func deletePersistentVolumeClaims(namespace string) {
 		fw.KubeClient.CoreV1().PersistentVolumeClaims(kafkaNamespace).Delete(pvc.Name, &emptyDeleteOptions)
 	}
 }
+
+func verifyCollectorImage(jaegerInstanceName, namespace string, expected bool) {
+	if expected {
```
pavolloffay, line 555: The if statement can be omitted there is `require.Equal(t, expected, got)`.  Then this function becomes a one-liner :)

.ci/run-e2e-tests.sh
```diff
@@ -31,6 +36,12 @@ then
 elif [ "${TEST_GROUP}" = "streaming" ]
 then
     echo "Running Streaming Tests"
+    export USE_OTEL_COLLECTOR=false
```
pavolloffay, line 39: this can be removed false is the default

PR #1102, opened 2020-06-25
Mount volumes from agent spec

Saad-Hussain1:
Context: https://github.com/jaegertracing/jaeger-operator/issues/1097

This PR uses any `volumes` or `volumeMounts` under the **agent** node in a Jaeger instance when creating an agent as a sidecar. This still doesn't use the common spec for volumes/mounts, as it seems, *"We don't want to mount all Jaeger internal volumes into user's deployments"*.

However, if someone were to specify a volume/mount under the agent node in the Jaeger instance, they would be mounted to the sidecar, since this would be very intentional.

This seems necessary if one needs to supply files to the agent (e.g. for TLS between the agent and collector with the arg `--reporter.grpc.tls.ca`).

### DIFF
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -192,9 +192,9 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 
 	commonSpec := util.Merge([]v1.JaegerCommonSpec{jaeger.Spec.Agent.JaegerCommonSpec, jaeger.Spec.JaegerCommonSpec})
 
-	// Use a different common spec for volumes and mounts.
+	// Use only the agent common spec for volumes and mounts.
 	// We don't want to mount all Jaeger internal volumes into user's deployments
-	volumesAndMountsSpec := &v1.JaegerCommonSpec{}
+	volumesAndMountsSpec := &jaeger.Spec.Agent.JaegerCommonSpec
 	otelConf, err := jaeger.Spec.Agent.Config.GetMap()
 	if err != nil {
 		jaeger.Logger().WithField("error", err).
--- a/pkg/inject/sidecar_test.go
+++ b/pkg/inject/sidecar_test.go
@@ -158,6 +158,36 @@ func TestInjectSidecarWithEnvVarsOverridePropagation(t *testing.T) {
 	assert.Contains(t, dep.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: envVarServiceName, Value: "testapp.default"})
 }
 
+func TestInjectSidecarWithVolumeMounts(t *testing.T) {
+	// prepare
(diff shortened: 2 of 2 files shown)


pkg/inject/sidecar.go
```diff
@@ -192,9 +192,9 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 
 	commonSpec := util.Merge([]v1.JaegerCommonSpec{jaeger.Spec.Agent.JaegerCommonSpec, jaeger.Spec.JaegerCommonSpec})
 
-	// Use a different common spec for volumes and mounts.
+	// Use only the agent common spec for volumes and mounts.
 	// We don't want to mount all Jaeger internal volumes into user's deployments
-	volumesAndMountsSpec := &v1.JaegerCommonSpec{}
+	volumesAndMountsSpec := &jaeger.Spec.Agent.JaegerCommonSpec
 	otelConf, err := jaeger.Spec.Agent.Config.GetMap()
```
jpkrohling, line 198: This would require users to always specify the volume under the `agent` node in order to be consumed here, right? Meaning, if I specified a volume and mount with the TLS at the root of the CR, it will be applied to every component *except* to the deployments where the sidecar has been injected.

I like this approach, perhaps @objectiser and/or @pavolloffay would also want to give their opinion.

pkg/inject/sidecar_test.go
```diff
@@ -158,6 +158,36 @@ func TestInjectSidecarWithEnvVarsOverridePropagation(t *testing.T) {
 	assert.Contains(t, dep.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: envVarServiceName, Value: "testapp.default"})
 }
 
+func TestInjectSidecarWithVolumeMounts(t *testing.T) {
+	// prepare
```
jpkrohling, line 162: Could you please add an explicit test for the condition I mentioned in the previous comment? Add a volume/mount at the root of the spec and make sure that it does *not* get mounted.

pkg/inject/sidecar.go
Saad-Hussain1: Yes, only the volumes under `spec.agent` would be added to the deployment, not those under just `spec`.

This was done to satisfy the comment in the code: 
`// We don't want to mount all Jaeger internal volumes into user's deployments`

So all the common spec volumes aren't added to the deployment still, but if someone specifically _does_ want a volume & volumeMount added to the sidecar's deployment & sidecar container respectively, they can specify it under the `agent` node.

pkg/inject/sidecar_test.go
Saad-Hussain1: Done

jpkrohling:
LGTM. If you think this is ready to be merged, change this from a Draft PR to a regular PR :)

Saad-Hussain1:
@jpkrohling Changed to regular PR

jpkrohling:
Thank you for your contribution, @Saad-Hussain1!

PR #1111, opened 2020-07-01
Add TraceTTL to cassandra schema spec

### LINKED ISSUE #1101: Allow TRACE_TTL to be defined in cassandraCreateSchema
It seems like the it is not possible to override the `TRACE_TTL` in the `cassandraCreateSchema`, it would be very handy to allow to do so since the default is too [low (2 days) ](https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/cassandra/schema/create.sh#L11).

Seems like it could be possible to add it as another option like in here https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/storage/cassandra_dependencies.go#L38

Thanks!

moolen:
This adds the `CassandraSchemaSpec.traceTTL` field to allow the user to specify the TTL for their traces.
The TRACE_TTL env var is picked up by this script which creates the schema: https://github.com/jaegertracing/jaeger/blob/master/plugin/storage/cassandra/schema/create.sh

fixes #1101

### DIFF
--- a/pkg/apis/jaegertracing/v1/jaeger_types.go
+++ b/pkg/apis/jaegertracing/v1/jaeger_types.go
@@ -431,16 +431,33 @@ type JaegerCassandraCreateSchemaSpec struct {
 	// +optional
 	Enabled *bool `json:"enabled,omitempty"`
 
+	// Image specifies the container image to use to create the cassandra schema.
+	// The Image is used by a Kubernetes Job, defaults to the image provided through the cli flag "jaeger-cassandra-schema-image" (default: jaegertracing/jaeger-cassandra-schema).
+	// See here for the jaeger-provided image: https://github.com/jaegertracing/jaeger/tree/master/plugin/storage/cassandra
 	// +optional
 	Image string `json:"image,omitempty"`
 
+	// Datacenter is a collection of racks in the cassandra topology.
+	// defaults to "test"
 	// +optional
 	Datacenter string `json:"datacenter,omitempty"`
 
+	// Mode controls the replication factor of your cassandra schema.
+	// Set it to "prod" (which is the default) to use the NetworkTopologyStrategy with a replication factor of 2, effectively meaning
+	// that at least 3 nodes are required in the cassandra cluster.
+	// When set to "test" the schema uses the SimpleStrategy with a replication factor of 1. You never want to do this in a production setup.
 	// +optional
 	Mode string `json:"mode,omitempty"`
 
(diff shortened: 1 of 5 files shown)


moolen:
The `es-otel` test seems flaky, it runs fine locally. Can i re-run this check individually without `git commit --amend` and force-pushing?

pkg/apis/jaegertracing/v1/jaeger_types.go
```diff
@@ -431,16 +431,33 @@ type JaegerCassandraCreateSchemaSpec struct {
 	// +optional
 	Enabled *bool `json:"enabled,omitempty"`
 
+	// Image specifies the docker image to use to create the cassandra schema.
```
jpkrohling, line 434: nit: s/docker/container

jpkrohling:
LGTM. @kevinearls, is this ES failure known to you? If so, we can merge this as is, or we can try to restart the job via GitHub (which typically will cause everything to randomly fail...). Or, @moolen could try to amend and force-push and see if the test passes :-/

kevinearls:
@jpkrohling which es failure?

moolen:
I take that back, i don't think it's flaky: I reproduced the issue locally (i missed exporting the `USE_OTEL_COLLECTOR` env var before running es tests) and it continuously fails in CI. I'll dig into that.
stacktrace of the pod that runs `jaegertracing/spark-dependencies` : https://gist.github.com/moolen/091b62d537bab7f01c9f8072ad6d19a2

jpkrohling:
If you nee any help, let us know!

pkg/apis/jaegertracing/v1/jaeger_types.go
```diff
@@ -431,16 +431,33 @@ type JaegerCassandraCreateSchemaSpec struct {
 	// +optional
 	Enabled *bool `json:"enabled,omitempty"`
 
+	// Image specifies the docker container to use to create the cassandra schema.
```
jpkrohling, line 434: Sorry, not `docker container` but `container image` :)

jpkrohling: This is not a blocker, though. Would you prefer me to merge this as is, or would you want to get this change in ?

moolen: Oh, what a misunderstanding :D I'll fix that for sure!

jpkrohling:

PR #1115, opened 2020-07-03
Add support for imagePullSecrets to sidecar's Deployment

Saad-Hussain1:
Original git issue: #1018 
Previous (erroneous) PR for context: #1103 

imagePullSecrets can be configured for components by providing a custom serviceAccount for the component and specifying the imagePullSecrets in the serviceAccount. However, the agent sidecar does not support custom serviceAccounts since its Deployment may already use a serviceAccount. This PR adds support for `imagePullSecrets` to the agent spec for cases where an imagePullSecret for the sidecar is desired.

PR to update documentation: https://github.com/jaegertracing/documentation/pull/416


pkg/inject/sidecar_test.go
```diff
@@ -222,6 +222,22 @@ func TestInjectSidecarWithVolumeMounts(t *testing.T) {
...
+func TestSidecarImagePullSecrets(t *testing.T) {
+
+	agentImagePullSecrets := []corev1.LocalObjectReference{{
+		Name: "agentImagePullSecret",
+	}}
+
+	jaeger := v1.NewJaeger(types.NamespacedName{Name: "TestSidecarImagePullSecrets"})
+	jaeger.Spec.Agent.ImagePullSecrets = agentImagePullSecrets
+
+	dep := dep(map[string]string{}, map[string]string{})
```
jpkrohling, line 234: Could you add another test, showing that a pull secret that is specified at the original `Deployment` is kept, while the new `imagePullSecret` is added as well?

Saad-Hussain1: Done

jpkrohling:
Thank you for another contribution!

Issue #990, opened 2020-03-26
confused about the service generated by operator


schecterdamien:
hi.  
 		
I follow the steps of the official website document to deploy step by step。there are some problem, the deploy.yml look like this:
    
```
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: agent-as-daemonset
spec:
  agent:
    strategy: DaemonSet
    options:
      log-level: debug

```

jaeger run in observability namespace and my app run in another namespace. I want to run agent as daemonset and open a port in host machine so all my app could cross namespace pass span to agent，but:
   
<img width="845" alt="截屏2020-03-2619 49 36" src="https://user-images.githubusercontent.com/22927604/77644375-db6a1d00-6f9b-11ea-8d36-1014d544a7fe.png">

I find all service is ClusterIP，what i expect is agent-as-daemonset-agent service run in NodePort type，so i am confused.

my purpose is to run the application and jaeger in two different namespaces and could send span to the agent. Is there any correct way to do this?

r0kas:
I second to this Issue, that there should be Service Type configuration available. Now it's left for default value which is not suitable for every use case.

Following link shows where the type is not configured and left to be at it's default setting: ClusterIP
https://github.com/jaegertracing/jaeger-operator/blob/3a76357a27208b29544890ab573b523717b00d32/pkg/service/query.go#L41

GregoryVds:
Agreed. I would also like to deploy the Jeager query service as NodePort.
It seems most Helm charts expose a `service.type` option to configure such things.

Aneurysm9:
> Agreed. I would also like to deploy the Jeager query service as NodePort.

I'll second this.  I'm trying to expose the query service through an ALB ingress and that expects the service to be of type NodePort.

jpkrohling:
Alright, I think there's enough people who care about this. The best approach, IMO, is to create a new service. Who'd be the first to send a PR with a proposal? 

Basically, this here has to be changed to return a list of services, which will include the current service + the new service:
https://github.com/jaegertracing/jaeger-operator/blob/1dce126cdc3aa0f1582d53a6d62fcb359c34586c/pkg/service/query.go#L12-L53

The collector's code can be used as a reference: 
https://github.com/jaegertracing/jaeger-operator/blob/1dce126cdc3aa0f1582d53a6d62fcb359c34586c/pkg/service/collector.go#L15-L21

jpkrohling:
[...]

PR #1132, opened 2020-07-16
Create option to specify type for Query service

Aneurysm9:
This change enables the specification of a service type for the query service created by a Jaeger resource.  It addresses an issue noted by [some](https://github.com/jaegertracing/jaeger-operator/issues/990#issuecomment-616478337) [commenters](https://github.com/jaegertracing/jaeger-operator/issues/990#issuecomment-638133201) [on](https://github.com/jaegertracing/jaeger-operator/issues/990#issuecomment-655092986) #990, though I don't believe it fixes the original issue raised there.


pkg/apis/jaegertracing/v1/jaeger_types.go
```diff
@@ -200,6 +200,9 @@ type JaegerQuerySpec struct {
 
 	// +optional
 	JaegerCommonSpec `json:",inline,omitempty"`
+
+	// +optional
+	ServiceType v1.ServiceType `json:"serviceType,omitempty"`
```
jpkrohling, line 205: While this might fix the current problem, I think the operator can try to come up with the required services to create, based on the environment. For instance, if we detect that we are running on AKS, we might return a list of services that include a NodePort service (in addition, or instead of another service). Hence my suggestion in the original issue to change the following to return a list of services:

https://github.com/jaegertracing/jaeger-operator/blob/1dce126cdc3aa0f1582d53a6d62fcb359c34586c/pkg/service/query.go#L12-L53

Similar to what we have for the collector:
https://github.com/jaegertracing/jaeger-operator/blob/1dce126cdc3aa0f1582d53a6d62fcb359c34586c/pkg/service/collector.go#L15-L21

If it's a problem to have them created by default in all platforms, then flags could be added to control whether they should be created.

Aneurysm9: I've started down the path of allowing a list of ServiceType identifiers to be specified, and I can certainly make the `service.NewQueryServices()` function return a list of Service definitions, but there are a few places that assume there is a single query service that I'm not sure how to handle.  The callers of `service.NewQueryService()` are already returning lists of Service definitions, so adapting them to receive a list of Service definitions isn't too hard.  

The place I'm running into trouble is naming of the services and the places that use `service.GetNameForQueryService()` and expect a single result.  Specifically, in the `ingress.QueryIngress` and `route.QueryRoute` types.  Do you have suggestions for how to handle that, or places similar situations are handled I can reference?

jpkrohling: > there are a few places that assume there is a single query service that I'm not sure how to handle

I think those are mostly tests, but if you have questions about concrete places, share the code and I'll take a look. 

>  Do you have suggestions for how to handle that, or places similar situations are handled I can reference?

If the new service is meant to be used by external constructs, like ALB, then the presence of an ingress/route doesn't matter, right? In that case, I would leave them pointing to the current cluster IP service.

Aneurysm9: > I think those are mostly tests, but if you have questions about concrete places, share the code and I'll take a look.

The specific places I'm looking at are the [ingress](https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/ingress/query.go#L41) and [route](https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/route/query.go#L69) packages.

> If the new service is meant to be used by external constructs, like ALB, then the presence of an ingress/route doesn't matter, right? In that case, I would leave them pointing to the current cluster IP service.

The ingress is actually the important part for my use case.  We're using the [ALB Ingress Controller](https://github.com/kubernetes-sigs/aws-alb-ingress-controller) which [requires `NodePort`](https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/ingress/annotation/#traffic-routing) (or `LoadBalancer`) services.  This is part of why I assumed it would be more straightforward to allow specifying the `ServiceType`.  As `NodePort` builds on `ClusterIP`, and `LoadBalancer` builds on `NodePort`, specifying a type with a broader scope should enable additional uses without limiting existing uses.

jpkrohling: I'm almost convinced that allowing people to set the service type is a indeed a better approach, but allow me one last question: can we somehow identify that `LoadBalancer` is indeed the right default option for the current platform? Is there a way to identify that we are running on AWS? There's typically a CRD that we can check to detect the platform, like we do for OpenShift...

Second point to clarify before allowing users to set any kind of values to the service type: `NodePort` and `LoadBalancer` might be service types requiring extra permissions. We might want to prevent bad values from being set on platforms that require higher privileges and possibly write the status to the spec. At the very least, we should try to document the behavior and how to enable `NodePort`/`LoadBalancer`.

If you are able to clarify this second point, great. Otherwise, please do create an issue so that I can take a look at it later.

Aneurysm9: > can we somehow identify that `LoadBalancer` is indeed the right default option for the current platform? Is there a way to identify that we are running on AWS?

Probably.  For EKS at least there is node metadata that would indicate an EKS cluster.  I don't think that's the right thing to be looking for, however.  The determinant is the ingress controller in use.  One could use the Nginx ingress controller on EKS and potentially be fine with a `ClusterIP` service.  The operator could look for the existence of a `kubernetes.io/ingress.class` annotation on the `Ingress` definition to determine which `ServiceType` to use.  I would worry about the maintenance load of that approach were I maintaining the operator, and might want more flexibility as a consumer to choose as-yet-unsupported ingress controllers, but I think I can update this PR to work in that manner if that's what you'd prefer.

To the second point, I'm not aware of any RBAC restrictions on `ServiceType` but I am definitely not an expert in that area and cannot speak authoritatively on the topic.

jpkrohling: Let's go with the simplest approach then, and figure out how to make the operator smarter in the future (ie: auto-detect EKS, like we can auto-detect OpenShift).

> To the second point, I'm not aware of any RBAC restrictions on ServiceType but I am definitely not an expert in that area and cannot speak authoritatively on the topic.

I can try that on OpenShift and see how it goes.

pkg/apis/jaegertracing/v1/jaeger_types.go
```diff
@@ -200,6 +200,9 @@ type JaegerQuerySpec struct {
 
 	// +optional
 	JaegerCommonSpec `json:",inline,omitempty"`
+
+	// +optional
```
jpkrohling, line 204: Could you please leave a simple comment with the possible values? If you don't want to list them all, you can point the user to the right direction (like the Kubernetes' Service documentation page).

jpkrohling:
This is good to be merged, pending only a godoc for the possible values for the new property.

pkg/apis/jaegertracing/v1/jaeger_types.go
```diff
@@ -200,6 +200,13 @@ type JaegerQuerySpec struct {
...
+	// Service type represents the type of Service to create.
```
jpkrohling, line 205:
```suggestion
	// ServiceType represents the type of Service to create.
```

jpkrohling:

PR #1138, opened 2020-07-20
Updated CI to use latest versions

### LINKED ISSUE #1137: Figure out what's wrong with the CI
The `minikube` step usually takes about a couple of minutes, but it's now getting stuck "forever", eventually getting the job killed due to a timeout somewhere.

The ticket #774139 has been opened with GitHub, but their support isn't the fastest in the world, so, we might need to figure out an alternative.

Currently, the following PRs are affected: #1136, #1133 and #1045.

jpkrohling:
1. Changed the CI to use the latest Ubuntu from GitHub Actions (20.04).
1. Changed the CI to use podman instead of docker.
1. Simplified the start-minikube.sh a bit

Fixes #1137

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,6 +7,14 @@ jobs:
       - checkout
       - run:
           command: >-
+            curl -L https://golang.org/dl/go1.14.6.linux-amd64.tar.gz | tar xzf - &&
+            sudo mv go/bin/go /usr/local/bin/ &&
+            sudo mv go/bin/gofmt /usr/local/bin/ && 
+            go version
+          name: Install recent go
+
+      - run:
+          command: >-
             curl -Lo minikube https://github.com/kubernetes/minikube/releases/download/v1.12.1/minikube-linux-amd64 &&
               chmod +x minikube &&
               sudo mv minikube /usr/local/bin/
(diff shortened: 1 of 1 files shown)


jpkrohling:
This might fix this one as well: #1021

jpkrohling:
Here's a list of available software in the machine:
https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu2004-README.md

It might be interesting to try `kind` to see if it would bring faster CI cycles.

jpkrohling:
And here's a reference build: https://github.com/jpkrohling/test-minikube/runs/889417682?check_suite_focus=true

jpkrohling:
Yeah, I'm not sure what's going on with the builds. I'll try to reproduce the build on a Ubuntu 20.04 VM, but we might need to move to `kind`.

.ci/start-minikube.sh
```diff
@@ -1,51 +1,16 @@
...
-minikube update-context
 
-# waiting for node(s) to be ready
-JSONPATH='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}'; until kubectl get nodes -o jsonpath="$JSONPATH" 2>&1 | grep -q "Ready=True"; do sleep 1; done
+minikube start \
+--cpus 6 \
+--memory 8192 \
+--container-runtime=crio \
+--addons=ingress \
+--vm-driver kvm2 \
```
pavolloffay, line 10: Does GH CI allow to run virtualization? Also there is a new driver in minikube - `docker` that does not require virtualization as `none` driver.

jpkrohling: This is still WIP and not working ;-) At this state, this PR is using self-hosted runners, which happens to be a VM inside my laptop. While I think it would work, I don't think we want to manage our own set of runners.

jpkrohling:
I'm closing this, as GitHub is working again, and we can figure CircleCI out some other time.

PR #1142, opened 2020-07-21
Add link to openshift console

rubenvp8510:

### DIFF
--- a/pkg/consolelink/main.go
+++ b/pkg/consolelink/main.go
@@ -18,6 +18,7 @@ func Get(jaeger *v1.Jaeger, route *routev1.Route) *consolev1.ConsoleLink {
 	return &consolev1.ConsoleLink{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: jaeger.Namespace + ".jaeger-consolelink-" + jaeger.Name,
+			Namespace:jaeger.Namespace,
 			Labels: map[string]string{
 				"app.kubernetes.io/instance":   jaeger.Name,
 				"app.kubernetes.io/managed-by": "jaeger-operator",
--- a/pkg/consolelink/main.go
+++ b/pkg/consolelink/main.go
@@ -15,6 +15,11 @@ var RouteAnnotation = "consolelink.jaegertracing.io/route"
 
 // Get returns an ConsoleLink specification for the current instance
 func Get(jaeger *v1.Jaeger, route *routev1.Route) *consolev1.ConsoleLink {
+	// If ingress is not enable there is no reason for create a console link
+	if jaeger.Spec.Ingress.Enabled != nil && *jaeger.Spec.Ingress.Enabled == false {
+		return nil
+	}
+
 	return &consolev1.ConsoleLink{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      jaeger.Namespace + ".jaeger-consolelink-" + jaeger.Name,
@@ -39,11 +44,23 @@ func Get(jaeger *v1.Jaeger, route *routev1.Route) *consolev1.ConsoleLink {
 			},
 		},
 	}
-
 }
 
 // UpdateHref returns an ConsoleLink with the href value derived from the route
-func UpdateHref(link consolev1.ConsoleLink, route routev1.Route) consolev1.ConsoleLink {
-	link.Spec.Href = fmt.Sprintf("https://%s", route.Spec.Host)
-	return link
+func UpdateHref(routes []routev1.Route, links []consolev1.ConsoleLink) []consolev1.ConsoleLink {
+	var updated []consolev1.ConsoleLink
+	mapRoutes := make(map[string]string)
+	for _, route := range routes {
+		mapRoutes[route.Name] = route.Spec.Host
+	}
+	for _, cl := range links {
+		routeName := cl.Annotations[RouteAnnotation]
+		// Only append it if we can found the route
+		if host, ok := mapRoutes[routeName]; ok {
+			cl.Spec.Href = fmt.Sprintf("https://%s", host)
+			updated = append(updated, cl)
+		}
+		//TODO: log if not found the route associated with the link
+	}
+	return updated
 }
--- a/pkg/consolelink/main.go
+++ b/pkg/consolelink/main.go
@@ -13,7 +13,7 @@ import (
 // RouteAnnotation used to annotate the link with the route name
 var RouteAnnotation = "consolelink.jaegertracing.io/route"
 
-// Get returns an ConsoleLink specification for the current instance
+// Get returns a ConsoleLink specification for the current instance
 func Get(jaeger *v1.Jaeger, route *routev1.Route) *consolev1.ConsoleLink {
 	// If ingress is not enable there is no reason for create a console link
 	if jaeger.Spec.Ingress.Enabled != nil && *jaeger.Spec.Ingress.Enabled == false {
@@ -22,10 +22,12 @@ func Get(jaeger *v1.Jaeger, route *routev1.Route) *consolev1.ConsoleLink {
 
 	return &consolev1.ConsoleLink{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      jaeger.Namespace + ".jaeger-consolelink-" + jaeger.Name,
-			Namespace: jaeger.Namespace,
+			Name:      "jaeger-" + jaeger.Namespace + "-" + jaeger.Name,
+			Namespace: jaeger.Namespace, // Prevent warning at creation time.
 			Labels: map[string]string{
-				"app.kubernetes.io/instance":   jaeger.Name,
+				"app.kubernetes.io/instance": jaeger.Name,
+				// Allow distinction between same jaeger instances in different namespaces
+				"app.kubernetes.io/namespace":  jaeger.Namespace,
 				"app.kubernetes.io/managed-by": "jaeger-operator",
 			},
 			Annotations: map[string]string{
--- a/pkg/consolelink/main.go
+++ b/pkg/consolelink/main.go
@@ -1,0 +1,43 @@
+package consolelink
+
+import (
+	"fmt"
+
+	consolev1 "github.com/openshift/api/console/v1"
+	routev1 "github.com/openshift/api/route/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
+)
+
+// RouteAnnotation used to annotate the link with the route name
+var RouteAnnotation = "consolelink.jaegertracing.io/route"
+
+// Get returns an ConsoleLink specification for the current instance
+func Get(jaeger *v1.Jaeger, route *routev1.Route) *consolev1.ConsoleLink {
+	return &consolev1.ConsoleLink{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: jaeger.Namespace + ".jaeger-consolelink-" + jaeger.Name,
+			Labels: map[string]string{
+				"app.kubernetes.io/instance":   jaeger.Name,
+				"app.kubernetes.io/managed-by": "jaeger-operator",
+			},
+			Annotations: map[string]string{
+				RouteAnnotation: route.Name,
+			},
+		},
+		Spec: consolev1.ConsoleLinkSpec{
+			Location:        consolev1.ApplicationMenu,
+			ApplicationMenu: &consolev1.ApplicationMenuSpec{},
+			Link: consolev1.Link{
+				Text: "Jaeger [" + jaeger.Namespace + "]",
+			},
+		},
+	}
+
+}
+
+// UpdateHref returns an ConsoleLink with the href value derived from the route
+func UpdateHref(link *consolev1.ConsoleLink, route routev1.Route) {
+	link.Spec.Href = fmt.Sprintf("https://%s", route.Spec.Host)
+}
--- a/pkg/consolelink/main.go
+++ b/pkg/consolelink/main.go
@@ -30,7 +30,7 @@ func Get(jaeger *v1.Jaeger, route *routev1.Route) *consolev1.ConsoleLink {
 			Location:        consolev1.ApplicationMenu,
 			ApplicationMenu: &consolev1.ApplicationMenuSpec{},
 			Link: consolev1.Link{
-				Text: "Jaeger [" + jaeger.Namespace + "]",
+				Text: "Jaeger [" + jaeger.Name + "." + jaeger.Namespace + "]",
 			},
 		},
 	}
--- a/pkg/consolelink/main.go
+++ b/pkg/consolelink/main.go
@@ -13,6 +13,11 @@ import (
 // RouteAnnotation used to annotate the link with the route name
 var RouteAnnotation = "consolelink.jaegertracing.io/route"
 
+//Name derived a console link resource name from jaeger instance
+func Name(jaeger *v1.Jaeger) string {
+	return "jaeger-" + jaeger.Namespace + "-" + jaeger.Name
+}
+
 // Get returns a ConsoleLink specification for the current instance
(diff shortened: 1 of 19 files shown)


pkg/controller/jaeger/consolelink_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
+	"github.com/jaegertracing/jaeger-operator/pkg/strategy"
+)
+
+func TestConsoleLinkCreate(t *testing.T) {
+	// prepare
+	nsn := types.NamespacedName{
+		Name: "TestCronJobsCreate",
```
jpkrohling, line 22: This name is wrong. Please, use `my-instance` for new tests, as having unique names isn't bringing much value (and causes copy/paste problems like the above ;) )

pkg/controller/jaeger/consolelink_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	viper.Set("platform", "openshift")
```
jpkrohling, line 24: I think you are missing a `defer viper.Reset()`

pkg/controller/jaeger/consolelink_test.go
```diff
@@ -0,0 +1,201 @@
...
+	assert.Equal(t, persistedName.Name, persisted.Name)
+	assert.NoError(t, err)
+}
+
+func TestConsoleLinkUpdate(t *testing.T) {
+	// prepare
+	nsn := types.NamespacedName{
+		Name: "TestConsoleLinkUpdate",
+	}
+	viper.Set("platform", "openshift")
```
jpkrohling, line 66: `defer viper.Reset()` here too

pkg/controller/jaeger/consolelink_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	err = cl.Get(context.Background(), persistedName, persisted)
+	assert.Equal(t, "new-value", persisted.Annotations["key"])
+	assert.NoError(t, err)
+}
+
+func TestConsoleLinkDelete(t *testing.T) {
+	// prepare
+	nsn := types.NamespacedName{
+		Name: "TestCronJobsDelete",
```
jpkrohling, line 109: Name :)

pkg/controller/jaeger/consolelink_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	viper.Set("platform", "openshift")
```
jpkrohling, line 111: defer

pkg/controller/jaeger/consolelink_test.go
```diff
@@ -0,0 +1,201 @@
...
+	// prepare
+	nsn := types.NamespacedName{
+		Name:      "TestCronJobsCreateExistingNameInAnotherNamespace",
+		Namespace: "tenant1",
+	}
+	nsnExisting := types.NamespacedName{
+		Name:      "TestCronJobsCreateExistingNameInAnotherNamespace",
+		Namespace: "tenant2",
+	}
+	viper.Set("platform", "openshift")
```
jpkrohling, line 155: defer

pkg/controller/jaeger/consolelink_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	err = cl.Get(context.Background(), persistedName, persisted)
+	assert.Empty(t, persisted.Name)
+	assert.Error(t, err) // not found
+}
+
+func TestConsoleLinksCreateExistingNameInAnotherNamespace(t *testing.T) {
+	// prepare
+	nsn := types.NamespacedName{
+		Name:      "TestCronJobsCreateExistingNameInAnotherNamespace",
```
jpkrohling, line 148: name

pkg/controller/jaeger/consolelink_test.go
```diff
@@ -0,0 +1,201 @@
...
+		Namespace: "tenant1",
+	}
+	nsnExisting := types.NamespacedName{
+		Name:      "TestCronJobsCreateExistingNameInAnotherNamespace",
```
jpkrohling, line 152: and here too

pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -336,6 +336,10 @@ func (r *ReconcileJaeger) apply(ctx context.Context, jaeger v1.Jaeger, str strat
 		if err := r.applyRoutes(ctx, jaeger, str.Routes()); err != nil {
 			return jaeger, tracing.HandleError(err, span)
 		}
+		if err := r.applyConsoleLink(ctx, jaeger, str.ConsoleLinks()); err != nil {
+			jaeger.Logger().WithError(tracing.HandleError(err, span)).Warn("failed to reconcile console links")
+			return jaeger, nil
```
jpkrohling, line 341: I think we should just log the error and let the function continue.

pkg/strategy/all_in_one.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"strings"
 
+	"github.com/jaegertracing/jaeger-operator/pkg/consolelink"
```
jpkrohling, line 7: This belongs together with the other in-module imports (third group)

pkg/strategy/all_in_one.go
```diff
@@ -82,6 +84,9 @@ func newAllInOneStrategy(ctx context.Context, jaeger *v1.Jaeger) S {
 	if viper.GetString("platform") == v1.FlagPlatformOpenShift {
 		if q := route.NewQueryRoute(jaeger).Get(); nil != q {
 			c.routes = append(c.routes, *q)
+			if link := consolelink.Get(jaeger, q); link != nil {
```
jpkrohling, line 87: This whole block looks strange, as neither `route.NewQueryRoute` nor `consolelink.Get` return nils. I think it can be simplified, can't it? They might even return the values directly instead of pointers.

pkg/strategy/production.go
```diff
@@ -87,6 +88,9 @@ func newProductionStrategy(ctx context.Context, jaeger *v1.Jaeger) S {
 	if viper.GetString("platform") == v1.FlagPlatformOpenShift {
 		if q := route.NewQueryRoute(jaeger).Get(); nil != q {
 			c.routes = append(c.routes, *q)
+			if link := consolelink.Get(jaeger, q); link != nil {
```
jpkrohling, line 91: It seems that you are missing tests for this. Same for the other strategies.

pkg/consolelink/main.go
```diff
@@ -0,0 +1,43 @@
...
+			},
+			Annotations: map[string]string{
+				RouteAnnotation: route.Name,
+			},
+		},
+		Spec: consolev1.ConsoleLinkSpec{
+			Location:        consolev1.ApplicationMenu,
+			ApplicationMenu: &consolev1.ApplicationMenuSpec{},
+			Link: consolev1.Link{
+				Text: "Jaeger [" + jaeger.Namespace + "]",
```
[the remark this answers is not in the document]
rubenvp8510: Yes, well we can use the CR name , to distinguish between them.

pkg/consolelink/main.go
```diff
@@ -0,0 +1,43 @@
...
+			},
+		},
+	}
+
+}
+
+// UpdateHref returns an ConsoleLink with the href value derived from the route
```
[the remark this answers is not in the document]
rubenvp8510: +1 for returning a new ConsoleLink, I'll change.

pkg/controller/jaeger/consolelink_test.go
rubenvp8510: ooops, you caught me.. I did a copy&paste..  :flushed: I'll fix this.

pkg/strategy/all_in_one.go
rubenvp8510: Well, route.NewQueryRoute(jaeger).Get() could potentially return nil if ingress is not enable.https://github.com/jaegertracing/jaeger-operator/blob/fd64f15cdd13bbe8360ca9fad2178a06838e3101/pkg/route/query.go#L25

In the case of consoleLink Would we want to have a flag for disable/enable it? If not of course I can simplify those lines. :)

pkg/consolelink/main.go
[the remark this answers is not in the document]
rubenvp8510: I changed, now I return a copy.

pkg/strategy/production.go
rubenvp8510: Added.

pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -336,6 +336,9 @@ func (r *ReconcileJaeger) apply(ctx context.Context, jaeger v1.Jaeger, str strat
 		if err := r.applyRoutes(ctx, jaeger, str.Routes()); err != nil {
 			return jaeger, tracing.HandleError(err, span)
 		}
+		if err := r.applyConsoleLinks(ctx, jaeger, r.updateHref(ctx, jaeger, str.ConsoleLinks())); err != nil {
```
jpkrohling, line 339: The controller shouldn't know how to transform the existing console link into desired. In other words, the `updateHref` logic shouldn't happen here.

The way our operator works, this code here should only call `applyConsoleLinks` passing `str.ConsoleLinks()` as parameter, which builds the list of desired console links. The `applyConsoleLinks` will then get this list over to the inventory, which will compare the existing with the desired and build a map of operations: what should be created, what should be updated, and what should be deleted. The inventory should take care of carrying over the value from the desired to the existing, in case of an update (which is what your `updateHref` is doing). Take a look at what we are doing with the secret or deployment. The batch job isn't a good example.

pkg/inventory/consolelink.go
```diff
@@ -0,0 +1,65 @@
...
+	mcreate := consoleLinkMap(desired)
+	mdelete := consoleLinkMap(existing)
+
+	for k, v := range mcreate {
+		if t, ok := mdelete[k]; ok {
+			tp := t.DeepCopy()
+			util.InitObjectMeta(tp)
+
+			// we can't blindly DeepCopyInto, so, we select what we bring from the new to the old object
+			tp.Spec = v.Spec
```
jpkrohling, line 28: This makes me wonder whether your `updateHref` is actually needed. This should be taking care of updating the Href already.

pkg/strategy/all_in_one.go
jpkrohling: If we don't have a query, do we want a console link? I have a feeling that this will not run cleanly in plain Kubernetes. It might work right now because this is shielded at the controller, though.

IMO, if the route is protecting itself by checking whether the platform is openshift, then the console link should do the same.

pkg/consolelink/main.go
```diff
@@ -0,0 +1,45 @@
...
+			Namespace:jaeger.Namespace,
+			Labels: map[string]string{
+				"app.kubernetes.io/instance":   jaeger.Name,
+				"app.kubernetes.io/managed-by": "jaeger-operator",
+			},
+			Annotations: map[string]string{
+				RouteAnnotation: route.Name,
+			},
+		},
+		Spec: consolev1.ConsoleLinkSpec{
```
objectiser, line 30: This is using a different structure to the examples of [kiali](https://github.com/kiali/kiali-operator/blob/v1.18.2/roles/default/kiali-deploy/templates/openshift/console-links.yaml#L13-L32) and the [blog](https://www.openshift.com/blog/openshift-4-3-the-project-launcher).

We need the console link to only display in the namespace where the Jaeger instance is deployed.

pkg/consolelink/main.go
```diff
@@ -0,0 +1,45 @@
...
+			Location:        consolev1.ApplicationMenu,
+			ApplicationMenu: &consolev1.ApplicationMenuSpec{},
+			Link: consolev1.Link{
+				Text: "Jaeger [" + jaeger.Name + "." + jaeger.Namespace + "]",
```
objectiser, line 34: Can you remove the namespace from the link text, as it will only be shown when viewing that namespace in the console.

pkg/strategy/all_in_one.go
rubenvp8510: This is what is happening here, isn't it? If there is no query, (if query returns nil) it won't create a link.  Indeed this is checking the platform two lines above. It check the platform, then try to create a `queryRoute`, if query router is successfully created, then try to create a `consoleLink`.

pkg/inventory/consolelink.go
rubenvp8510: May be I chose a bad name for the function, the propose of `updateHref` is to fetch the route, take the Host and assign it to the Href part of the console CR. The think is, the Host in the route CR is empty at the moment of creating the route.We need to create the route first, then get it to know the resulting host.

pkg/consolelink/main.go
rubenvp8510: Sure

pkg/consolelink/main.go
rubenvp8510: Done

pkg/consolelink/main.go
rubenvp8510: Done

jpkrohling:
There's something odd. I have two instances in two namespaces, with the same instance name:

```console
$ kubectl get jaegers --all-namespaces
NAMESPACE       NAME       STATUS    VERSION   STRATEGY   STORAGE   AGE
myapp           simplest   Running   1.18.1    allinone   memory    2d2h
observability   simplest   Running   1.18.1    allinone   memory    2m42s
```

And somehow, the console links got kinda messed up:

```console
$ kubectl get consolelinks --all-namespaces
NAME                                TEXT                URL                                               MENU   AGE
myapp.jaeger-consolelink-simplest   Jaeger [simplest]   https://simplest-observability.apps-crc.testing          9m2s
```

The link ended showing up in the web console for the `myapp` namespace, but with the link pointing to the instance at `observability`.

pkg/controller/jaeger/consolelink.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/jaegertracing/jaeger-operator/pkg/tracing"
+)
+
+func (r *ReconcileJaeger) applyConsoleLinks(ctx context.Context, jaeger v1.Jaeger, desired []osconsolev1.ConsoleLink) error {
+	tracer := global.TraceProvider().GetTracer(v1.ReconciliationTracer)
+	ctx, span := tracer.Start(ctx, "applyConsoleLinks")
+	defer span.End()
+
+	opts := []client.ListOption{
+		client.InNamespace(jaeger.Namespace),
```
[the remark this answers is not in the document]
rubenvp8510: I'm ok with 1), We should check for cluster admin before trying to create cluster object.
 
2) Is not clear to me, I can fetch all consolelinks but checking for `app.kubernetes.io/instance`? In the case of two instances with same name in different namespaces, I'll have two of them with the same name... and we will have the same problem... or am I missunderstanding something? I was thinking on adding a new label with the namespace, so I can filter by namespace AND jaeger instance name. Does that make sense?

pkg/controller/jaeger/consolelink_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+	objs := []runtime.Object{
+		v1.NewJaeger(nsn),
+		v1.NewJaeger(nsnExisting),
+		&osconsolev1.ConsoleLink{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      nsnExisting.Name,
+				Namespace: nsnExisting.Namespace,
+				Annotations: map[string]string{
+					consolelink.RouteAnnotation: "my-route-1",
```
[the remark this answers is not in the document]
rubenvp8510: Yes.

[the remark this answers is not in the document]
rubenvp8510: This should test when the route, and jaeger instance names are the same, but they are on different namespaces. We neet to make sure that the associated route is the correct one, the one that is in the same namespace as the jaeger instance.

PR #1144, opened 2020-07-27
Add common field to  jaeger-es-rollover-create-mapping

lighteness:
Happen to see the job `jaeger-es-rollover-create-mapping` can't be like other components,`cronjob.batch/jaeger-es-index-cleaner, cronjob.batch/jaeger-es-lookback, cronjob.batch/jaeger-es-rollover` to have Affinity, Tolerations, SecurityContext and ServiceAccountName

### DIFF
--- a/pkg/storage/elasticsearch_dependencies.go
+++ b/pkg/storage/elasticsearch_dependencies.go
@@ -1,7 +1,6 @@
 package storage
 
 import (
-	"github.com/jaegertracing/jaeger-operator/pkg/account"
 	"strings"
 
 	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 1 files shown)


pkg/storage/elasticsearch_dependencies.go
```diff
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"github.com/jaegertracing/jaeger-operator/pkg/account"
```
pavolloffay, line 4: this should be moved to the last import group

lighteness: thanks for the comment, the chance has been done

lighteness:
what i need to do for the next step in order to make code merged?

pavolloffay:
thanks @lighteness

Issue #716, opened 2019-10-21
Flaky tests on 'pkg/storage/elasticsearch_secrets_test.go'


volmedo:
Some tests in [`elasticsearch_secrets_test.go`](https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/storage/elasticsearch_secrets_test.go) use `assert.EqualError` to check for specific error conditions. IMHO, this is not a great approach in general, but in this case things are worse because the error message is coming from the OS. Thus, tests running on different platforms could throw different error messages and make the tests fail.

Concretely, [`TestExtractSecretsToFile_Err`](https://github.com/jaegertracing/jaeger-operator/blob/01e39a38e95e6edb1dbc5c0edf042d6ed9df0c0e/pkg/storage/elasticsearch_secrets_test.go#L112) and [`TestWriteToWorkingDir`](https://github.com/jaegertracing/jaeger-operator/blob/01e39a38e95e6edb1dbc5c0edf042d6ed9df0c0e/pkg/storage/elasticsearch_secrets_test.go#L138) are failing on my machine (macOS 10.15 Catalina). The tests expect a permissions problem, but in my machine the file cannot be written because the file system is mounted in read-only mode, so the message is different.

[`TestCreateESSecretsError `](https://github.com/jaegertracing/jaeger-operator/blob/01e39a38e95e6edb1dbc5c0edf042d6ed9df0c0e/pkg/storage/elasticsearch_secrets_test.go#L20) also uses `assert.EqualError` but is not failing for me.

Tests should be refactored to avoid depending on specific error message strings if possible. If the underlying error doesn't provide more information about its nature (like it having a more specific type, for example), a possibility could be to wrap it into a new error that does.

PR #1151, opened 2020-08-09
Avoid error message assertions on OS dependent errors (#716)

albertteoh:
A proposed fix to Issue #716 where tests are failing on macOS Catalina, which look like:
```
--- FAIL: TestExtractSecretsToFile_Err (0.00s)
    elasticsearch_secrets_test.go:114: 
                Error Trace:    elasticsearch_secrets_test.go:114
                Error:          Error message not equal:
                                expected: "open /root/foo: permission denied"
                                actual  : "mkdir /root: read-only file system"
                Test:           TestExtractSecretsToFile_Err
--- FAIL: TestWriteToWorkingDir (0.00s)
    elasticsearch_secrets_test.go:164: 
                Error Trace:    elasticsearch_secrets_test.go:164
                Error:          Error message not equal:
                                expected: "mkdir /foo: permission denied"
                                actual  : "mkdir /foo: read-only file system"
                Test:           TestWriteToWorkingDir
    elasticsearch_secrets_test.go:164: 
                Error Trace:    elasticsearch_secrets_test.go:164
                Error:          Error message not equal:
                                expected: "open /root/bla: permission denied"
                                actual  : "mkdir /root: read-only file system"
                Test:           TestWriteToWorkingDir
```

As alluded to in @volmedo's suggestion, the change here is to assert on the error type `os.PathError` instead of the OS-specific error message.

Given [`TestExtractSecretsToFile_FileExists`](https://github.com/jaegertracing/jaeger-operator/blob/01e39a38e95e6edb1dbc5c0edf042d6ed9df0c0e/pkg/storage/elasticsearch_secrets_test.go#L138) is a table-driven tests with the possibility of more than one error type as an outcome of a particular permutation of inputs, it was decided to use reflection and compare on the error type rather than a `err boolean` and hard-code a comparison to `os.PathError`.

For [`TestExtractSecretsToFile_Err`](https://github.com/jaegertracing/jaeger-operator/blob/01e39a38e95e6edb1dbc5c0edf042d6ed9df0c0e/pkg/storage/elasticsearch_secrets_test.go#L112), using reflection would unnecessarily complicate the code as there is only one possible error outcome.

As for [`TestCreateESSecretsError`](https://github.com/jaegertracing/jaeger-operator/blob/01e39a38e95e6edb1dbc5c0edf042d6ed9df0c0e/pkg/storage/elasticsearch_secrets_test.go#L20), the error returned may differ between OSes, but it does not appear to be a known breaking issue AFAIK.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -963,6 +963,7 @@ github.com/spf13/viper v1.4.0 h1:yXHLWeravcrgGyFSyCgdYpXQ9dR9c/WED3pg1RhxqEU=
 github.com/spf13/viper v1.4.0/go.mod h1:PTJ7Z/lr49W6bUbkmS1V3by4uWynFiR9p7+dSq/yZzE=
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/objx v0.2.0 h1:Hbg2NidpLE8veEBkEZTL3CvlkUIVzuU9jDplZO54c48=
 github.com/stretchr/objx v0.2.0/go.mod h1:qt09Ya8vawLte6SNmTgCsAVtYtaKzEcn8ATUoHMkEqE=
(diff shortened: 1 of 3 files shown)


albertteoh:
I'm not sure why pkg/apis/jaegertracing/v1/options.go had caused reduced codecov/project coverage when I didn't change that file, nor modify non-test code.  Anyone know what's causing the alert and how I can fix it?

jpkrohling:

PR #1153, opened 2020-08-11
Support custom labels in Jaeger all-in-one deployments (#629)

albertteoh:
Proposed fix for Issue #629. Wrote test to fail where "name", "hello" and "another" custom labels have empty corresponding strings in the selectors, then update selector to use `commonSpec.Labels`, matching the template's labels. Tests pass after change.

Diff also contains a few gofmt and go.sum updates.

### DIFF
--- a/pkg/deployment/all_in_one.go
+++ b/pkg/deployment/all_in_one.go
@@ -37,7 +37,6 @@ func NewAllInOne(jaeger *v1.Jaeger) *AllInOne {
 // Get returns a pod for the current all-in-one configuration
 func (a *AllInOne) Get() *appsv1.Deployment {
 	a.jaeger.Logger().Debug("Assembling an all-in-one deployment")
-	labels := a.labels()
 	trueVar := true
 
 	args := append(a.jaeger.Spec.AllInOne.Options.ToArgs())
@@ -51,7 +50,7 @@ func (a *AllInOne) Get() *appsv1.Deployment {
 			"sidecar.istio.io/inject": "false",
 			"linkerd.io/inject":       "disabled",
 		},
-		Labels: labels,
+		Labels: a.labels(),
 	}
 
 	commonSpec := util.Merge([]v1.JaegerCommonSpec{a.jaeger.Spec.AllInOne.JaegerCommonSpec, a.jaeger.Spec.JaegerCommonSpec, baseCommonSpec})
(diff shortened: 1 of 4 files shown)


pkg/deployment/all_in_one.go
```diff
@@ -120,7 +120,7 @@ func (a *AllInOne) Get() *appsv1.Deployment {
 		},
 		Spec: appsv1.DeploymentSpec{
 			Selector: &metav1.LabelSelector{
-				MatchLabels: labels,
```
jpkrohling, line 123: Is `labels` being used elsewhere? I have the feeling that it should be the other way around: both should use `labels`, and `labels` should take `commonSpec.Labels` into consideration. This is because `labels` might have different values, more specific values than `commonSpec.Labels`.

albertteoh: `labels` looks like it's the `default` set of recommended labels (in [utils.go](https://github.com/jaegertracing/jaeger-operator/blob/55f0191947a60e2d995353c1a6aa21ea619d39bb/pkg/util/util.go#L163)) is only used in two places:
- to initialise the `baseCommonSpec` ([L54](https://github.com/jaegertracing/jaeger-operator/blob/11a063f97c97560f1a811d95478828b68c7aeba1/pkg/deployment/all_in_one.go#L54)) which appears to be the `default` spec.
- and this line in question, which is to use this `default` set of labels as the label selector.

The more specific set of labels is actually `commonSpec.Labels` as it is the result of merging the `labels` with the custom set of labels defined in `a.jaeger.Spec.AllInOne.JaegerCommonSpec` and `a.jaeger.Spec.JaegerCommonSpec` (see [L57](https://github.com/jaegertracing/jaeger-operator/blob/11a063f97c97560f1a811d95478828b68c7aeba1/pkg/deployment/all_in_one.go#L57)).

If you agree with the above, it does raise an interesting point that perhaps the variable naming could be improved as I would agree the name `labels` sounds more specific than `commonSpec.Labels`. Perhaps `labels` could be renamed to `defaultLabels` or `recommendedLabels` and `commonSpec` -> `mergedSpec`?

jpkrohling: If this line is being changed to use `commonSpec.Labels`, then we don't need `labels` as a var anymore. We could use `a.labels()` directly in [L54](https://github.com/jaegertracing/jaeger-operator/blob/11a063f97c97560f1a811d95478828b68c7aeba1/pkg/deployment/all_in_one.go#L54)

jpkrohling:
Thank you, @albertteoh!

PR #1161, opened 2020-08-19
Disabled service links

### LINKED ISSUE #1158: Disable service links to prevent environment variable conflict
It is common to name the Jaeger agent service `jaeger-agent`. This forces Kubernetes to create an environment variable `JAEGER_AGENT_PORT` as part of the service links. This conflicts with the an internal environmental variable used in the client library.

https://github.com/jaegertracing/jaeger-client-go/blob/704ce2849e9abb8af82edb2a3375e9e163092427/config/config_env.go#L51

This will cause the query UI to crash with the following error:
```
{"level":"fatal","ts":1597712251.5144613,"caller":"command-line-arguments/main.go:76","msg":"Failed to read tracer configuration","error":"cannot obtain reporter config from env: cannot parse env var JAEGER_AGENT_PORT=udp://100.69.94.248:5775: strconv.ParseInt: parsing \"udp://100.69.94.248:5775\": invalid syntax" ...
```

If this variable is set it is meant to be `6831` but Kubernetes sets it to the first port listed in the service for the agent. Jaeger query UI does not expect this value.

Service links can be disabled as part of the pod specification.

> If the service environment variables are not desired (because possible clashing with expected program ones, too many variables to process, only using DNS, etc) you can disable this mode by setting the `enableServiceLinks` flag to false on the pod spec.

See also: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#pod-v1-core

By automatically setting `enableServiceLinks` to `false` this edge case will no longer occur.

This should be added to all deployment specifications created by the operator.

mikelorant:
Disabled information about services being injected into pod's environment variables.

This prevents service links environmental variables from conflicting with internal Jaeger environmental variables. 

Image has been verified to work as expected. For verification purposes the image is available at:
fairfaxmedia/jaeger-operator:latest

Thanks to @dackroyd for assistance in these creating this fix.

Fixes #1158 
Relates to jaegertracing/jaeger#1986

### DIFF
--- a/pkg/deployment/agent.go
+++ b/pkg/deployment/agent.go
@@ -65,6 +65,7 @@ func (a *Agent) Get() *appsv1.DaemonSet {
 	adminPort := util.GetPort("--admin-http-port=", args, 14271)
 
 	trueVar := true
+	falseVar := false
 	labels := util.Labels(a.name(), "agent", *a.jaeger)
 
 	baseCommonSpec := v1.JaegerCommonSpec{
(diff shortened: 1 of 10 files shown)


jpkrohling (approved):
LGTM!

jpkrohling:

PR #1167, opened 2020-08-24
Apply the securityContext to the cassandraCreateSchema job

### LINKED ISSUE #1166: Allow configuring the securityContext of the CassandraCreateSchema job
Currently, the CassandraCreateSchema job does not take the securityContext specified in Jaeger.spec.securityContext into account: https://github.com/jaegertracing/jaeger-operator/blob/9cde42a9ccc291545de8cfaf7a9dbebf6b15af12/pkg/storage/cassandra_dependencies.go#L129

This causes the job not to start when running in a cluster with very restrictive PSPs.

I think this could be resolved quite easily by setting the Jaeger.spec.securityContext (globally set in the Jaeger CRD) to the Job's container/pod securityContext. I could create a PR for this, although it would be the my first time writing Go 😅

chgl:
Resolves #1166

### DIFF
--- a/pkg/storage/cassandra_dependencies.go
+++ b/pkg/storage/cassandra_dependencies.go
@@ -134,6 +134,7 @@ func cassandraDeps(jaeger *v1.Jaeger) []batchv1.Job {
 					},
 					Spec: corev1.PodSpec{
 						ActiveDeadlineSeconds: podTimeout,
+						SecurityContext:       jaeger.Spec.SecurityContext,
 						Containers: []corev1.Container{{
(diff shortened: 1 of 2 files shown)


jpkrohling (commented):
The fix looks good, would you be able to add a test as well, to ensure we don't have a regression in the future?

chgl:
@jpkrohling does this look ok?

jpkrohling (approved):
LGTM! Will merge on green.

jpkrohling:

chgl:
Thank you and the team for making the contribution experience so smooth :). Hope this doesn't sound impatient: is there an ETA for the next release?

jpkrohling:
You are actually lucky: we are planning a release this week.

chgl:
Great to hear, thank you!

PR #1168, opened 2020-08-25
Remove explicitly setting agent's reporter type

pavolloffay:
Tchannel reporter has been removed in 1.18.0
The GRPC reporter was set as default in 1.11.0

This also helps with using OTEL based images since they do not expose `--reporter.type` flag.

### DIFF
--- a/deploy/examples/statefulset-manual-sidecar.yaml
+++ b/deploy/examples/statefulset-manual-sidecar.yaml
@@ -43,4 +43,3 @@ spec:
               protocol: TCP
           args:
             - --reporter.grpc.host-port=dns:///jaeger-collector-headless.example-ns:14250
-            - --reporter.type=grpc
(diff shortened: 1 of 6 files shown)


jpkrohling (approved):
I'm OK with this change, but has this flag been deprecated? If not, it would be good to deprecate it now and set a date for removal.

pavolloffay:
The flag hasn't been deprecated or removed, but there is only one reporter type and no current plans to add another one, hence we can use the default.

jpkrohling:
From our operator's perspective, sure. But I mean that other people out there might be using this flag as well. Just removing them without deprecating will break them.

pavolloffay:
How would it break them? The flag is still supported in Jaeger.

jpkrohling:
Sorry, my bad. I thought that having this flag would cause an error with the new binaries.

PR #1176, opened 2020-09-01
Create daemonsets after services and deployments.

### LINKED ISSUE #1175: DaemonSet started before deployment can't get collector IPs
Currently, we have daemonsets being created before services and deployments:

https://github.com/jaegertracing/jaeger-operator/blob/6451f060aa759ff9cd82b39371c666b12ffcd2e2/pkg/controller/jaeger/jaeger_controller.go#L321-L334

Depending on the timing, the daemonset's pod might be up and running before the deployment gets even created, so that the headless service won't get a chance to get its IP for at least one backing collector. This causes the daemonset to not find any IPs upon bootstrap. The grpc-go client will not attempt to do any DNS refreshes at all (or only after 30m?), turning the agent unusable in this scenario.

This *might* be related to grpc/grpc-go#3353, so, it might be worth also bumping gRPC for Jaeger itself.

jpkrohling:
Fixes #1175

### DIFF
--- a/pkg/controller/jaeger/jaeger_controller.go
+++ b/pkg/controller/jaeger/jaeger_controller.go
@@ -318,10 +318,6 @@ func (r *ReconcileJaeger) apply(ctx context.Context, jaeger v1.Jaeger, str strat
 		return jaeger, tracing.HandleError(err, span)
 	}
 
-	if err := r.applyDaemonSets(ctx, jaeger, str.DaemonSets()); err != nil {
-		return jaeger, tracing.HandleError(err, span)
-	}
-
 	// seems counter intuitive to have services created *before* deployments,
 	// but some resources used by deployments are created by services, such as TLS certs
 	// for the oauth proxy, if one is used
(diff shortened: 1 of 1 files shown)


pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -358,5 +354,11 @@ func (r *ReconcileJaeger) apply(ctx context.Context, jaeger v1.Jaeger, str strat
 		return jaeger, nil
 	}
 
+	// we apply the daemonsets after everything else, to increase the changes of having services and deployments
```
kevinearls, line 357: I'm guessing you mean "increase the chances"

jpkrohling: Fixed!

jpkrohling:
@kevinearls if this looks good to you, I'll merge it.

kevinearls:
@jpkrohling Unfortunately this is still failing for me.  It's slightly better, as it passed 6 out of 10 times as opposed to 5 out of 10 previously, but it doesn't seem to be completely resolved.

jpkrohling:
This won't be 100% until the real fix (jaegertracing/jaeger#2443) is in place.

kevinearls:
@jpkrohling Is that likely to happen soon?  If not I can either skip this test or see if I can come up with a retry mechanism of some sort.

jpkrohling:
I'm working on it at the moment, but a release is unlikely to happen during the next couple of weeks.

PR #1180, opened 2020-09-03
Changed agent-as-daemonset image for the jaeger-agent

jpkrohling:


jpkrohling:
@kevinearls the fix has landed in master with a different gRPC version than the PR that was used to build the current image. Using `latest` is typically something to avoid, as it might mean anything (not only the latest master), but Jaeger does push to latest only on commits landing to master, so, it should be safe to use it.

kevinearls (approved):
Works for me.

PR #1188, opened 2020-09-09
FOSSA scan enabled

idvoretskyi:
Ref.: https://github.com/cncf/foundation/issues/109

Signed-off-by: Ihor Dvoretskyi <ihor@linux.com>

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -1,0 +1,27 @@
+name: FOSSA
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "^1.14.x"
+      - run: go version
+      # Runs a set of commands to initialize and analyze with FOSSA
+      - name: run FOSSA analysis
+        env:
+          # FOSSA Push-Only API Token
+          FOSSA_API_KEY: '304657e2357ba57b416b94e6b119131b'
+        run: |
+          export GOPATH=$HOME/go
+          export PATH=$PATH:$(go env GOPATH)/bin
+          curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh | bash
+          fossa init
+          fossa analyze


idvoretskyi:
cc @yurishkuro

jpkrohling:
LGTM, but I don't seem to have maintainer access to the repo: 
image

.github/workflows/fossa.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "^1.14.x"
+      - run: go version
+      # Runs a set of commands to initialize and analyze with FOSSA
+      - name: run FOSSA analysis
+        env:
+          # FOSSA Push-Only API Token
+          FOSSA_API_KEY: '304657e2357ba57b416b94e6b119131b'
```
jpkrohling, line 21: Sorry if this has been asked before, but shouldn't this be set in an env var, as a secret?

idvoretskyi: @jpkrohling it can't be set as a secret because it won't be passed as a variable when someone will submit a PR from the fork (GitHub Actions do not pass secrets from forks).

This is an API token that can be safely exposed publicly (it was _designed_ to be exposed publicly) - https://docs.fossa.com/docs/api-reference#push-only-api-token.

idvoretskyi:
@jpkrohling I've invited you, let me know if you received an invitation.

jpkrohling:
I received, but looks like it's for registration? I have an account already (jpkroehling), and accessing the link from the email did not made me join the organization.

idvoretskyi:
I see. The issue is that FOSSA doesn't allow users to be members of multiple organizations (@yurishkuro had the same issue).

I can ask FOSSA support to move you from the existing organization to the CNCF one, or you can create a new FOSSA account for this purpose.

jpkrohling:
> to move you from the existing organization to the CNCF one

Sure, sounds good.

idvoretskyi:
@jpkrohling this felt into cracks, my apologies. If this is still actual, I'll go ahead and work on adding you to the CNCF FOSSA org.

jpkrohling:
We still don't have FOSSA in this repo, and as long as contributors are able to clearly understand what is wrong in case of a failure, I think it would be nice to have.

jpkrohling:
The check seems to be failing:

```
github.com/spf13/pflag imported from implicitly required module; to add
missing requirements, run:
	go get github.com/spf13/pflag@v1.0.5
LINK: https://github.com/fossas/fossa-cli/blob/master/docs/integrations/golang.md#analysis

FATAL Could not parse modules: No modules provided
```

PR #1190, opened 2020-09-10
Added configuration for the agent's securityContext

### LINKED ISSUE #1186: Allow for configuring the injected sidecar container's securityContext
Currently, the securityContext of the injected jaeger-agent container can not be configured, causing the agent container to fail with "container has runAsNonRoot and image will run as root" when running in a cluster with strict PSPs.

We could reuse the securityContext from the common jaeger spec (as in #1166) or add a a securityContext element to the Spec.Agent type.

In addition, a lot of headaches may be avoided by setting the the jaeger-agent container user to a non-root one by default. This should be realtively straight-forward as I don't think the jaeger-agent requires any privileges or disk access. I'd like to create a PR for this, but I can't find the jaeger-agent image's Dockerfile.

chgl:
Unfortunately, I can't just use the Jaeger CommonSpec's securityContext as that is a `PodSecurityContext` whereas the sidecar can only have a `SecurityContext`. This required me to add an additional config option `containerSecurityContext`. This does add some redundancy as the `Jaeger.Spec.SecurityContext` is not reused.

Is `sidecarContainerSecurityContext` a better name for this config option as the other agent deployments do use the default pod security context?

Resolves #1186

### DIFF
--- a/deploy/crds/jaegertracing.io_jaegers_crd.yaml
+++ b/deploy/crds/jaegertracing.io_jaegers_crd.yaml
@@ -662,6 +662,56 @@ spec:
                   type: object
                 serviceAccount:
                   type: string
+                sidecarSecurityContext:
+                  properties:
+                    allowPrivilegeEscalation:
+                      type: boolean
+                    capabilities:
+                      properties:
+                        add:
+                          items:
+                            type: string
+                          type: array
+                        drop:
+                          items:
+                            type: string
+                          type: array
+                      type: object
+                    privileged:
+                      type: boolean
+                    procMount:
+                      type: string
+                    readOnlyRootFilesystem:
+                      type: boolean
+                    runAsGroup:
+                      format: int64
+                      type: integer
+                    runAsNonRoot:
+                      type: boolean
+                    runAsUser:
+                      format: int64
+                      type: integer
+                    seLinuxOptions:
+                      properties:
+                        level:
+                          type: string
+                        role:
+                          type: string
+                        type:
+                          type: string
+                        user:
+                          type: string
+                      type: object
+                    windowsOptions:
+                      properties:
+                        gmsaCredentialSpec:
+                          type: string
+                        gmsaCredentialSpecName:
+                          type: string
+                        runAsUserName:
+                          type: string
+                      type: object
+                  type: object
                 strategy:
                   type: string
                 tolerations:
--- a/pkg/apis/jaegertracing/v1/jaeger_types.go
+++ b/pkg/apis/jaegertracing/v1/jaeger_types.go
@@ -383,6 +383,9 @@ type JaegerAgentSpec struct {
 
 	// +optional
 	Config FreeForm `json:"config,omitempty"`
+
+	// +optional
+	SidecarSecurityContext *v1.SecurityContext `json:"sidecarSecurityContext,omitempty"`
 }
 
 // JaegerStorageSpec defines the common storage options to be used for the query and collector
--- a/pkg/apis/jaegertracing/v1/zz_generated.deepcopy.go
+++ b/pkg/apis/jaegertracing/v1/zz_generated.deepcopy.go
@@ -133,6 +133,11 @@ func (in *JaegerAgentSpec) DeepCopyInto(out *JaegerAgentSpec) {
 	in.Options.DeepCopyInto(&out.Options)
 	in.JaegerCommonSpec.DeepCopyInto(&out.JaegerCommonSpec)
 	in.Config.DeepCopyInto(&out.Config)
+	if in.SidecarSecurityContext != nil {
+		in, out := &in.SidecarSecurityContext, &out.SidecarSecurityContext
+		*out = new(corev1.SecurityContext)
+		(*in).DeepCopyInto(*out)
+	}
 	return
 }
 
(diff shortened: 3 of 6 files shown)


jpkrohling:
That's very unfortunate. Most of the fields that are commonly used are in both objects, but unless we settle for our own lowest common denominator, we can't reuse the info. 

About the name: what do you think about `sidecarSecurityContext`?

chgl:
@jpkrohling Oh, I like that one! Done.

Just a quick question because I'm not too failiar with the CRD generation process, running `make generate` creates additional changes to the jaegertracing.io_jaegers_crd.yaml beyond just the added sidecarSecurityContext:

```diff
diff --git a/deploy/crds/jaegertracing.io_jaegers_crd.yaml b/deploy/crds/jaegertracing.io_jaegers_crd.yaml
index d3c60575..580f8c90 100644
--- a/deploy/crds/jaegertracing.io_jaegers_crd.yaml
+++ b/deploy/crds/jaegertracing.io_jaegers_crd.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.4
+    controller-gen.kubebuilder.io/version: v0.3.0
   creationTimestamp: null
   name: jaegers.jaegertracing.io
 spec:
@@ -589,6 +589,7 @@ spec:
                         type: string
                     type: object
                   type: array
+                  x-kubernetes-list-type: atomic
                 labels:
                   additionalProperties:
                     type: string
@@ -600,11 +601,19 @@ spec:
                   properties:
                     limits:
                       additionalProperties:
-                        type: string
+                        anyOf:
+                        - type: integer
+                        - type: string
+                        pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                        x-kubernetes-int-or-string: true
                       type: object
```

Is this OK?

jpkrohling:
No, that's not expected. You probably have a different Operator SDK version than the one that was last used with `make generate`. I think we are still with v0.18.2:

https://github.com/jaegertracing/jaeger-operator/blob/c6ba26b0a08d5b96c750724f8c217e2f4a39321f/.github/workflows/base-checks.yaml#L14

Which version are you using?

chgl:
> Which version are you using?

```sh
$ operator-sdk version
operator-sdk version: "v0.18.2", commit: "f059b5e17447b0bbcef50846859519340c17ffad", kubernetes version: "v1.18.2", go version: "go1.13.10 linux/amd64"
```
However running `make generate` causes an error so I had to manually install `controller-gen`, `openapi-gen`, etc. via `go install sigs.k8s.io/controller-tools/cmd/controller-gen` - which probably installed a wrong version. Which version should the `*-gen` tools be? (or am I doing something entirely wrong?)

```sh
make generate
./.ci/generate.sh: line 28: /home/chgl/go/bin/controller-gen: No such file or directory
Failed to generate CRDs.
make: *** [Makefile:309: internal-generate] Error 127
```

jpkrohling:
We have a make target (`install-tools`) for installing the tools we need for all other targets:

https://github.com/jaegertracing/jaeger-operator/blob/5f7472f33e2e9fbb12c4b8250ea23e5b1e41f656/Makefile#L329-L337

I believe this should use the information from the `go.mod`. Could you check whether we have this info in the CONTRIBUTING.md? If we don't, it would also be a great contribution to the project to add this info there ;-)

jpkrohling (changes requested):
I'm blocking this from being merged for now, to sort out the controller-gen changes. The remaining of the PR is looking good.

chgl:
Thank you, I completely missed the `make install-tools` or `make install` (although I'd have expected that one to actually install the operator as opposed to installing build tools). I can definitely update the docs on this, but maybe making `install-sdk` invoke `install-tools` makes more sense? Or simply suggest a single `make install` as part of the build setup.


I've been struggling quite a bit to get it working, even downgrading go from 1.15 to 1.14, but there are still quite a lot of diffs between master and what a local `make generate` does. To reproduce:

```Dockerfile
FROM ubuntu:20.04
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
    build-essential \
    git \
    curl \
    golang-1.14

RUN export GOROOT=/usr/lib/go-1.14 && \
    export PATH=$PATH:/usr/lib/go-1.14/bin/ && \
    export PATH=$PATH:/root/go/bin && \
    git clone https://github.com/jaegertracing/jaeger-operator.git && \
    cd jaeger-operator && \
    make install && \
    make generate && \
    git --no-pager diff
```

It always updates the kubebuilder version to v0.3.0 and mostly seems to add additional validation around resource limits/requests (which doesn't seem too bad (?)). I still don't understand why.

jpkrohling:
> which doesn't seem too bad

No, the change in itself isn't bad. Just worrying that this happens unrelated to the actual change. I can confirm that I see the same once I delete the cached binary that I had. I wonder why the CI isn't having this...

> I still don't understand why.

I can't either, but I can confirm that it does work when I do a `go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4`.

chgl:
@jpkrohling  I've tried again with  sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4 and it looks better now. Thank you for your help & patience!

jpkrohling (approved):
LGTM!

PR #1198, opened 2020-09-17
RFC:Add resource limit to example operator

chenfengjin:


jpkrohling:
I think it's a good idea to place some limits, but those seem quite high. How did you figure these values out? I'd rather say that we request 128MiB, with a limit of 1GiB (or even 512MiB). Are your operators really consuming up to 8GiB during a peak?

chenfengjin:
@jpkrohling  As you recommend, I modify memory limit to 128/1024 Mi

jpkrohling:
@chenfengjin do you have some metrics to share about the operator's resource consumption? Even my numbers are just guesses, it would be good to be backed by metrics from a real system using it.

PR #1201, opened 2020-09-20
remove metrics and servicemonitor

sschne:
Currently jaeger-operator is creating its own service and serviceMonitor objects in Kubernetes. This seems to be bad practice and the functions `metrics.CreateMetricsService` and `metrics.CreateServiceMonitors` are removed in the operator-sdk by 1.0.0 https://github.com/operator-framework/operator-sdk/pull/3484 .

Instead, the service and servicemonitor objects are to be created with the Helm chart https://github.com/jaegertracing/helm-charts/pull/152

### DIFF
--- a/deploy/service_monitor.yaml
+++ b/deploy/service_monitor.yaml
@@ -1,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: jaeger-operator-metrics
+  labels:
+    name: jaeger-operator
+spec:
+  endpoints:
+    - path: /metrics
+      port: http-metrics
+    - path: /metrics
+      port: cr-metrics
+  selector:
+    matchLabels:
+      name: jaeger-operator
--- a/pkg/cmd/start/bootstrap.go
+++ b/pkg/cmd/start/bootstrap.go
@@ -11,7 +11,6 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	kubemetrics "github.com/operator-framework/operator-sdk/pkg/kube-metrics"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
-	"github.com/operator-framework/operator-sdk/pkg/metrics"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	"go.opentelemetry.io/otel/api/core"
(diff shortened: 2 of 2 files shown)


jpkrohling:
> This seems to be bad practice

Do you have any further info on that?

sschne:
The scope of an operator should be limited to managing the resources that are defined by its CRD. As reference you could take the operator pattern documentation in Kubernetes https://kubernetes.io/docs/concepts/extend-kubernetes/operator/
It says that "Operators are software extensions to Kubernetes that make use of custom resources to manage applications and their components."

The service and servicemonitor resources the operator is deploying here, are not part of any CRD, so the operator should not touch them. This would also remove the confusion that is currently ongoing as the Helm chart is deploying the metrics service and the operator is overwriting this very service, mixing up scopes.

jpkrohling:
I agree in principle with you. For the Operator SDK 1.0, `ServiceMonitors` will be part of the OLM bundle, but not sure how that implies for our current version. Are you able to find out if service monitors can be added to OLM bundles for the version of the SDK we are currently using? Just removing the service monitor would break the expectation for people currently installing the operator via OLM, so, we probably need to adjust the bundle generation.

sschne:
Sadly i have no experience in Openshift and OLM. I tried to figure out how to get OLM running on native Kubernetes, but without success. Should we add a condition to only remove servicemonitor when running on Kubernetes / installed with Helm?

jpkrohling:
> Should we add a condition to only remove servicemonitor when running on Kubernetes / installed with Helm?

If it's not urgent, I prefer to assign this to me and try it out on OpenShift/OLM in the next few weeks.

sschne:
It is not urgent, thanks for looking into it.

jpkrohling:
I just tried it out and it doesn't look like we can merge this PR here, unfortunately. I added a new `service_monitor.yaml` in the `deploy` directory, which is typically the source of resources when generating a CSV. There might be a filter on which resource types are included in the CSV, because that wasn't enough to include it there.

At this moment, I don't know how we could include the service monitor as part of the bundle for this operator-sdk's version, so, we'd be breaking compat with people expecting this service monitor to exist.

For completeness, I updated this PR to include the service monitor that would be required to be part of the bundle. Additionally, I rebased the branch.

To test this, you can download the operator-sdk v0.18.2 and run this:

```
$ operator-sdk generate csv --csv-channel=stable --make-manifests=false --csv-version=1.21.0 --from-version=1.20.0
```

It should result in a CSV in `deploy/olm-catalog/jaeger-operator/1.21.0/jaeger-operator.v1.21.0.clusterserviceversion.yaml`. To proceed with this PR here, we'd need the service monitor to be included in this file.

jpkrohling:
Looks like the Operator SDK 1.2.0 will allow this: https://github.com/operator-framework/operator-sdk/pull/4137 .

@sschne I'm very sorry for closing this PR without merging it, but I don't know a solution for this that won't require us to migrate to Operator SDK 1.2.0. Moving to that version isn't an easy task and this is being tracked in another issue. Thank you for the time you invested in this PR, and again, sorry for closing it without merging.

PR #1212, opened 2020-09-24
Completely replace the sidecar on each reconciliation, call patch instead of update.

### LINKED ISSUE #1089: Noticing weird kubernetes pods being deployed and terminated immediately
When I use the annotation  "sidecar.jaegertracing.io/inject": "true" , there are extra pods that get created that terminate right away. IN my example below, the two api pods that get created named `api-5955f589d5-fkpfg` and `api-5955f589d5-rhhm6` are created and immediately terminated. the correct replicaset that ends up being persisted and deployed are the first 3 api pods (with ages 1s and 2s). Anybody run into this before?

```$ k get po
NAME                              READY   STATUS              RESTARTS   AGE
api-57ffc5c59b-bq2vb              0/2     ContainerCreating   0          1s
api-57ffc5c59b-hvr9l              0/2     ContainerCreating   0          1s
api-57ffc5c59b-jv5zx              0/2     ContainerCreating   0          2s
api-5955f589d5-8fjml              2/2     Running             0          74m 
api-5955f589d5-fkpfg              1/2     Terminating         0          9s
api-5955f589d5-rhhm6              1/2     Running             0          9s
client-78d7747758-b8m2b           1/1     Running             0          74m
client-78d7747758-wrmrp           0/1     Terminating         0          9s
client-7c74cd6d89-ptn2z           0/1     ContainerCreating   0          1s
client-7c74cd6d89-q2lxg           0/1     ContainerCreating   0          2s```

### PR DESCRIPTION
**Pull request by rubenvp8510**
This is an attempt to stabilize the injection mechanism, before this patch we created a new sidecar and compare the new sidecar with the old one. The issue with this approach is the function that creates the sidecar uses the length of `dep.Spec.Template.Spec.Containers` to derive part of the arguments (See https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/inject/sidecar.go#L198) . So in the first reconciliation process it inject one sidecar, but the next reconciliation process will inject a different sidecar with an extra argument. This triggers the pod re-creation process twice.

This approach recreated the sidecar in each deployment reconciliation 

This also uses patch instead of update in an effort to alleviate some errors related to the version of the resource.

Fixes #1089

### DIFF
--- a/pkg/controller/deployment/deployment_controller.go
+++ b/pkg/controller/deployment/deployment_controller.go
@@ -157,24 +157,34 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 				}
 			}
 
-			// a suitable jaeger instance was found! let's inject a sidecar pointing to it then
-			// Verified that jaeger instance was found and is not marked for deletion.
-			log.WithFields(log.Fields{
-				"deployment":       dep.Name,
-				"namespace":        dep.Namespace,
-				"jaeger":           jaeger.Name,
-				"jaeger-namespace": jaeger.Namespace,
-			}).Info("Injecting Jaeger Agent sidecar")
-
-			injectedDep := inject.Sidecar(jaeger, dep.DeepCopy())
-
-			if !inject.EqualSidecar(injectedDep, dep) {
-				if err := r.client.Update(ctx, injectedDep); err != nil {
+			depHasAgent, depAgentIndex := inject.HasJaegerAgent(dep)
+			patch := client.MergeFrom(dep.DeepCopy())
+			if !depHasAgent {
+				// a suitable jaeger instance was found! let's inject a sidecar pointing to it then
+				// Verified that jaeger instance was found and is not marked for deletion.
+				log.WithFields(log.Fields{
+					"deployment":       dep.Name,
+					"namespace":        dep.Namespace,
+					"jaeger":           jaeger.Name,
+					"jaeger-namespace": jaeger.Namespace,
+				}).Info("Injecting Jaeger Agent sidecar")
+				injectedDep := inject.Sidecar(jaeger, dep)
+				if err := r.client.Patch(ctx, injectedDep, patch); err != nil {
+					log.WithField("deployment", injectedDep).WithError(err).Error("failed to update")
+					return reconcile.Result{}, tracing.HandleError(err, span)
+				}
+			} else {
+				// This could be an update, re-inject the sidecar in case the parameters changed.
+				// if any change is detected the cluster will create a new version of the deployment
+				// otherwise it will do nothing.
+				containers := dep.Spec.Template.Spec.Containers
+				dep.Spec.Template.Spec.Containers = append(containers[:depAgentIndex], containers[depAgentIndex+1:]...)
+				injectedDep := inject.Sidecar(jaeger, dep)
+				if err := r.client.Patch(ctx, injectedDep, patch); err != nil {
 					log.WithField("deployment", injectedDep).WithError(err).Error("failed to update")
 					return reconcile.Result{}, tracing.HandleError(err, span)
 				}
 			}
-
 		} else {
 			log.WithField("deployment", dep.Name).Info("No suitable Jaeger instances found to inject a sidecar")
 		}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rubenvp8510** [comment]
Before this patch

```
[rvargasp@rubens-laptop examples]$ oc apply -f business-application-injected-sidecar.yaml
[rvargasp@rubens-laptop examples]$ oc rollout history deployment/myapp
deployments "myapp"
REVISION  CHANGE-CAUSE
1         <none>
2         <none>
3         <none>
```


After this patch:
```

[rvargasp@rubens-laptop examples]$ oc apply -f business-application-injected-sidecar.yaml
[rvargasp@rubens-laptop examples]$ oc rollout history deployment/myapp
deployments "myapp"
REVISION  CHANGE-CAUSE
1         <none>
2         <none>

```

Also, the message errors `"the object has been modified; please apply your changes to the latest version and try again" `were gone.

File: pkg/controller/deployment/deployment_controller.go [hunk 1]
```diff
@@ -157,24 +157,34 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
...
+				if err := r.client.Patch(ctx, injectedDep, patch); err != nil {
+					log.WithField("deployment", injectedDep).WithError(err).Error("failed to update")
+					return reconcile.Result{}, tracing.HandleError(err, span)
+				}
+			} else {
+				// This could be an update, re-inject the sidecar in case the parameters changed.
+				// if any change is detected the cluster will create a new version of the deployment
+				// otherwise it will do nothing.
+				containers := dep.Spec.Template.Spec.Containers
+				dep.Spec.Template.Spec.Containers = containers[:depAgentIndex+copy(containers[depAgentIndex:],containers[depAgentIndex+1:])]
```
jpkrohling, line 181: What is this line doing? I get the `copy` part, but what does it do when combined with `containers[:depAgentIndex]`?

pkg/controller/namespace/namespace_controller.go
```diff
@@ -136,7 +136,7 @@ func (r *ReconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Res
 				log.WithError(err).Error("failed to get the available Jaeger pods")
 				return reconcile.Result{}, tracing.HandleError(err, span)
 			}
-
+			patch := client.MergeFrom(dep.DeepCopy())
```
jpkrohling, line 139: Can it be moved closer to where it's being used? Also, you might not need `dep.DeepCopy()`

pkg/inject/sidecar.go
```diff
@@ -375,5 +375,6 @@ func EqualSidecar(dep, oldDep *appsv1.Deployment) bool {
 	}
 	depContainer := dep.Spec.Template.Spec.Containers[depAgentIndex]
 	oldDepContainer := oldDep.Spec.Template.Spec.Containers[oldDepIndex]
+	// Only compare args, if args are equal, assuming sidecars are equal
```
jpkrohling, line 378: What's "compare args" here? Arguments for the containers? The `DeepEqual` is comparing the whole container, isn't it?

jpkrohling (commented):
We should really consider writing unit tests for this...

pkg/inject/sidecar.go
rubenvp8510: Old comment. I forgot to remove it.

pkg/controller/namespace/namespace_controller.go
rubenvp8510: Not sure I don't need the deepcopy, the thing is, the structure could have some other nested structures as a pointers. and if I just use dep the tests doesn't pass :(

pkg/inject/sidecar.go
jpkrohling: As soon as it's removed, I'll approve this PR and it will be merged when the CI passes.

pkg/controller/deployment/deployment_controller.go
rubenvp8510: Changed for a more readable version.

pkg/inject/sidecar.go
rubenvp8510: Removed.

jpkrohling (approved):

PR #1229, opened 2020-10-01
Create ingress-with-annotations.yaml

prageethw:
sample to enable ingress with some annotations.


deploy/examples/ingress-with-annotations.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: jaegertracing.io/v1
+kind: Jaeger
+metadata:
+  name: jaeger
```
jpkrohling, line 4: We typically use the example name in this field: `name: ingress-with-annotations` would be more appropriate.

Still on the topic of naming: I think the example name (and file name) should actually be: `ingress-with-host`.

deploy/examples/ingress-with-annotations.yaml
```diff
@@ -0,0 +1,19 @@
...
+spec:
+  strategy: allInOne 
+  storage:
+    type: memory 
+    options:
```
jpkrohling, line 9: Could you please leave the example with only the minimum set of fields required to work?

deploy/examples/ingress-with-annotations.yaml
```diff
@@ -0,0 +1,19 @@
...
+  strategy: allInOne 
+  storage:
+    type: memory 
+    options: 
+      memory: 
+        max-traces: 1000000
+  ingress:
+    enabled: true
+    annotations:
+      kubernetes.io/ingress.class: nginx
```
jpkrohling, line 15: Is this needed?

deploy/examples/ingress-with-annotations.yaml
```diff
@@ -0,0 +1,19 @@
...
+    hosts:
+      - mesh-jaeger.xxx.com #your domain name.
+  agent:
```
jpkrohling, line 18: This node can also be removed.

jpkrohling (commented):
Thank you for providing this example! I made a few comments, but it's looking good.

jpkrohling:
The DCO check is failing. Could you please sign-off your commit? Instructions are here: https://github.com/jaegertracing/jaeger-operator/runs/1192444215

deploy/examples/ingress-with-annotations.yaml
prageethw: I was thinking to show how to use ingress annotations here

deploy/examples/ingress-with-annotations.yaml
prageethw: agent removed

deploy/examples/ingress-with-annotations.yaml
jpkrohling: Then we can have two examples, each with one use-case? Unless all the people looking for an example of how to set the annotation will also need to set the host, and vice-versa.

prageethw:
closing and will make separate pull with 2 examples

PR #1230, opened 2020-10-02
Added liveness and readiness probes to injected sidecar

### LINKED ISSUE #1202: Readiness probes missing from the injected sidecar container
The injected sidecar agent currently does not set either a readiness or liveness probe. The other deployment types do: 

https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/deployment/agent.go#L156-L175

kadern0:
Signed-off-by: Pablo Caderno <p_caderno@wargaming.net>

Fixes #1202


pkg/inject/sidecar.go
```diff
@@ -274,6 +275,26 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 				Name:          "admin-http",
 			},
 		},
+		LivenessProbe: &corev1.Probe{
+			Handler: corev1.Handler{
+				HTTPGet: &corev1.HTTPGetAction{
+					Path: "/",
+					Port: intstr.FromInt(int(adminPort)),
```
jpkrohling, line 282: What happens if the other container, with the business application, exposes the same port?

jpkrohling (commented):
Looks good, and I don't think the CI failure is due to this PR. In any case, I have one question, and one request:

1. What happens when the other container (with the business application) already exposes the same port? Should we do a check here and just not add the checks in that case?
1. Please, add a test for this, or change an existing one, asserting that those checks are in place. If we need a logic to handle the case above, a test is also needed to verify that behavior.

pkg/inject/sidecar.go
kadern0: Hi @jpkrohling that's a very good point, is there already any logic to verify that the other ports do not clash either?

jpkrohling: I don't think we have it, as this is the only place where we deal with resources we don't create ourselves. Shouldn't be hard, though: I think we can just skip adding the probes in case the main container has the port being declared?

kadern0: @jpkrohling I've added a function to check if the port is available.

pkg/inject/sidecar.go
```diff
@@ -278,6 +279,29 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 		SecurityContext: jaeger.Spec.Agent.SidecarSecurityContext,
 		VolumeMounts:    volumesAndMountsSpec.VolumeMounts,
 	}
+	if IsJaegerPortAvailable(adminPort, dep) {
```
jpkrohling, line 282: A better name could probably be `ContainerPort` instead of `JaegerPort`.

pkg/inject/sidecar.go
```diff
@@ -367,6 +391,20 @@ func HasJaegerAgent(dep *appsv1.Deployment) (bool, int) {
 	return false, -1
 }
 
+// IsJaegerPortAvailable checks whether deployment is already using some port
+func IsJaegerPortAvailable(port int32, dep *appsv1.Deployment) bool {
```
jpkrohling, line 395: This is currently being used only by the package itself, I would make it private for now.

pkg/inject/sidecar.go
```diff
@@ -367,6 +391,20 @@ func HasJaegerAgent(dep *appsv1.Deployment) (bool, int) {
...
+	// this pod is annotated, it should have a sidecar
+	// but does it already have one?
```
jpkrohling, line 397: Is this comment part of a copy/paste? Seems like it belongs to some other place :-)

jpkrohling (commented):
Looks good, but perhaps the function could have a better name?

pkg/inject/sidecar.go
kadern0: Agree 100%

pkg/inject/sidecar.go
kadern0: :man_facepalming:

jpkrohling:
@kadern0 , looks like the DCO check is failing: would you please take a look? Some tests failed, and I just restarted the CI.

jpkrohling (approved):
LGTM.

frzifus (commented):
is that still a thing? could you try to update the branch @rubenvp8510 ?

frzifus:
I will close it, since its inactive.

PR #1203, opened 2020-09-21
Multi arch docker build

morlay:
I maintained a multi arch images for istio stack https://github.com/querycap/istio
since istio-operator deprecated addonComponents, need jaeger-operator support muti arch.

PR #1239, opened 2020-10-07
S390x changes

madeelrh:


Makefile
```diff
@@ -1,5 +1,5 @@
 VERSION_DATE ?= $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
-GO_FLAGS ?= GOOS=linux GOARCH=amd64 CGO_ENABLED=0 GO111MODULE=on
+GO_FLAGS ?= GOOS=linux GOARCH=s390x CGO_ENABLED=0 GO111MODULE=on
```
jpkrohling, line 2: We probably want a multi-arch image instead, like the one being proposed as part of #1203.

Makefile
```diff
@@ -8,7 +8,7 @@ FMT_LOG=fmt.log
 
 OPERATOR_NAME ?= jaeger-operator
 NAMESPACE ?= "$(USER)"
-BUILD_IMAGE ?= "$(NAMESPACE)/$(OPERATOR_NAME):latest"
+BUILD_IMAGE ?= "quay.io/maistra/jaeger-rhel7-operator:1.17.6-ibm"
```
jpkrohling, line 11: The proper way to override this is via command line, like:

```console
make run BUILD_IMAGE=quay.io/maistra/jaeger-rhel7-operator:1.17.6-ibm
```

pkg/cmd/start/main.go
```diff
@@ -18,11 +18,11 @@ import (
 func AddFlags(cmd *cobra.Command) {
 	cmd.Flags().String("log-level", "info", "The log-level for the operator. Possible values: trace, debug, info, warning, error, fatal, panic")
 	cmd.Flags().String("jaeger-version", version.DefaultJaeger(), "Deprecated: the Jaeger version is now managed entirely by the operator. This option is currently no-op.")
-	cmd.Flags().String("jaeger-agent-image", "jaegertracing/jaeger-agent", "The Docker image for the Jaeger Agent")
-	cmd.Flags().String("jaeger-query-image", "jaegertracing/jaeger-query", "The Docker image for the Jaeger Query")
-	cmd.Flags().String("jaeger-collector-image", "jaegertracing/jaeger-collector", "The Docker image for the Jaeger Collector")
-	cmd.Flags().String("jaeger-ingester-image", "jaegertracing/jaeger-ingester", "The Docker image for the Jaeger Ingester")
-	cmd.Flags().String("jaeger-all-in-one-image", "jaegertracing/all-in-one", "The Docker image for the Jaeger all-in-one")
+	cmd.Flags().String("jaeger-agent-image", "quay.io/maistra/jaeger-agent-rhel7:1.17.6-ibm", "The Quay image for the Jaeger Agent")
```
jpkrohling, line 21: Can't you just pass the flags instead? You can change the following:

https://github.com/jaegertracing/jaeger-operator/blob/729d8aaaa08aada07a790a900a18867dbd98deb1/deploy/operator.yaml#L24

To use:

```yaml
        args: ["start", "--jaeger-all-in-one-image=quay.io/maistra/jaeger-all-in-one-rhel7:1.17.6-ibm"]
```

This should be applied to your local `operator.yaml`. This change really doesn't belong to this repo here.

jpkrohling (commented):
I might be missing something, but I don't think these changes belong to this repository here... @jkandasa, do you have more context for this?

pkg/cmd/start/main.go
jkandasa: @madeelrh when you install jaeger-operator from OperatorHub these changes will be included. No actions required here.

test/e2e/daemonset_test.go
```diff
@@ -171,7 +171,7 @@ func getVertxDeployment(namespace string, selector map[string]string) *appsv1.De
 				},
 				Spec: corev1.PodSpec{
 					Containers: []corev1.Container{{
-						Image: "jaegertracing/vertx-create-span:operator-e2e-tests",
+						Image: "quay.io/maistra/jaeger-vertx-create-span:0.0-ibm-z",
```
jkandasa, line 174: @madeelrh we should pass the image details as an environment variable and the default image should be `jaegertracing/vertx-create-span:operator-e2e-tests`
like this https://github.com/jaegertracing/jaeger-operator/blob/master/test/e2e/utils.go#L43~L52

test/e2e/daemonset_test.go
```diff
@@ -196,6 +196,7 @@ func getVertxDeployment(namespace string, selector map[string]string) *appsv1.De
 								},
 							},
 							InitialDelaySeconds: 1,
+							TimeoutSeconds:      3,
```
jkandasa, line 199: @madeelrh what is the default timeout before? Does it cause any issue?

test/e2e/sidecar_test.go
```diff
@@ -160,7 +160,7 @@ func getVertxDefinition(deploymentName string, annotations map[string]string) *a
 				},
 				Spec: corev1.PodSpec{
 					Containers: []corev1.Container{{
-						Image: "jaegertracing/vertx-create-span:operator-e2e-tests",
+						Image: "quay.io/maistra/jaeger-vertx-create-span:0.0-ibm-z",
```
jkandasa, line 163: @madeelrh same as above, should be changed via environment variable

test/e2e/daemonset_test.go
madeelrh: Default is 1 second. The Vertx code has a delay of 1 second. With the delay in place, the Liveness service fails with 1 second default timeout. 
The delay can be found here:
 https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_jaegertracing_vertx-2Dcreate-2Dspan_blob_7014795d0d59f22c218b5dd4edc6d9bda0b9496b_src_main_java_com_example_demo_MainVerticle.java-23L92&d=DwIDaQ&c=jf_iaSHvJObTbx-siA1ZOg&r=1FWlI4SNjakgGk6Vgm0TFwPV1aews_uxyG3WGENXUtg&m=fO0TkOxFlw65Sa8EPsFKFcpHAS14M3cuwEbNKXSWb9c&s=936v1QacYoguKnSpJo4N1ftn4djTxaHIy2uVETI-Osc&e=

jpkrohling: The `InitialDelaySeconds` is already accounting for the 1s delay. There should not be any need for a `TimeoutSeconds` here because of that.

madeelrh:
@jkandasa will do the fix for picking up image names from env. variable.
I will delete the Pull request because I think there are no changes required from our side. Thanks!

PR #1253, opened 2020-10-14
Update instances status using client.Status().update interface

rubenvp8510:

Fixes https://github.com/jaegertracing/jaeger-operator/issues/1242

### DIFF
--- a/pkg/controller/jaeger/jaeger_controller.go
+++ b/pkg/controller/jaeger/jaeger_controller.go
@@ -188,9 +188,10 @@ func (r *ReconcileJaeger) Reconcile(request reconcile.Request) (reconcile.Result
 		}
 	}
 
-	// update the status to "Ready"
-	if instance.Status.Phase != v1.JaegerPhaseRunning {
+	// set the status version to the updated instance version if versions doesn't match
+	if updated.Status.Version != originalInstance.Status.Version || instance.Status.Phase != v1.JaegerPhaseRunning {
 		instance.Status.Phase = v1.JaegerPhaseRunning
+		instance.Status.Version = updated.Status.Version
 		if err := r.client.Status().Update(ctx, instance); err != nil {
 			logFields.WithError(err).Error("failed to store the running status into the current CustomResource")
 			return reconcile.Result{}, tracing.HandleError(err, span)
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -63,8 +63,8 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
 			// nothing to do at this level, just go to the next instance
 			continue
 		}
-		version := jaeger.Status.Version
 		if !reflect.DeepEqual(jaeger, j) {
+			version := jaeger.Status.Version
 			// the CR has changed, store it!
 			if err := c.Update(ctx, &jaeger); err == nil {
 				jaeger.Status.Version = version
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -63,10 +63,19 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
 			// nothing to do at this level, just go to the next instance
 			continue
 		}
-
+		version := jaeger.Status.Version
 		if !reflect.DeepEqual(jaeger, j) {
 			// the CR has changed, store it!
-			if err := c.Update(ctx, &jaeger); err != nil {
+			if err := c.Update(ctx, &jaeger); err == nil {
+				jaeger.Status.Version = version
+				if err := c.Status().Update(ctx, &jaeger); err != nil {
+					log.WithFields(log.Fields{
+						"instance":  jaeger.Name,
+						"namespace": jaeger.Namespace,
+					}).WithError(err).Error("failed update status of the upgraded instance")
+					tracing.HandleError(err, span)
+				}
+			} else {
 				log.WithFields(log.Fields{
 					"instance":  jaeger.Name,
 					"namespace": jaeger.Namespace,
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -57,7 +57,7 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
 			}).Debug("skipping CR upgrade as we are not owners")
 			continue
 		}
-
+		patch := client.MergeFrom(j.DeepCopy())
 		jaeger, err := ManagedInstance(ctx, c, j, latestVersion)
 		if err != nil {
 			// nothing to do at this level, just go to the next instance
@@ -66,9 +66,11 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
 		if !reflect.DeepEqual(jaeger, j) {
 			version := jaeger.Status.Version
 			// the CR has changed, store it!
-			if err := c.Update(ctx, &jaeger); err == nil {
+
+			if err := c.Patch(ctx, &jaeger, patch); err == nil {
+				patch := client.MergeFrom(jaeger.DeepCopy())
 				jaeger.Status.Version = version
-				if err := c.Status().Update(ctx, &jaeger); err != nil {
+				if err := c.Status().Patch(ctx, &jaeger, patch); err != nil {
 					log.WithFields(log.Fields{
 						"instance":  jaeger.Name,
 						"namespace": jaeger.Namespace,
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -74,7 +74,8 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
 					log.WithFields(log.Fields{
 						"instance":  jaeger.Name,
 						"namespace": jaeger.Namespace,
-					}).WithError(err).Error("failed update status of the upgraded instance")
+						"version":   version,
+					}).WithError(err).Error("failed to update status the upgraded instance")
 					tracing.HandleError(err, span)
 				}
 			} else {
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -4,24 +4,44 @@ package e2e
 
 import (
 	"context"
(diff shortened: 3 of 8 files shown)


pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -191,10 +191,15 @@ func (r *ReconcileJaeger) Reconcile(request reconcile.Request) (reconcile.Result
...
 	if instance.Status.Phase != v1.JaegerPhaseRunning {
 		instance.Status.Phase = v1.JaegerPhaseRunning
-		if err := r.client.Status().Update(ctx, instance); err != nil {
-			logFields.WithError(err).Error("failed to store the running status into the current CustomResource")
-			return reconcile.Result{}, tracing.HandleError(err, span)
-		}
+	}
+
+	// set the status version to the updated instance version
+	instance.Status.Version = updated.Status.Version
```
jpkrohling, line 197: I don't think this is the right place to do this... We either do this during the instance creation, or after an upgrade.

pkg/upgrade/upgrade.go
```diff
@@ -72,6 +72,15 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
 					"namespace": jaeger.Namespace,
 				}).WithError(err).Error("failed to store the upgraded instance")
 				tracing.HandleError(err, span)
+			} else {
```
jpkrohling, line 75: It's more readable if you invert the situation: "if err == nil { success } else { failure }". We use "if err != nil" only when that's the only code branch.

Also, you might want to change from Update to Patch.

jpkrohling (commented):
This is missing a test. If you can reproduce the problem with a simple unit test, that's optimal, otherwise, an e2e test is required.

pkg/upgrade/upgrade.go
rubenvp8510: ok, that make sense, about changing Update to Patch I would prefer to do it in a separate PR if it's required.

pkg/controller/jaeger/jaeger_controller.go
rubenvp8510: > This is missing a test. If you can reproduce the problem with a simple unit test, that's optimal, otherwise, an e2e test is required.

I'll add an e2e test, we have unit tests in place for the update process, even with those tests we didn't detect the issue before.

pkg/upgrade/upgrade.go
```diff
@@ -63,10 +63,19 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
...
 			continue
 		}
-
 		if !reflect.DeepEqual(jaeger, j) {
+			version := jaeger.Status.Version
 			// the CR has changed, store it!
-			if err := c.Update(ctx, &jaeger); err != nil {
+			if err := c.Update(ctx, &jaeger); err == nil {
+				jaeger.Status.Version = version
+				if err := c.Status().Update(ctx, &jaeger); err != nil {
```
jpkrohling, line 71: s/Update/Patch/ ?

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"os"
+	"strings"
+	"testing"
+)
+
+const EnvUpdateVersionKey = "UPDATE_TEST_VERSION"
```
jpkrohling, line 18: Those two consts can be part of a single `const ()` declaration.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,61 @@
...
+const upgradeTestTag = "next"
+
+func TestOperatorUpgrade(t *testing.T) {
+
+	upgradeTestVersion := os.Getenv(EnvUpdateVersionKey)
+	t.Log(upgradeTestVersion)
```
jpkrohling, line 24: Add some context to this log entry. Like, `Attempting to upgrade to version ...`

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,61 @@
...
+	fw = framework.Global
+	createdJaeger := &v1.Jaeger{}
+	key := types.NamespacedName{Name: "my-jaeger", Namespace: ctx.GetID()}
+	fw.Client.Get(context.Background(), key, createdJaeger)
+	deployment := &appsv1.Deployment{}
+	fw.Client.Get(context.Background(), types.NamespacedName{Name: "jaeger-operator", Namespace: ctx.GetID()}, deployment)
+	image := deployment.Spec.Template.Spec.Containers[0].Image
+	image = strings.Replace(image, "latest", upgradeTestTag, 1)
+	deployment.Spec.Template.Spec.Containers[0].Image = image
+	fw.Client.Update(context.Background(), deployment)
```
jpkrohling, line 45: s/Update/Patch/?

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"strings"
+	"testing"
+)
+
+const EnvUpdateVersionKey = "UPDATE_TEST_VERSION"
+const upgradeTestTag = "next"
+
+func TestOperatorUpgrade(t *testing.T) {
+
+	upgradeTestVersion := os.Getenv(EnvUpdateVersionKey)
```
jpkrohling, line 23: What happens when the env var is empty?

jpkrohling (commented):
@kevinearls, could you review this one as well?

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,61 @@
...
+		require.FailNow(t, "Failed in prepare")
+	}
+	defer ctx.Cleanup()
+	addToFrameworkSchemeForSmokeTests(t)
+	if err := simplest(t, framework.Global, ctx); err != nil {
+		t.Fatal(err)
+	}
+	fw = framework.Global
+	createdJaeger := &v1.Jaeger{}
+	key := types.NamespacedName{Name: "my-jaeger", Namespace: ctx.GetID()}
```
kevinearls, line 38: @rubenvp8510 I realize this is a short test, but I'd prefer if you followed the example of the other tests, used stretchr suites, and put all of this setup code in a SetupSuite().  That would make it consist with all of the other E2E tests

kevinearls (changes requested):

pkg/upgrade/upgrade.go
rubenvp8510: I can do the change but What would be the benefit of doing patch for this case?

test/e2e/upgrade_test.go
rubenvp8510: It fails :)

jpkrohling: When? With which message? Is it easy to understand what's going on?

rubenvp8510: is not, I'll add a validation here to make it more clear.

test/e2e/upgrade_test.go
rubenvp8510: This log should not be here, was part of my debugging, but yeah a log message could be useful.

pkg/upgrade/upgrade.go
rubenvp8510: Nevermind, I did the change.

pkg/upgrade/upgrade.go
```diff
@@ -57,16 +57,27 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
...
 			// the CR has changed, store it!
-			if err := c.Update(ctx, &jaeger); err != nil {
+
+			if err := c.Patch(ctx, &jaeger, patch); err == nil {
+				patch := client.MergeFrom(jaeger.DeepCopy())
+				jaeger.Status.Version = version
+				if err := c.Status().Patch(ctx, &jaeger, patch); err != nil {
+					log.WithFields(log.Fields{
+						"instance":  jaeger.Name,
+						"namespace": jaeger.Namespace,
```
jpkrohling, line 76:
```suggestion
						"namespace": jaeger.Namespace,
						"version": version,
```

pkg/upgrade/upgrade.go
```diff
@@ -57,16 +57,27 @@ func ManagedInstances(ctx context.Context, c client.Client, reader client.Reader
...
+					}).WithError(err).Error("failed update status of the upgraded instance")
```
jpkrohling, line 77:
```suggestion
					}).WithError(err).Error("failed to update the status of the upgraded instance")
```

jpkrohling (commented):
Looks good to me, but the final review is on @kevinearls :-)

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (suite *OperatorUpgradeTestSuite) TestUpgrade() {
+	upgradeTestVersion := os.Getenv(envUpgradeVersionKey)
+	versionRegexp := regexp.MustCompile(`^[\d]+\.[\d]+\.[\d]+`)
+
+	require.Regexp(t, versionRegexp, upgradeTestVersion,
+		"Invalid upgrade version, need to specify a version to upgrade in format X.Y.Z")
+
+	ctx, err := prepare(t)
```
kevinearls, line 46: @rubenvp8510 A couple of more small things, which are more for consistently with other tests than correctness.

1. Can you move everything from the ""ctx, err := prepare()" down to "fw = framework.Global" into a SetupSuite() method, making sure to remove the "defer ctx.Cleanup() call.
2. Then add a TearDownSuite() which should look like this one: https://github.com/jaegertracing/jaeger-operator/blob/master/test/e2e/elasticsearch_test.go#L56-L58    handleSuiteTearDown will take care of the ctx.Cleanup()

kevinearls (changes requested):

rubenvp8510:
I've already done all requested changes :=) 

Thanks for the review!

kevinearls (approved):

rubenvp8510:
Any other comments on this PR? :)

jpkrohling (approved):

PR #1259, opened 2020-10-14
Allow overriding the vertx example app image and config values

kevinearls:

This will be useful for testing on non-x86 architectures


jpkrohling:
> This will be useful for testing on non-x86 architectures

Wouldn't it be better to then have a multi-arch image? See #1203 for an example on how to build such an image.

PR #1260, opened 2020-10-14
elasticsearch: add SYS_CHROOT capability

haircommander:
many container runtimes give SYS_CHROOT by default.  However, CRI-O does not give pods SYS_CHROOT by default.
Instead of relying on the default of a runtime, the pod description should request the capabilities it needs

Signed-off-by: Peter Hunt <pehunt@redhat.com>

### DIFF
--- a/test/elasticsearch.yml
+++ b/test/elasticsearch.yml
@@ -43,6 +43,9 @@ spec:
             initialDelaySeconds: 5
             periodSeconds: 5
             timeoutSeconds: 4
+          securityContext:
+            capabilities:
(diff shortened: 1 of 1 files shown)


jpkrohling:
Did you mean to add this to the CR used for the e2e tests?

jpkrohling:
I was wondering if you did intend to change the files used by the e2e tests, and apparently you were indeed having problems with the tests, so, your change does look good :-)

jpkrohling (approved):

PR #1262, opened 2020-10-15
Skip detectClusterRoles for Kubernetes

### LINKED ISSUE #1261: Skip detectClusterRoles for Kubernetes
Based on #1258, we should skip the execution of this function if we are *not* running on OpenShift:

https://github.com/jaegertracing/jaeger-operator/blob/619c92481921131da1af876b29b973ac6a6a4cdc/pkg/autodetect/main.go#L199-L221

johanavril:
Fixes #1261

### DIFF
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -196,6 +196,9 @@ func (b *Background) detectKafka(ctx context.Context, apiList *metav1.APIGroupLi
 }
 
 func (b *Background) detectClusterRoles(ctx context.Context) {
+	if viper.GetString("platform") != v1.FlagPlatformOpenShift {
+		return
+	}
 	tr := &authenticationapi.TokenReview{
 		ObjectMeta: metav1.ObjectMeta{Name: "jaeger-operator-TEST"},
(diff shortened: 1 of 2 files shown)


pkg/autodetect/main.go
```diff
@@ -195,7 +195,10 @@ func (b *Background) detectKafka(ctx context.Context, apiList *metav1.APIGroupLi
 	}
 }
 
-func (b *Background) detectClusterRoles(ctx context.Context) {
+func (b *Background) detectClusterRoles(ctx context.Context, apiList *metav1.APIGroupList) {
+	if !isOpenShift(apiList) {
```
jpkrohling, line 199: At this stage, I think you can do something like this instead:

```go
if viper.GetString("platform") != v1.FlagPlatformOpenShift {
 ...
}
```

This way, you don't need the `apiList`.

johanavril: Thanks! That's looks better, I will update later on today.

jpkrohling (approved):

jpkrohling:
Thanks for your PR!

PR #1257, opened 2020-10-14
Add agent hostNetwork option

faceair:
Add the `hostNetwork` option to the agent DaemonSet.

Fixes #676.

PR #1264, opened 2020-10-16
Fix the dependencies

faceair:
Previous discussions https://github.com/jaegertracing/jaeger-operator/pull/1257

### DIFF
--- a/.ci/vgot.sh
+++ b/.ci/vgot.sh
@@ -5,7 +5,7 @@ if [ $# = 0 ]; then
 fi
 d=`mktemp -d`
 cd "$d"
-echo 'module temp' > go.mod
+go mod init temp >/dev/null 2>&1
 for i; do
     pkg=`echo $i | sed 's/@.*//'`
     go get "$i" &&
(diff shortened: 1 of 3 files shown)


.ci/vgot.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+if [ $# = 0 ]; then
+    usage: vgot cmdpackage[@version]... >&2
+    exit 2
+fi
+d=`mktemp -d`
+cd "$d"
+echo 'module temp' > go.mod
```
jpkrohling, line 8: what's the advantage of this vs. `go mod init temp`?

faceair: There is no real difference, if you prefer `go mod init temp` I can change it to this

jpkrohling: I think I would prefer it, yes.

faceair: It should be faster.

jpkrohling (approved):

PR #1267, opened 2020-10-20
Speed up buildx process

morlay:
avoid to run `gobuild` with `qemu` to speed up multi arch build.

added `make dockerx` for local run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,8 @@
 VERSION_DATE ?= $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
-GO_FLAGS ?= GOOS=$(go env GOOS) GOARCH=$(go env GOARCH) CGO_ENABLED=0 GO111MODULE=on
+PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x
+GOARCH ?= $(go env GOARCH)
+GOOS ?= $(go env GOOS)
+GO_FLAGS ?= GOOS=$(GOOS) GOARCH=$(GOARCH) CGO_ENABLED=0 GO111MODULE=on
 KUBERNETES_CONFIG ?= "$(HOME)/.kube/config"
 WATCH_NAMESPACE ?= ""
 BIN_DIR ?= "build/_output/bin"
@@ -75,7 +78,11 @@ gobuild:
 
 .PHONY: docker
 docker:
-	@[ ! -z "$(PIPELINE)" ] || docker build --build-arg=GOPROXY=${GOPROXY} --file build/Dockerfile -t "$(BUILD_IMAGE)" .
+	@[ ! -z "$(PIPELINE)" ] || docker build --build-arg=GOPROXY=${GOPROXY} --build-arg=TARGETARCH=$(GOARCH) --file build/Dockerfile -t "$(BUILD_IMAGE)" .
+
+.PHONY: dockerx
+dockerx:
+	@[ ! -z "$(PIPELINE)" ] || docker buildx build --push --progress=plain --build-arg=GOPROXY=${GOPROXY} --platform=$(PLATFORMS) --file build/Dockerfile -t "$(BUILD_IMAGE)" ${IMAGE_TAGS} .
 
 .PHONY: push
 push:
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,11 +1,23 @@
-FROM golang:1.14 as builder
-
-ARG GOPROXY
-ENV GOPROXY=${GOPROXY}
+FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.14 as builder
 
 COPY . /go/src/github.com/jaegertracing/jaeger-operator/
 WORKDIR /go/src/github.com/jaegertracing/jaeger-operator
-RUN make gobuild
+
+ARG GOPROXY
+# download deps before gobuild
+RUN go mod download -x
+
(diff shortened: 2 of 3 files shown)


build/Dockerfile
```diff
@@ -1,11 +1,29 @@
-FROM golang:1.14 as builder
+# syntax = docker/dockerfile:experimental
```
jpkrohling, line 1: Is this parsed by something?

Makefile
```diff
@@ -77,6 +78,10 @@ gobuild:
 docker:
 	@[ ! -z "$(PIPELINE)" ] || docker build --build-arg=GOPROXY=${GOPROXY} --file build/Dockerfile -t "$(BUILD_IMAGE)" .
 
+.PHONY: dockerx
+dockerx:
+	@[ ! -z "$(PIPELINE)" ] || docker buildx build --progress=plain --build-arg=GOPROXY=${GOPROXY} --platform=linux/arm64,linux/amd64 --file build/Dockerfile -t "$(BUILD_IMAGE)" .
```
jpkrohling, line 83: Shouldn't `./.ci/publish-images.sh` be changed to use this new target?

jpkrohling: Not for this PR, but how hard would it be to add s390? Is it just a matter of adding a new platform, or would it typically imply other changes as well?

build/Dockerfile
morlay: Dockerfile used this to use new features

Makefile
morlay: golang support s390x.
so just added platform linux/s390x.
without this pr

old way will build in qemu, which will be very slow. this pr is for fixing this

morlay: it should, i missing the changes, update it later

morlay:
`make dockerx` (with `linux/s930x` too) results: https://hub.docker.com/r/morlay/jaeger-operator/tags

@jgehrcke i change a little for `make docker` to make e2e testing happy (hope). maybe switch to buildx in future.

build/Dockerfile
morlay: not need for current build. i removed this.

morlay:
previous way means

We build both `linux/amd64` host and`linux/arm64` host (simulated by `qemu`, which will be very slow for compiling.)

current changes means.

We build all on `linux/amd64` host, and compiling to `arm64` and `arm64`.


`qemu` is cool for executing, but is a disater for compiling. the costs show the difference.

image

Makefile
```diff
@@ -75,7 +78,11 @@ gobuild:
 
 .PHONY: docker
 docker:
-	@[ ! -z "$(PIPELINE)" ] || docker build --build-arg=GOPROXY=${GOPROXY} --file build/Dockerfile -t "$(BUILD_IMAGE)" .
+	@[ ! -z "$(PIPELINE)" ] || docker build --build-arg=GOPROXY=${GOPROXY} --build-arg=TARGETARCH=$(GOARCH) --file build/Dockerfile -t "$(BUILD_IMAGE)" .
```
jpkrohling, line 81: Is there anything still using this target? Should they still be using this target? If all users should move to `dockerx`, then we can just replace this target with the new one (ie, the contents of `dockerx` goes into this).

jpkrohling:
@rubenvp8510 if your upgrade e2e test gets merged after this one, you might want to align the building of the container with the changes from this PR.

Makefile
morlay: e2e testing still use this，the Dockerfile could be built by docker directly.

moving to `dockerx` could be next stage.
for testing, `--push` need to be `--load`

jpkrohling (approved):

jpkrohling:
Looks like this has caused the `publish` workflow to fail. Should I revert this last change, or are you able to take a look?

https://github.com/jaegertracing/jaeger-operator/runs/1285761358

morlay:
@jpkrohling 

is docke login success to docker.io and quay.io both?

i only see one `docker login` logs

jpkrohling:
The credentials in the env vars were fine in the previous commit: https://github.com/jaegertracing/jaeger-operator/runs/1281540702

morlay:
@jpkrohling 

could you re run this https://github.com/jaegertracing/jaeger-operator/runs/1285761358?
sometimes may network issues.

jpkrohling:
The message is quite clear: 

> failed to solve: rpc error: code = Unknown desc = server message: insufficient_scope: authorization failed

But I'll trigger it again.

morlay:
@jpkrohling 

quay.io robot account need to set permissions for repo. it that changed?

image
image

I create a quay.io account, got `authorization failed` too without setting permissions.

now pushed https://quay.io/repository/morlay/jaeger-operator?tag=latest&tab=tags

jpkrohling:
It hasn't changed since the last successful commit.

morlay:
Could you check the setting on `quay.io` ?

or re run the last success workflow https://github.com/jaegertracing/jaeger-operator/runs/1281540702.

this pr doesn't change the pushing flow.

morlay:
Final pushing should like this.
image

But now failed on first pushing. 
maybe the logon user to docker hub have no access to `jaegertracing/jaeger-operator`
image

morlay:
@jpkrohling 

is `jaegertracing/jaeger-operator:latest` forbidden to publish for master branch ?

morlay:
here is a additional tag `jaegertracing/jaeger-operator:latest`

which cause the diffrence.

PR #1270, opened 2020-10-21
Publish container - dockerx should not use tag BUILD_IMAGE

morlay:
fix #1267

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ FMT_LOG=fmt.log
 OPERATOR_NAME ?= jaeger-operator
 NAMESPACE ?= "$(USER)"
 BUILD_IMAGE ?= "$(NAMESPACE)/$(OPERATOR_NAME):latest"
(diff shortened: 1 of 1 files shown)


jpkrohling:
Under which tag/name is it available with this change?

morlay:
Nope. 

Final image tags passed in `.ci/publish-images.sh`

```
IMAGE_TAGS=${IMAGE_TAGS} make dockerx
```

May refactor in future when all in dockerx

jpkrohling:
> IMAGE_TAGS=${IMAGE_TAGS} make dockerx

The `IMAGE_TAGS` has then to have a default value, like:

```bash
IMAGE_TAGS ?= "--tag $(BUILD_IMAGE)"
```

So that `make dockerx` won't require IMAGE_TAGS when executed locally.

morlay:
Yes, i will upgrade this

jpkrohling (approved):

PR #1275, opened 2020-10-27
add servicemonitor for operands

### LINKED ISSUE #1156: Create ServiceMonitor for operands
The Jaeger Operator should be creating service monitor objects for operands, so that they can be used out of the box with Prometheus instances managed by the Prometheus Operator.

Context: #1114

sschne:
Closes #1156 
This PR adds servicemonitor used by prometheus-operator for the jaeger operands.

It can currently scrape metrics from the all-in-one jaeger installation, as well as the production and streaming installation, except the agent, as it is deployed as sidecar with the query deployment and does not bring its own Kubernetes Service.

Should I also add e2e-tests for this feature?


jpkrohling:
> Should I also add e2e-tests for this feature?

@kevinearls, what do you think? In general, I would say yes to e2e tests, but more e2e tests means more waiting times for CI results.

kevinearls:
@jpkrohling @sschne I'd say e2e tests are always a good thing.  If CI starts taking too much time we can decide which need to be run on a per PR basis and run others nightly or per release or on some other schedule.

jpkrohling (commented):
Looks great so far, and you even made me realize that we are using a deprecated flag all over our code base (`admin-http-port`). Thanks!

There are a few minor comments, but the direction looks good.

sschne:
> Looks great so far, and you even made me realize that we are using a deprecated flag all over our code base (`admin-http-port`). Thanks!
> 
> There are a few minor comments, but the direction looks good.

Thanks for your review. I have added all your requested changes and also added a e2e-test suite for the servicemonitors.

pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -352,7 +352,12 @@ func (r *ReconcileJaeger) apply(ctx context.Context, jaeger v1.Jaeger, str strat
 	if err := r.applyHorizontalPodAutoscalers(ctx, jaeger, str.HorizontalPodAutoscalers()); err != nil {
 		// we don't want to fail the whole reconciliation when this fails
 		jaeger.Logger().WithError(tracing.HandleError(err, span)).Warn("failed to reconcile pod autoscalers")
-		return jaeger, nil
+	}
+
+	if jaeger.Spec.ServiceMonitor.Enabled != nil && *jaeger.Spec.ServiceMonitor.Enabled {
```
jpkrohling, line 357: I think we might need a safe-guard here. When the service monitor isn't available but the CR sets it as "enabled=true", we see this:

```
WARN[0060] failed to reconcile servicemonitor            error="no matches for kind \"ServiceMonitor\" in version \"monitoring.coreos.com/v1\"" instance=simple-prod namespace=default
```

We do detect already whether the cluster has support for this API:

```
$ make run
...
INFO[0001] Install prometheus-operator in your cluster to create ServiceMonitor objects  error="no ServiceMonitor registered with the API"
```

This is probably done during the bootstrap, so, we could just add to viper whether we should attempt to create service monitors if they are enabled.

Once we have that, it might make sense to set the default value of the "Enabled" flag to `true` in case the Prometheus Operator is available, and to `false` otherwise. This way, the flag is only needed in exceptional cases.

What do you think?

pkg/service/query.go
```diff
@@ -9,6 +9,17 @@ import (
 	"github.com/jaegertracing/jaeger-operator/pkg/util"
 )
 
+// NewQueryAdminService returns a new Kubernetes service for Jaeger Query with admin port enabled
+func NewQueryAdminService(jaeger *v1.Jaeger, selector map[string]string) *corev1.Service {
+	service := NewQueryService(jaeger, selector)
+	service.Spec.Ports = append(service.Spec.Ports,
```
jpkrohling, line 15: Sorry for not having caught this before, but I think we might need an extra service. I'm not sure it's a good idea to add one extra port to the services here, as people might be exposing these services currently and expect only one (or a specific set) of ports to be open.

jpkrohling: The same applies to the other components.

Makefile
```diff
@@ -28,7 +28,7 @@ SDK_VERSION=v0.18.2
 GOPATH ?= "$(HOME)/go"
 GOROOT ?= "$(shell go env GOROOT)"
 
-PROMETHEUS_OPERATOR_TAG ?= v0.39.0
+PROMETHEUS_OPERATOR_TAG ?= v0.43.0
```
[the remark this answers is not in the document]
sschne: The bump was not due to any version requirements, but just to test against an uptodate version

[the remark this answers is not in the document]
jpkrohling: Ok, that's fine.

pkg/controller/jaeger/jaeger_controller.go
sschne: Sounds good. I'm not sure if we should enable the ServiceMonitor by default, since it is changing default behavior, but i would leave this decision up to you.

jpkrohling: @kevinearls, @rubenvp8510, @objectiser, what do you think? One one hand, having two defaults might be confusing. On the other hand, it kinda makes sense to create a service monitor by default if the API for it exists. I think we have a similar behavior with the OAuth Proxy.

kevinearls: @jpkrohling  if by two defaults you mean "If the prometheus operator is installed we create a service monitor by default, otherwise we don't" I think that would be fine, especially if we already have this behavior elsewhere.

objectiser: @jpkrohling I think maybe for 1.x we should keep the same behaviour, but possibly revisit the defaults for 2.x?

objectiser: Sorry misunderstood - yes I think it is ok to create if API exists, as long as we document that as the behaviour

jpkrohling: @sschne, looks like we have a decision. Do you know what's needed to implement this?

sschne: @jpkrohling i see there has been done similar things with Elasticsearch Operator. So i detect whether the API is available via the autodetect package and add it to viper.

jpkrohling: I'd still keep the CR flag in place: people should be able to individually turn it off on an individual basis, perhaps because the instance is a dev/staging one that doesn't need to be monitored.

jpkrohling:
@sschne are you still interested in this one?

sschne:
Hi @jpkrohling , sorry for not looking into this for quite a while. I will try to get this PR back to life and resolve your comments

jpkrohling:
Let me know if you need anything!

pkg/service/query.go
sschne: collector, ingester and query component now create an additional Service with the admin port enabled

pkg/controller/jaeger/jaeger_controller.go
sschne: The CLI flag is now in place with an default auto-detection of the prometheus-operator CRDs.

If the CR field is set to "enable" but the CRDs are not available or the global flag is explicitly disabled, it will issue a warning. This is handled in the jaeger-controller.

The admin services however, which are now dedicated Service objects, will always be created if the CR field is set to "enable", despite the status of the CLI flag.

rubenvp8510:
First pass looks good I'll do another detailed review of this tomorrow morning

frzifus:
hi @sschne, i know its quite a while. But now there are quite a few conflicting files. do you want to update your branch again? :racehorse:

frzifus:
lets reopen this pr when the work continues.

PR #1281, opened 2020-10-28
Use New Admin Port Flag

### LINKED ISSUE #1276: Usage of deprecated 'admin-http-port' flag in the code
While reviewing #1275, I realized that we are using a deprecated flag in quite a few places:

```
$ grep "admin-http-port=" . -r
./pkg/deployment/ingester.go:	adminPort := util.GetPort("--admin-http-port=", args, 14270)
./pkg/deployment/all_in_one.go:	adminPort := util.GetPort("--admin-http-port=", args, 14269)
./pkg/deployment/collector.go:	adminPort := util.GetPort("--admin-http-port=", args, 14269)
./pkg/deployment/agent.go:	adminPort := util.GetPort("--admin-http-port=", args, 14271)
./pkg/deployment/query.go:	adminPort := util.GetPort("--admin-http-port=", args, 16687)
./pkg/inject/sidecar.go:	adminPort := util.GetPort("--admin-http-port=", args, 14271)
```

We should *also* check the new flag, like:

```go
adminPort := 0
if util.FindItem("--admin-http-port=", args) { // legacy flag, to be removed after 1.20.0
  adminPort = util.GetPort("--admin-http-port=", args, 14271) 
} else {
  adminPort = util.GetPort("--admin.http.host-port=", args, 14271) 
}
```

johanavril:
Closes #1276

### DIFF
--- a/pkg/deployment/agent.go
+++ b/pkg/deployment/agent.go
@@ -58,7 +58,7 @@ func (a *Agent) Get() *appsv1.DaemonSet {
 	configRest := util.GetPort("--http-server.host-port=", args, 5778)
 	jgCompactTrft := util.GetPort("--processor.jaeger-compact.server-host-port=", args, 6831)
 	jgBinaryTrft := util.GetPort("--processor.jaeger-binary.server-host-port=", args, 6832)
-	adminPort := util.GetPort("--admin-http-port=", args, 14271)
+	adminPort := util.GetAdminPort(args, 14271)
 
 	trueVar := true
 	falseVar := false
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -353,6 +353,42 @@ func TestGetPortSpecified(t *testing.T) {
 	assert.Equal(t, int32(6831), GetPort("--processor.jaeger-compact.server-host-port=", args, 1234))
 }
 
+func TestGetAdminPort(t *testing.T) {
+	tests := map[string]struct {
+		opts         v1.Options
+		defaultPort  int32
+		expectedPort int32
+	}{
+		"Use default port when no admin port flag provided": {
+			opts:         v1.NewOptions(map[string]interface{}{}),
+			defaultPort:  1234,
+			expectedPort: 1234,
+		},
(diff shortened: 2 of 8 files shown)


pkg/deployment/agent.go
```diff
@@ -58,7 +58,12 @@ func (a *Agent) Get() *appsv1.DaemonSet {
 	configRest := util.GetPort("--http-server.host-port=", args, 5778)
 	jgCompactTrft := util.GetPort("--processor.jaeger-compact.server-host-port=", args, 6831)
 	jgBinaryTrft := util.GetPort("--processor.jaeger-binary.server-host-port=", args, 6832)
-	adminPort := util.GetPort("--admin-http-port=", args, 14271)
+	var adminPort int32
```
jpkrohling, line 61: We are reusing the same code in 6 places: it would probably be a good idea to have a func in the `utils` package. Perhaps something like `AdminPortFromArgs(args) int32`? This should also be accompanied from unit tests.

johanavril: Thanks for the feedback! I made the util function alongside the unit tests.

pkg/util/util_test.go
```diff
@@ -353,6 +353,35 @@ func TestGetPortSpecified(t *testing.T) {
 	assert.Equal(t, int32(6831), GetPort("--processor.jaeger-compact.server-host-port=", args, 1234))
 }
 
+func TestGetAdminPortDefault(t *testing.T) {
```
jpkrohling, line 356: This looks good, but you could have used a table test here, with the args, the default port and the expected outcome as the test's struct:

https://github.com/golang/go/wiki/TableDrivenTests

jpkrohling (commented):
LGTM. If you don't want to use the table tests for the unit tests, let me know and I'll approve this.

pkg/util/util_test.go
johanavril: That looks better, will change to table test now

jpkrohling (approved):

jpkrohling:
Thanks for your contribution!

PR #1290, opened 2020-11-03
Update UI documentation link if is present

rubenvp8510:
Fixes https://github.com/jaegertracing/jaeger-operator/issues/1292

### DIFF
--- a/pkg/strategy/controller.go
+++ b/pkg/strategy/controller.go
@@ -226,13 +226,53 @@ func disableDependenciesTab(uiOpts map[string]interface{}, storage v1.JaegerStor
 	}
 }
 
+func hasDocumentationLink(menus []interface{}) (bool, int) {
+	// Verify if a documentation entry exists.
+	// for now the only way we have to see if a documentation link exists is comparing labels
+	for menuIndex, menuItem := range menus {
+		converted, ok := menuItem.(map[string]interface{})
+		if !ok {
+			return false, 0
+		}
+		if label, ok := converted["label"]; ok && label == "About" {
+			items := converted["items"]
+			convertedItems, ok := items.([]interface{})
+			if !ok {
+				return false, 0
+			}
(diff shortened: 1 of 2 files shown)


pkg/strategy/controller.go
```diff
@@ -226,13 +226,53 @@ func disableDependenciesTab(uiOpts map[string]interface{}, storage v1.JaegerStor
...
+			convertedItems, ok := items.([]interface{})
+			if !ok {
+				return false, 0
+			}
+			for _, item := range convertedItems {
+				itemMap, ok := item.(map[string]interface{})
+				if !ok {
+					return false, 0
+				}
+				if itemLabel, ok := itemMap["label"]; ok && itemLabel == "Documentation" {
```
jpkrohling, line 248: Do we set this initially, or is it only set by users? Can we somehow check whether the value was set by us? We should really avoid changing a value that was set by users.

objectiser: As this issue is related to the doc URL being supplied via CLI flag, can we just provide a fix for that situation (i.e. check if the CLI flag has been defined)?

rubenvp8510: @jpkrohling TBH I don't like this either, unfortunately there is not a way to know if this was set by us or, not that I'm aware of. :( 

@objectiser This code is only executed if the documentation flag is set, See: https://github.com/jaegertracing/jaeger-operator/pull/1290/files#diff-8be2eee66bfcb8efd08e4b1fe2795663a53935ff150187ad426d371534045b01R258

objectiser: @jpkrohling In this situation, if the doc URL is specified via CLI, could we just take that as the true value - so not overridden by user defined value?

objectiser (approved):

jpkrohling:
About the decrease in the code coverage: looks like the error conditions aren't being tested:

image

PR #1301, opened 2020-11-12
Regenerate self-provisioned ES TLS cert when it's outdated

kevinearls:
Force cert regeneration for self prov elasticsearch instances if SAN is invalid

### DIFF
--- a/scripts/cert_generation.sh
+++ b/scripts/cert_generation.sh
@@ -207,6 +207,12 @@ function generate_certs() {
   local component=$1
   local extensions=${2:-}
 
+  # For TRACING-1631 - if we can't find the namespace in the cert it's bad, regenerate everything
+  if [ $REGENERATE_NEEDED = 0 ] && [ "${component}" == "elasticsearch" ] && [ -f ${WORKING_DIR}/logging-es.crt ]  ; then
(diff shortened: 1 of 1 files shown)


scripts/cert_generation.sh
```diff
@@ -207,6 +207,12 @@ function generate_certs() {
   local component=$1
   local extensions=${2:-}
 
+  # For TRACING-1631 - if we can't find the namespace in the cert it's bad, regenerate everything
+  if [ $REGENERATE_NEEDED = 0 ] && [ "${component}" == "elasticsearch" ] && [ -f ${WORKING_DIR}/logging-es.crt ]  ; then
+    openssl x509 -in ${WORKING_DIR}/logging-es.crt -text | grep -q "DNS:elasticsearch.${NAMESPACE}.svc"
```
jpkrohling, line 212: We can merge without this change, but isn't there a command to extract only this field from the cert?

jpkrohling (commented):
This can be merged right away if we need to, but wanted to clarify whether we can extract the SAN explicitly from the cert instead of getting the cert's text and using grep to find something that looks like a SAN field. This improvement can be done later, though.

jpkrohling (approved):

objectiser (approved):

kevinearls:
@jpkrohling Agreed on the SAN.  The problem here is that I could not explicitly extract the SAN with the version of OpenSSL I had on my Mac, even after update, as it did not have the -ext option.

PR #1316, opened 2020-11-25
Enable running tests without an external ES instance

kevinearls:

### DIFF
--- a/test/e2e/elasticsearch_token_propagation_test.go
+++ b/test/e2e/elasticsearch_token_propagation_test.go
@@ -334,6 +334,18 @@ func testAccountToken() string {
 }
 
 func jaegerInstance() *v1.Jaeger {
+	var esResourcesRequest corev1.ResourceList
+	if skipESExternal {
+		esResourcesRequest = corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse(requestMemory),
+			corev1.ResourceCPU:    resource.MustParse(requestCPU),
+		}
+	} else {
+		esResourcesRequest = corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse(requestMemory),
+		}
+	}
+
 	exampleJaeger := &v1.Jaeger{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Jaeger",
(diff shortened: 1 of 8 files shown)


test/e2e/examples2_test.go
```diff
@@ -86,16 +86,25 @@ func (suite *ExamplesTestSuite2) TestServiceTypesExample() {
 }
 
 func (suite *ExamplesTestSuite2) TestSimpleProdWithVolumes() {
+	if skipESExternal {
```
jpkrohling, line 89: Why is this being skipped? Is it be because the example needs an ES to exist at a specific example?

test/e2e/utils.go
```diff
@@ -733,6 +736,18 @@ func waitForElasticSearch() {
 }
 
 func getJaegerSelfProvSimpleProd(instanceName, namespace string, nodeCount int32) *v1.Jaeger {
+	var esResourcesRequest corev1.ResourceList
+	if skipESExternal {
+		esResourcesRequest = corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse(requestMemory),
+			corev1.ResourceCPU:    resource.MustParse(requestCPU),
```
jpkrohling, line 743: Same as the previous comment: why do we need a CPU resource here but not in the `else` branch?

test/e2e/elasticsearch_token_propagation_test.go
```diff
@@ -334,6 +334,18 @@ func testAccountToken() string {
 }
 
 func jaegerInstance() *v1.Jaeger {
+	var esResourcesRequest corev1.ResourceList
+	if skipESExternal {
```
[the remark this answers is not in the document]
kevinearls: @jpkrohling We don't.  I will back this out and submit it as a separate PR

test/e2e/examples2_test.go
kevinearls: Yes, the examples are hard coded to use an external instance.

jpkrohling: Could you add that as a comment to the code?

jpkrohling: I see you did that already :-)

jpkrohling (approved):
LGTM

PR #1319, opened 2020-11-27
Try to resolve container.name from the injected agent args

### LINKED ISSUE #1318: Pod recreation due to container.name tag missing in the jaeger.tags argument
We have observed Pod recreation events in the production cluster very similar to this issue https://github.com/jaegertracing/jaeger-operator/issues/1089

We find the following logs in the operator container,

```
...
time="2020-11-27T01:24:15Z" level=info msg="Injecting Jaeger Agent sidecar" deployment=hera-alertgateway-ack-deploy jaeger=jaeger-ack jaeger-namespace=observability namespace=sqb
time="2020-11-27T01:24:15Z" level=info msg="Injecting Jaeger Agent sidecar" deployment=kubernetes-ack-jack-ack-deploy jaeger=jaeger-ack jaeger-namespace=observability namespace=sqb
time="2020-11-27T01:24:15Z" level=info msg="Injecting Jaeger Agent sidecar" deployment=merchant-dmp-datacenter-ack-deploy jaeger=jaeger-ack jaeger-namespace=observability namespace=sqb
time="2020-11-27T01:24:15Z" level=info msg="Injecting Jaeger Agent sidecar" deployment=merchant-dmp-service-ack-deploy jaeger=jaeger-ack jaeger-namespace=observability namespace=sqb
```

These are the four deployments annotated with `sidecar.jaegertracing.io/inject: 'true'`, and these event repeatedly occur recently. In the meantime, we find the following events from kubernetes, the times exactly coincide.

image

Thus, we compared two replicaSets generated and found the arguments of two agent containers are different,

```
$ diff merchant-dmp-datacenter-ack-deploy-6d5dc79477 merchant-dmp-datacenter-ack-deploy-68b5cc4bd6
...(not related)
120c120
<         - --jaeger.tags=cluster=undefined,deployment.name=merchant-dmp-datacenter-ack-deploy,pod.namespace=sqb,pod.name=${POD_NAME:},host.ip=${HOST_IP:},container.name=merchant-dmp-datacenter-ack
---
>         - --jaeger.tags=cluster=undefined,deployment.name=merchant-dmp-datacenter-ack-deploy,pod.namespace=sqb,pod.name=${POD_NAME:},host.ip=${HOST_IP:}
``` 

I suppose it is related to https://github.com/jaegertracing/jaeger-operator/blob/f250749b7029cfe2c211a07b482e233a65496f47/pkg/inject/sidecar.go#L198-L202 since the agent container has already been injected.

lujiajing1126:
Resolve #1318

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
 	github.com/go-openapi/spec v0.19.8
 	github.com/googleapis/gnostic v0.3.1
-	github.com/miekg/dns v1.1.35 // indirect
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/openshift/api v0.0.0-20200701144905-de5b010b2b38
 	github.com/operator-framework/operator-sdk v0.18.2
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
 	github.com/go-openapi/spec v0.19.8
 	github.com/googleapis/gnostic v0.3.1
+	github.com/miekg/dns v1.1.35 // indirect
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/openshift/api v0.0.0-20200701144905-de5b010b2b38
 	github.com/operator-framework/operator-sdk v0.18.2
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -193,9 +193,11 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment, agentIdx int) corev1.C
 		defaultAgentTagsMap["pod.name"] = fmt.Sprintf("${%s:}", envVarPodName)
 		defaultAgentTagsMap["host.ip"] = fmt.Sprintf("${%s:}", envVarHostIP)
 
-		if (agentIdx == -1 && len(dep.Spec.Template.Spec.Containers) == 1) || // if jaeger-agent does not exist, check if len of containers is one
-			(agentIdx > -1 && len(dep.Spec.Template.Spec.Containers)-1 == 1) { // otherwise check if len of containers is one with jaeger-agent filtered out
-			defaultAgentTagsMap["container.name"] = dep.Spec.Template.Spec.Containers[0].Name
+		defaultContainerName := getContainerName(dep.Spec.Template.Spec.Containers, agentIdx)
+
+		// if we can deduce the container name from the PodSpec
+		if defaultContainerName != "" {
+			defaultAgentTagsMap["container.name"] = defaultContainerName
 		}
 
 		if agentIdx > -1 {
@@ -408,4 +410,17 @@ func joinTags(tags map[string]string) string {
 	}
 	sort.Strings(tagsSlice)
 	return strings.Join(tagsSlice, ",")
+}
+
+func getContainerName(containers []corev1.Container, agentIdx int) string {
+	if agentIdx == -1 && len(containers) == 1 { // we only have one single container and it is not the agent
+		return containers[0].Name
+	} else if agentIdx > -1 && len(containers)-1 == 1 { // we have one single container besides the agent
+		// agent: 0, app: 1
+		// agent: 1, app: 0
+		return containers[1-agentIdx].Name
+	} else {
+		// otherwise, we cannot determine `container.name`
+		return ""
+	}
 }
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -60,10 +60,9 @@ func Sidecar(jaeger *v1.Jaeger, dep *appsv1.Deployment) *appsv1.Deployment {
 	hasAgent, agentContainerIndex := HasJaegerAgent(dep)
 	logFields.Debug("injecting sidecar")
 	if hasAgent { // This is an update
-		dep.Spec.Template.Spec.Containers[agentContainerIndex] = container(jaeger, dep)
+		dep.Spec.Template.Spec.Containers[agentContainerIndex] = container(jaeger, dep, agentContainerIndex)
 	} else {
-		dep.Spec.Template.Spec.Containers = append(dep.Spec.Template.Spec.Containers, container(jaeger, dep))
-
+		dep.Spec.Template.Spec.Containers = append(dep.Spec.Template.Spec.Containers, container(jaeger, dep, -1))
 	}
 
 	jaegerName := util.Truncate(jaeger.Name, 63)
@@ -164,7 +163,7 @@ func getJaeger(name string, jaegers *v1.JaegerList) *v1.Jaeger {
 	return nil
 }
 
-func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
+func container(jaeger *v1.Jaeger, dep *appsv1.Deployment, agentIdx int) corev1.Container {
 	args := append(jaeger.Spec.Agent.Options.ToArgs())
 
 	// we only add the grpc host if we are adding the reporter type and there's no explicit value yet
@@ -199,6 +198,15 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 			agentTags = fmt.Sprintf("%s,%s=%s", agentTags,
 				"container.name", dep.Spec.Template.Spec.Containers[0].Name,
 			)
+		} else if agentIdx > -1 {
+			agentContainer := dep.Spec.Template.Spec.Containers[agentIdx]
+			// try to resolve the tag with key `container.name` in the existing Args of the agent container
+			existingContainerNamePair := util.FindItem("container.name=", parseAgentTags(agentContainer.Args))
+			if len(existingContainerNamePair) > 0 {
+				agentTags = fmt.Sprintf("%s,%s", agentTags,
+					existingContainerNamePair,
+				)
+			}
 		}
 
 		args = append(args, fmt.Sprintf(`--jaeger.tags=%s`, agentTags))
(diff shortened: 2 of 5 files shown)


pkg/inject/sidecar.go
```diff
@@ -199,6 +198,15 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
 			agentTags = fmt.Sprintf("%s,%s=%s", agentTags,
 				"container.name", dep.Spec.Template.Spec.Containers[0].Name,
 			)
+		} else if agentIdx > -1 {
```
jpkrohling, line 201: I think the original condition isn't accurate anymore and needs to be fixed. We should rather get a list of containers with the agent filtered out, and then check the length. The case here is that we might have more containers in the pod besides the Jaeger Agent, and we wouldn't know which value to use as `container.name`.

That said, I think this whole conditional deserves a refactoring already. It should place all the tags as a `map[string]string` and join them once the map is complete. This way, we add the `container.name` to the map in only one place.

pkg/inject/sidecar.go
```diff
@@ -378,3 +386,12 @@ func EqualSidecar(dep, oldDep *appsv1.Deployment) bool {
 	oldDepContainer := oldDep.Spec.Template.Spec.Containers[oldDepIndex]
 	return reflect.DeepEqual(depContainer, oldDepContainer)
 }
+
+func parseAgentTags(args []string) []string {
```
jpkrohling, line 390: If we follow the refactoring mentioned above, we might not need this. The reason is that we'd have a consistent result for the map, as long as the container name doesn't change. Once it changes, we want this new name to be used in the map/args.

jpkrohling (commented):
cc @rubenvp8510

pkg/inject/sidecar.go
lujiajing1126: I partially agree for this point. Suppose we have a deployment with one single container at the very beginning, of course we can readily deduce the `container.name`. Then we inject several sidecars base on the annotations, let's say, istio-proxy and jaeger-agent, we will finally have three containers in this deployment.

At the second time, even if we are able to filter the jaeger-agent out, we still have two containers which we are not able to pick the correct `container.name` without a prior. So I think it is still necessary to parse the Tags / Args from the existing agent container.

lujiajing1126:
I try to refactor the code following the comment made by @jpkrohling 

- A default map of tags is created
- Existing tags are parsed if jaeger-agent is already injected
- The old map is overwritten by new generated default values
- The tags are now sorted in alphabetic order to keep it stable

pkg/inject/sidecar.go
jpkrohling: We should really leave the `container.name` out if we have more than one container (not counting the Jaeger Agent one). It's certainly possible that traces are coming from both containers in that case, so, the annotation will be wrong. Reusing the container name from an older version of the pod doesn't seem right to me.

lujiajing1126: > We should really leave the `container.name` out if we have more than one container (not counting the Jaeger Agent one). It's certainly possible that traces are coming from both containers in that case, so, the annotation will be wrong. Reusing the container name from an older version of the pod doesn't seem right to me.

I understand your argument and I think it is quite reasonable. But the deployment may still probably restart once if we inject more than one sidecar. That would cause a serious problem in a large cluster (we have ~500 deployments in our production cluster). How can we prevent from this situation?

pkg/inject/sidecar_test.go
```diff
@@ -523,8 +525,8 @@ func TestSidecarExplicitTags(t *testing.T) {
 
 	// verify
 	assert.Len(t, dep.Spec.Template.Spec.Containers, 2)
-	agentTags := agentTags(dep.Spec.Template.Spec.Containers[1].Args)
-	assert.Equal(t, []string{"key=val"}, agentTags)
+	agentTags := parseAgentTags(dep.Spec.Template.Spec.Containers[1].Args)
+	assert.True(t, reflect.DeepEqual(agentTags, map[string]string{"key": "val"}))
```
jpkrohling, line 529: Why can't this be `assert.Equal` as it was before?

pkg/inject/sidecar.go
jpkrohling: This will happen only if you didn't have another container before, but now you have, right? The only way to prevent that, from what I can see, is to add the agent sidecar manually, so that the operator won't manage this container for you.

pkg/inject/sidecar.go
```diff
@@ -187,21 +186,29 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment) corev1.Container {
...
+		defaultAgentTagsMap := make(map[string]string)
+		defaultAgentTagsMap["cluster"] = "undefined" // this value isn't currently available
+		defaultAgentTagsMap["deployment.name"] = dep.Name
+		defaultAgentTagsMap["pod.namespace"] = dep.Namespace
+		defaultAgentTagsMap["pod.name"] = fmt.Sprintf("${%s:}", envVarPodName)
+		defaultAgentTagsMap["host.ip"] = fmt.Sprintf("${%s:}", envVarHostIP)
+
+		if (agentIdx == -1 && len(dep.Spec.Template.Spec.Containers) == 1) || // if jaeger-agent does not exist, check if len of containers is one
+			(agentIdx > -1 && len(dep.Spec.Template.Spec.Containers)-1 == 1) { // otherwise check if len of containers is one with jaeger-agent filtered out
+			defaultAgentTagsMap["container.name"] = dep.Spec.Template.Spec.Containers[0].Name
```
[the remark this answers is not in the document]
lujiajing1126: I have extracted the logic to `getContainerName` and fixed the problem

pkg/inject/sidecar.go
```diff
@@ -378,3 +385,27 @@ func EqualSidecar(dep, oldDep *appsv1.Deployment) bool {
 	oldDepContainer := oldDep.Spec.Template.Spec.Containers[oldDepIndex]
 	return reflect.DeepEqual(depContainer, oldDepContainer)
 }
+
+func parseAgentTags(args []string) map[string]string {
+	tagsArg := util.FindItem("--jaeger.tags=", args)
+	if tagsArg == "" {
+		return map[string]string{}
```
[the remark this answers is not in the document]
lujiajing1126: Changed

pkg/inject/sidecar_test.go
lujiajing1126: Changed

go.mod
```diff
@@ -7,7 +7,6 @@ require (
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
 	github.com/go-openapi/spec v0.19.8
 	github.com/googleapis/gnostic v0.3.1
-	github.com/miekg/dns v1.1.35 // indirect
```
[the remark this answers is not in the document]
lujiajing1126: I have reverted all changes to `go.mod` and `go.sum`. But may be this is an indirect dependency, so it will be cleaned by `go mod tidy`?

pkg/inject/sidecar.go
lujiajing1126: Specifically, I am concerned with the scenario with two operators: one is `jaeger-operator`, the other is `istio-operator`. 

Emm, what about adding an annotation to specify the container-name, if the user want to set the name explicitly.

pkg/inject/sidecar.go
```diff
@@ -409,3 +411,16 @@ func joinTags(tags map[string]string) string {
 	sort.Strings(tagsSlice)
 	return strings.Join(tagsSlice, ",")
 }
+
+func getContainerName(containers []corev1.Container, agentIdx int) string {
```
jpkrohling, line 415: Now that it's factored out, it should now be easy to add tests for all conditions here ;-)

lujiajing1126: Tests added

go.mod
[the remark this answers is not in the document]
jpkrohling: Might be. Once this gets merged, would you like to run a `go mod tidy` against master and seeing if there are differences in the go.mod?

pkg/inject/sidecar.go
jpkrohling: Users can do that already by adding env vars to their own container (JAEGER_TAGS). The tags we add to the sidecar container are ones that we add based on information we can infer from the env.

pkg/inject/sidecar_test.go
```diff
@@ -508,8 +508,9 @@ func TestSidecarOrderOfArguments(t *testing.T) {
 	containsOptionWithPrefix(t, dep.Spec.Template.Spec.Containers[1].Args, "--c-option")
 	containsOptionWithPrefix(t, dep.Spec.Template.Spec.Containers[1].Args, "--jaeger.tags")
 	containsOptionWithPrefix(t, dep.Spec.Template.Spec.Containers[1].Args, "--reporter.grpc.host-port")
-	agentTags := agentTags(dep.Spec.Template.Spec.Containers[1].Args)
-	assert.Contains(t, agentTags, "container.name=only_container")
+	agentTagsMap := parseAgentTags(dep.Spec.Template.Spec.Containers[1].Args)
+	assert.Contains(t, agentTagsMap, "container.name")
+	assert.Equal(t, agentTagsMap["container.name"], "only_container")
```
jpkrohling, line 513: If you are changing this PR for some other reason, you might want to remove the `Contains`, as the `Equals` assertion already covers it.

jpkrohling (commented):
LGTM. @rubenvp8510, please take a look as well. Once you give your approval, I'll merge this.

go.mod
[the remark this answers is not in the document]
lujiajing1126: Sure. I would like to have a try.

pkg/inject/sidecar_test.go
lujiajing1126: Yes, you're right. Unnecessary `Contains` Removed. Also fix the same issues in other places.

lujiajing1126:
Another problem may possibly come from the different logic in `deployment_controller` and `namespace_controller`, in the latter case we do not check whether the sidecar has already been injected.

But in the `deployment_controller`, we check that and exclude the container from the existing container list.

IMO, it will also cause unstable status.

jpkrohling:
Great catch. Would you like to change this PR here to align the logic?

lujiajing1126:
> Great catch. Would you like to change this PR here to align the logic?

Sure. And I prefer not to exclude the agent from the container list since we can already reach convergence even if we inject the agent twice. What's your opionion?

jpkrohling:
I'm not quite sure I understand what you mean, but send the code and I'll review it.

lujiajing1126:
> I'm not quite sure I understand what you mean, but send the code and I'll review it.

I have aligned `deployment_controller` with `namespace_controller`

rubenvp8510:
This looks good to me. Good refactoring!

rubenvp8510 (approved):

PR #1323, opened 2020-11-30
Make ServiceMonitor creation optional

igorwwwwwwwwwwwwwwwwwwww:
This patch introduces a new `-service-monitor-enabled` flag for the `start` command, defaulting to true. It allows turning off the creation of `ServiceMonitor` objects by the operator.

The use case we have for this is propagating more labels on the ServiceMonitor. The default one that is created with `github.com/operator-framework/operator-sdk/pkg/metrics.CreateServiceMonitors` does not have the ability for customization on labels.

Our monitoring highly relies on having certain labels on all metrics. So in this case it's easier for us to disable the existing ServiceMonitor and provide our own.

This patch has not yet been tested -- I wanted to first gauge whether you'd be willing to accept something like this.

See also [our downstream issue](https://gitlab.com/gitlab-com/gl-infra/k8s-workloads/tanka-deployments/-/merge_requests/58).


jpkrohling:
Sounds reasonable to me!

siku4:
I also welcome this patch! We have the same problem as described by @igorwwwwwwwwwwwwwwwwwwww. We are forced to deploy a second ServiceMonitor manually so that we are able to define additional labels.

At the latest if you manage multiple Prometheus instances with the Prometheus operator you *must* define a label to in your Pod/ServiceMonitor to select the target Prometheus.

jpkrohling:
Not sure why and how, but I lost track of this PR. I just rebased and would certainly consider merging it if it's still up to date.

jpkrohling (approved):

PR #1327, opened 2020-12-02
Fix unstable mod tidy

### LINKED ISSUE #1326: Unable to lock version for indirect dependency
As we have discussed in PR #1319, I tried to run `go mod tidy` on the master branch and the indirect version of `github.com/miekg/dns@v1.1.35` went away. This dependency is introduced in #1298 intended to fix a CVE issue.

I suppose as per https://github.com/golang/go/issues/40784#issuecomment-675161073, it is not possible to lock/pin an indirect dependency. So we have to be very careful with this indirect dep.

Just a comment. I am not sure what we can do now. Maybe add a git-hook to ensure? @jpkrohling

lujiajing1126:
Fix #1326 

Signed-off-by: Megrez Lu <lujiajing1126@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
 	github.com/go-openapi/spec v0.19.8
 	github.com/googleapis/gnostic v0.3.1
-	github.com/miekg/dns v1.1.35 // indirect
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/openshift/api v0.0.0-20200701144905-de5b010b2b38
 	github.com/operator-framework/operator-sdk v0.18.2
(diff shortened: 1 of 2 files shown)


jpkrohling (commented):
Looks good, but can you find a way to provide a positive confirmation that this is indeed the version being included? Running `go mod graph | grep dns` shows all versions that our dependencies use for this, so, it wouldn't help much I think...

lujiajing1126:
> Looks good, but can you find a way to provide a positive confirmation that this is indeed the version being included? Running `go mod graph | grep dns` shows all versions that our dependencies use for this, so, it wouldn't help much I think...

After some investigation, I find that "github.com/miekg/dns" module does not exist in the binary built with `make gobuild` at all. So I think it is safe to run `go mod tidy` to make a cleanup.

I have tested with both `objdump` and `go tool objdump`.

jpkrohling:
Sounds good to me. We might want to get an approval from @objectiser as well, as it was introduced by him some time ago.

objectiser:
The second commit, where the dependency is listed in the `replace` section, seemed ok - but without that it reverts the version of the dependency which is affected by the CVE. What was the issue with the second commit?

lujiajing1126:
> The second commit, where the dependency is listed in the `replace` section, seemed ok - but without that it reverts the version of the dependency which is affected by the CVE. What was the issue with the second commit?

The affected module "github.com/miekg/dns" seems not appear in the generated binary. So I suppose it is safe to run `go mod tidy` here.

objectiser (approved):

PR #1329, opened 2020-12-03
Upgrade minikube, kubectl, ubuntu, and go versions

kevinearls:

### DIFF
--- a/.github/workflows/e2e-kubernetes.yaml
+++ b/.github/workflows/e2e-kubernetes.yaml
@@ -15,6 +15,7 @@ jobs:
       with:
         minikube-version: v1.15.1
     - uses: jpkrohling/setup-kubectl@v1.0.2
+    - uses: jpkrohling/setup-operator-sdk@v1.0.2
       with:
         operator-sdk-version: v0.18.2
     - uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-kubernetes.yaml
```diff
@@ -3,17 +3,17 @@ on: [push, pull_request]
 
 jobs:
   end-to-end:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
```
jpkrohling, line 6: There's 20.04 available as well, IIRC.

kevinearls: @jpkrohling I did try 20.04 and ran into the same problems.  This was mostly just an experiment to see if I could just change versions and have everything magically work.  Unfortunately that was not the case, and it looks like this upgrade will require at least a few more config changes.  I will probably pause this and get back to it when I have a few free hours.

jpkrohling: If you get things working in minikube, you can probably get it working with kind, and I know kind does work on ubuntu-20.04. I think Linkerd uses it, and even had a KubeCon talk on the subject :)

kevinearls:
This is now seriously out of date.

PR #1333, opened 2020-12-08
Copy common spec to avoid touching persisted CR spec

rubenvp8510:


objectiser:
@rubenvp8510 could you add a test to confirm that the supplied Jaeger CR is not modified - and that the returned container has the volume/mounts?

objectiser:
@rubenvp8510 Could you add
```
	assert.Len(t, jaeger.Spec.Agent.VolumeMounts, 0)
	assert.Len(t, jaeger.Spec.Agent.Volumes, 0)
```
to https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/inject/sidecar_test.go#L52 and again at the end of the test - i.e. to check there are no volumes/mounts in the initial Jaeger CR and after the sidecar has been injected.

Then I think it is ready to merge.

objectiser (approved):

Issue #1315, opened 2020-11-25
Existing apps with sidecar agent inject annotation not injected when first appropriate jaeger instance created


objectiser:
If an app has been deployed, before the Jaeger instance is created, the operator log shows:
```
"No suitable Jaeger instances found to inject a sidecar" deployment=myapp
```
However, when the Jaeger instance is created (and available), the operator does not rescan the apps to see whether an agent for that Jaeger instance can now be injected.

Reproduced on OCP4.5 with Jaeger 1.17 and 1.20.

objectiser:
cc @jpkrohling @rubenvp8510 

Haven't tried on latest, as was not available in the operatorhub at the time, so might be fixed?

Also tried with app having injected sidecar, then removed the jaeger instance (which cleaned up the sidecar agent), and then created a new jaeger instance - and the agent doesn't get re-injected into the app.

BTW - the Jaeger instance and business app are deployed to different namespaces.

jpkrohling:
There were quite a few changes made by @rubenvp8510 in the past on this logic, so, this might indeed be fixed already. But would be nice to get a confirmation from him.

rubenvp8510:
I changed the logic to address the case of injection across namespaces. and IIRC I put some logic to scan apps for injection
I'll check this case and see if can reproduce it.

PR #1334, opened 2020-12-09
Trigger deployments reconciliation when jaeger instance is created

rubenvp8510:

Fixes https://github.com/jaegertracing/jaeger-operator/issues/1315

### DIFF
--- a/pkg/controller/deployment/deployment_controller.go
+++ b/pkg/controller/deployment/deployment_controller.go
@@ -51,27 +51,9 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	err = c.Watch(&source.Kind{Type: &v1.Jaeger{}}, &handler.EnqueueRequestsFromMapFunc{
-		ToRequests: handler.ToRequestsFunc(func(object handler.MapObject) []reconcile.Request {
-			deployments := appsv1.DeploymentList{}
-			err := mgr.GetClient().List(context.Background(), &deployments)
-			reconciliations := []reconcile.Request{ }
-			if err != nil {
-				return reconciliations
-			}
-			for _, dep := range deployments.Items {
-				reconciliations = append(reconciliations, reconcile.Request{
-					NamespacedName: types.NamespacedName{
-						Name:      dep.Name,
-						Namespace: dep.Namespace,
-					},
-				})
-
-			}
-			return reconciliations
-		}),
+	return c.Watch(&source.Kind{Type: &v1.Jaeger{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(r.(*ReconcileDeployment).syncOnJaegerChanges),
 	})
-	return nil
 }
 
 var _ reconcile.Reconciler = &ReconcileDeployment{}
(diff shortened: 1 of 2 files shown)


rubenvp8510:
@jpkrohling  @objectiser The changes looks really good. 

Regarding of codecov errors, it seems those are due uncovered lines when the client returns errors and one with assert error. I think those  are hard to cover/test. 

Is this good enough to be merged as it is?

pkg/controller/deployment/deployment_controller_test.go
```diff
@@ -0,0 +1,133 @@
...
+				Annotations: map[string]string{
+					inject.Annotation: "true",
+				},
+			},
+		},
+		&appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "dep-with-another-jaegers-label",
+				Namespace: "ns-without-annotation",
+				Labels: map[string]string{
```
objectiser, line 73: Should the inject annotation be included here, as in the deployment below - but as it is using a different Jaeger in the label it wouldn't be returned?

rubenvp8510: Well, may be yes to be consistent with the expected states of a deployment, as  I cannot imagine an scenario where we have the label and not the annotation.

On the other side, for what I this is testing it doesn't matter.

https://github.com/jaegertracing/jaeger-operator/pull/1334/files#diff-c553184997fa4c9f63e697af2f39eb2e6fb43eeff64a0ef66c9e955f61347b02R205

Whether this has or not an annotation, the first thing the `syncOnJaegerChanges` checks, is for the label, so it doesn't matter if we don't have an annotation.

pkg/controller/deployment/deployment_controller.go
```diff
@@ -178,3 +180,54 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
...
+	reconciliations := []reconcile.Request{}
+	nss := map[string]corev1.Namespace{} // namespace cache
+
+	jaeger, ok := event.Object.(*v1.Jaeger)
+	if !ok {
+		return reconciliations
+	}
+
+	deployments := appsv1.DeploymentList{}
+	err := r.client.List(context.Background(), &deployments)
```
rubenvp8510, line 194: Question

In the case of the operator watching one or a couple of namespaces (no cluster wide) Should we limit this to those namespaces?

objectiser: I don't think we have a way to restrict the namespaces being observed for injection? the WATCH_NAMESPACE part is which namespaces to check for CRs, not deployments.
@jpkrohling is that also your understanding?

pkg/controller/deployment/deployment_controller_test.go
objectiser: Agree, it was more to reflect a realistic deployment to avoid any failures later if the logic was to change.

pkg/controller/deployment/deployment_controller.go
rubenvp8510: Ok, I thought WATCH_NAMESPACE restricts injection too. My bad.

pkg/controller/deployment/deployment_controller_test.go
rubenvp8510: In that case. Yes , it should be added :)

pkg/controller/deployment/deployment_controller.go
jpkrohling: I'd expect the client.List to return the deployments for all namespaces where the jaeger-agent service account has access to. Perhaps QE could confirm that.

jpkrohling:
> Regarding of codecov errors, it seems those are due uncovered lines when the client returns errors and one with assert error

I think it's actually because there's a new test, which covers only one function, the new one. When there was 0% coverage, this package wasn't being counted at all.

> Is this good enough to be merged as it is?

Was it confirmed already that this is indeed the fix for the original problem?

pkg/controller/deployment/deployment_controller_test.go
jpkrohling: > Should the inject annotation be included here, as in the deployment below - but as it is using a different Jaeger in the label it wouldn't be returned?

I left only the `Label` here as that's the only thing the code for this case will use. If we think we need to check both the annotation and the label, then both the test and the code need to be changed.

objectiser:
@jpkrohling It fixes one of the problems - and I believe the other PR (removing the persisted volume/mounts from the CR) addresses the other issue described, although we haven't been able to reproduce that.

rubenvp8510:
This also fixes an upstream issue, I changed the description of this PR to reflect that.

pkg/controller/deployment/deployment_controller_test.go
rubenvp8510: I don't think we need to check both, Is more that the case should reflect a real deployment. 

I understand that this test is not using the annotation, for this particular case it only check for the label. But if in the future we change this logic it could help to have the deployment as close as real as possible.

PR #1361, opened 2021-01-13
Skipping Kafka related tests in examples2 when Kafka is not available

madeelrh:
On some platform, AMQ streaming is not available. With this patch we skip Kafka related tests and execute rest of them in examples2 test suite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -288,12 +288,16 @@ endif
 
 .PHONY: kafka
 kafka: deploy-kafka-operator
+ifeq ($(WAIT_FOR_KAFKA),false)
+	@echo Skipping Kafka/external ES related tests
+else
 	@echo Creating namespace $(KAFKA_NAMESPACE)
 	@kubectl create namespace $(KAFKA_NAMESPACE) 2>&1 | grep -v "already exists" || true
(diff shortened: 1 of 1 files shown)


jkandasa:
@objectiser @kevinearls can you please review this PR?

kevinearls (changes requested):
Hi @madeelrh  Thanks for the PR.  Sorry to nit-pick, but rather than use WAIT_FOR_KAFKA I would suggest using SKIP_KAFKA or something like that.

WAIT_FOR_KAFKA sounds like what we do in internal CI where we wait for the Kafka cluster to finish deploying, whereas here we are not going to deploy it at all.

madeelrh:
Hi @kevinearls, thanks for your comments. WAIT_FOR_KAFKA is used intentionally. Because SKIP_KAFKA would add a new job-param to the CI which already has many :) 
The logic here is if you check WAIT_FOR_KAFKA in CI, it will work for x86 and if left uncheck it will work for other platforms.(off course after this PR is merged)

If e2e tests are running manually, you also won't have to set env WAIT_FOR_KAFKA. We will have to set it to false for our platform. Thanks.

PR #1379, opened 2021-02-04
Remove support for the experimental OpenTelemetry-based Jaeger

jpkrohling:
This commit removes the support for the experimental OpenTelemetry-based Jaeger that is part of Jaeger v1, as additional binaries/images. Jaeger v2 will be completely based on OpenTelemetry and we'll have the Jaeger Operator v2 to handle this.


kevinearls (approved):

examples/simplest.yaml
```diff
@@ -2,3 +2,8 @@ apiVersion: jaegertracing.io/v1
 kind: Jaeger
 metadata:
   name: simplest
+spec:
+  allInOne:
+    image: jaegertracing/all-in-one:latest
+  agent:
+    image: jaegertracing/jaeger-agent:latest
```
jkandasa, line 9: @jpkrohling Is there any specific reason to include image tags on this template? 
Some tests are failing in IBM, since jaegertracing/all-in-one:latest do not have P and Z arch images.

jpkrohling: What should be used instead?

jkandasa: @jpkrohling we can leave it blank. Jaeger operator will update (based on the operator version)
If it is ok, I can create PR to remove changes on this file.

jpkrohling: Those are example files, they can't be left blank. Can't they adapt this before running the test?

PR #1380, opened 2021-02-04
Migrate jaeger.tags in existing CRs

jpkrohling:
This commit adds a migration for existing CRs, moving from the deprecated jaeger.tags to collector.tags or agent.tags, according to the components being used.


pkg/upgrade/v1_22_0_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func TestUpgradeJaegerTagssv1_22_0(t *testing.T) {
+	latestVersion := "1.22.0"
+	opts := v1.NewOptions(map[string]interface{}{
+		"jaeger.tags": "somekey=somevalue",
+	})
+
+	nsn := types.NamespacedName{Name: "my-instance"}
+	existing := v1.NewJaeger(nsn)
+	existing.Status.Version = "1.19.0"
```
objectiser, line 24: nit: Probably best to have as most recent version (e.g. `1.21.0`) so does not trigger any other upgrade code.

objectiser (commented):
LGTM - just one minor comment, but if you want to leave it, I will approve as is.

pkg/upgrade/v1_22_0_test.go
jpkrohling: Done!

objectiser (approved):

Issue #1307, opened 2020-11-18
Allow specific values for nodeport to be configured


CSP197:
As a user of jaeger-operator, I would like to expose the jaeger-query service in my K8s VPC. 

As per #1263, I understand that a service type can be set via -
```
query:
      serviceType: NodePort
```
But it would be nice if a user could specify a constant value for the exposed nodeport as follows -
```
query:
      serviceType: NodePort
      nodePort: 3####
```

Thanks all,

dogukanturan:
I guess we can't add NodePort and we can't make a static ClusterIP definition either.

CSP197:
@jpkrohling - Can you pls review my PR: https://github.com/jaegertracing/jaeger-operator/pull/1394?

jpkrohling:
@rubenvp8510 will take a look at this one

PR #1391, opened 2021-02-15
#1307: Adding support for NodePort in Jaeger Query Service

CSP197:
Adding fix to allow for user defined NodePort value to be used as a part of deployed Jaeger Query Service.


CSP197:
Closing this PR in hopes for fixing DCO error and will open a new one later...

PR #1392, opened 2021-02-15
Add ability to indicate PriorityClass for agent

elkh510:


jpkrohling:
Please update the generated files:

> Build failed: a model has been changed but the generated resources aren't up to date. Run 'make generate' and update your PR.

Also, take a look at the DCO failure. Other than that, looks good!

elkh510:
@jpkrohling done
review pls)

jpkrohling (commented):
Looks good, but I just realized that this is missing some documentation. Please, adapt an existing example to include this new option, and consider adding it to the main documentation as well: https://github.com/jaegertracing/documentation/blob/master/content/docs/next-release/operator.md

elkh510:
@jpkrohling added [example](https://github.com/jaegertracing/jaeger-operator/blob/1dfa2730273f31b8bbe556b842b677c6871e989c/examples/agent-with-priority-class.yaml)
pls review

examples/agent-with-priority-class.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: jaegertracing.io/v1
+kind: Jaeger
+metadata:
+  name: agent-as-daemonset
+spec:
+  agent:
+    strategy: DaemonSet
+    PriorityClassName: test # priorityClassName here
```
jpkrohling, line 8:
```suggestion
    priorityClassName: test # priorityClassName here
```

This example should include an example of a priority class, like (from the documentation):

```yaml
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: high-priority
value: 1000000
globalDefault: false
description: "This priority class should be used for XYZ service pods only."
```

If you use this in the example, don't forget to change the name to `high-priority` in the Jaeger CR.

elkh510: yes, sorry.
added changes as above
@jpkrohling pls review

jpkrohling (approved):
Thanks!

PR #1394, opened 2021-02-17
Add support for NodePort in Jaeger Query Service

### LINKED ISSUE #1307: Allow specific values for nodeport to be configured

CSP197:
This PR should allow users to expose Jaeger's Query service as a NodePort at a specific port. Earlier, the Jaeger Operator did not support specifying a port value and would depend on K8s to randomly select an apt port value. This should resolve this blocker.

Resolves #1307

### DIFF
--- a/pkg/service/query.go
+++ b/pkg/service/query.go
@@ -18,6 +18,17 @@ func NewQueryService(jaeger *v1.Jaeger, selector map[string]string) *corev1.Serv
 		annotations["service.alpha.openshift.io/serving-cert-secret-name"] = GetTLSSecretNameForQueryService(jaeger)
 	}
 
+	ports := []corev1.ServicePort{
+		{
+			Name:       getPortNameForQueryService(jaeger),
+			Port:       int32(GetPortForQueryService(jaeger)),
+			TargetPort: intstr.FromInt(getTargetPortForQueryService(jaeger)),
+		},
+	}
+	if jaeger.Spec.Query.ServiceType == corev1.ServiceTypeNodePort {
+		ports[0].NodePort = int32(GetNodePortForQueryService(jaeger))
+	}
+
 	return &corev1.Service{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Service",
@@ -41,13 +52,7 @@ func NewQueryService(jaeger *v1.Jaeger, selector map[string]string) *corev1.Serv
 		Spec: corev1.ServiceSpec{
 			Selector: selector,
 			Type:     getTypeForQueryService(jaeger),
-			Ports: []corev1.ServicePort{
-				{
-					Name:       getPortNameForQueryService(jaeger),
-					Port:       int32(GetPortForQueryService(jaeger)),
-					TargetPort: intstr.FromInt(getTargetPortForQueryService(jaeger)),
-				},
-			},
+			Ports:    ports,
 		},
 	}
 }
@@ -89,4 +94,9 @@ func getTypeForQueryService(jaeger *v1.Jaeger) corev1.ServiceType {
 		return jaeger.Spec.Query.ServiceType
 	}
 	return corev1.ServiceTypeClusterIP
+}
+
+// GetNodePortForQueryService returns the query service NodePort for this Jaeger instance
+func GetNodePortForQueryService(jaeger *v1.Jaeger) int {
+	return int(jaeger.Spec.Query.NodePort)
 }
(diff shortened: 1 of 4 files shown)


jpkrohling:
@rubenvp8510 , would you please review this one?

pkg/service/query.go
```diff
@@ -90,3 +95,8 @@ func getTypeForQueryService(jaeger *v1.Jaeger) corev1.ServiceType {
 	}
 	return corev1.ServiceTypeClusterIP
 }
+
+// GetNodePortForQueryService returns the query service NodePort for this Jaeger instance
+func GetNodePortForQueryService(jaeger *v1.Jaeger) int {
```
rubenvp8510, line 100: Why this function is returning int instead of int32. When you use it you cast it again to int32
I would suggest to change this to:
```
func GetNodePortForQueryService(jaeger *v1.Jaeger) int32 {
	return jaeger.Spec.Query.NodePort
}
```
And ged rid of all `int32(GetNodePortForQueryService(jaeger))` casts.

pkg/service/query_test.go
```diff
@@ -84,3 +85,21 @@ func TestQueryServiceLoadBalancerWithIngress(t *testing.T) {
...
+	selector := map[string]string{"app": "myapp", "jaeger": name, "jaeger-component": "query"}
+
+	jaeger := v1.NewJaeger(types.NamespacedName{Name: name})
+	jaeger.Spec.Query.ServiceType = corev1.ServiceTypeNodePort
+	jaeger.Spec.Query.NodePort = 32767
+	svc := NewQueryService(jaeger, selector)
+
+	assert.Equal(t, "testqueryservicespecifiednodeportwithingress-query", svc.ObjectMeta.Name)
+	assert.Len(t, svc.Spec.Ports, 1)
+	assert.Equal(t, int32(16686), svc.Spec.Ports[0].Port)
```
rubenvp8510, line 100:
```suggestion
	assert.Equal(t, 16686, svc.Spec.Ports[0].Port)
```

pkg/service/query_test.go
```diff
@@ -84,3 +85,21 @@ func TestQueryServiceLoadBalancerWithIngress(t *testing.T) {
...
+	assert.Equal(t, "http-query", svc.Spec.Ports[0].Name)
+	assert.Equal(t, int32(32767), svc.Spec.Ports[0].NodePort) // make sure we get the same NodePort as set above
```
rubenvp8510, line 102:
```suggestion
	assert.Equal(t, 32767, svc.Spec.Ports[0].NodePort) // make sure we get the same NodePort as set above
```

pkg/service/query_test.go
```diff
@@ -65,6 +65,7 @@ func TestQueryServiceNodePortWithIngress(t *testing.T) {
 	assert.Len(t, svc.Spec.Ports, 1)
 	assert.Equal(t, int32(16686), svc.Spec.Ports[0].Port)
 	assert.Equal(t, "http-query", svc.Spec.Ports[0].Name)
+	assert.Equal(t, int32(0), svc.Spec.Ports[0].NodePort)
```
rubenvp8510, line 68:
```suggestion
	assert.Equal(t, 0, svc.Spec.Ports[0].NodePort)
```

rubenvp8510:
Small comments, overall looks good.

pkg/service/query.go
CSP197: Hi -
I set the function to return an `int` in lieu of an `int32`, for consistency with the `getTargetPortForQueryService` function. Let me make the change and update this MR 👍

pkg/service/query_test.go
CSP197: This is will cause the test to fail, since the value of `t` is expected to be an `int32` value. While the values themselves will be equal, the `assert.Equal` statement will fail due to a type mismatch.

pkg/service/query_test.go
CSP197: This is will cause the test to fail, since the value of `t` is expected to be an `int32` value. While the values themselves will be equal, the `assert.Equal` statement will fail due to a type mismatch.

pkg/service/query_test.go
CSP197: This is will cause the test to fail, since the value of `t` is expected to be an `int32` value. While the values themselves will be equal, the `assert.Equal` statement will fail due to a type mismatch.

rubenvp8510 (approved):
LGTM

PR #1412, opened 2021-03-17
Preparing release v1.22.0

rubenvp8510:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ Changes by Version
 
 1.22.0 (2021-03-16)
 -------------------
-
+* Add ability to indicate PriorityClass for collector and query ([#1413](https://github.com/jaegertracing/jaeger-operator/pull/1413), [@majidazimi](https://github.com/majidazimi))
 * simplest example file should be as simplest ([#1404](https://github.com/jaegertracing/jaeger-operator/pull/1404), [@jkandasa](https://github.com/jkandasa))
 * Add ability to indicate PriorityClass for agent ([#1392](https://github.com/jaegertracing/jaeger-operator/pull/1392), [@elkh510](https://github.com/elkh510))
 * Migrate jaeger.tags in existing CRs ([#1380](https://github.com/jaegertracing/jaeger-operator/pull/1380), [@jpkrohling](https://github.com/jpkrohling))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@ Changes by Version
 
 * simplest example file should be as simplest ([#1404](https://github.com/jaegertracing/jaeger-operator/pull/1404), [@jkandasa](https://github.com/jkandasa))
 * Add ability to indicate PriorityClass for agent ([#1392](https://github.com/jaegertracing/jaeger-operator/pull/1392), [@elkh510](https://github.com/elkh510))
-* Update operatorhub PR template ([#1388](https://github.com/jaegertracing/jaeger-operator/pull/1388), [@jpkrohling](https://github.com/jpkrohling))
 * Migrate jaeger.tags in existing CRs ([#1380](https://github.com/jaegertracing/jaeger-operator/pull/1380), [@jpkrohling](https://github.com/jpkrohling))
 
 1.21.3 (2021-02-09)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,17 @@
 Changes by Version
 ==================
 
+1.22.0 (2021-03-16)
+-------------------
+
+* simplest example file should be as simplest ([#1404](https://github.com/jaegertracing/jaeger-operator/pull/1404), [@jkandasa](https://github.com/jkandasa))
+* Add ability to indicate PriorityClass for agent ([#1392](https://github.com/jaegertracing/jaeger-operator/pull/1392), [@elkh510](https://github.com/elkh510))
+* Update operatorhub PR template ([#1388](https://github.com/jaegertracing/jaeger-operator/pull/1388), [@jpkrohling](https://github.com/jpkrohling))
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,10 +1,17 @@
 Changes by Version
 ==================
 
-1.21.3 (2021-02-09)
+1.22.0 (2021-03-16)
 -------------------
 
+* simplest example file should be as simplest ([#1404](https://github.com/jaegertracing/jaeger-operator/pull/1404), [@jkandasa](https://github.com/jkandasa))
+* Add ability to indicate PriorityClass for agent ([#1392](https://github.com/jaegertracing/jaeger-operator/pull/1392), [@elkh510](https://github.com/elkh510))
+* Update operatorhub PR template ([#1388](https://github.com/jaegertracing/jaeger-operator/pull/1388), [@jpkrohling](https://github.com/jpkrohling))
```
jpkrohling, line 9: This can be left out.

pkg/upgrade/v1_22_0.go
```diff
@@ -19,10 +19,27 @@ func upgrade1_22_0(ctx context.Context, client client.Client, jaeger v1.Jaeger)
...
+			to:   "",
+		},
+		{
+			from: "downsampling.ratio",
+			to:   "",
+		},
+	}
+
+	flagsAll := []deprecationFlagMap{{
+		from: "cassandra.tls.verify-host",
```
jpkrohling, line 34: This probably needs further handling to invert the flag. The names suggest that `verify-host: true` will verify the hostname, while `skip-host-verify: true` suggest that it will not verify the hostname... Ideally, the flag would be removed altogether if the new value is the same as the default value for the new flag: `verify-host: true` will be changed to `skip-host-verify: false`, which is the default value, so, the flag can be removed.

pkg/upgrade/v1_22_0.go
```diff
@@ -30,16 +31,45 @@ func upgrade1_22_0(ctx context.Context, client client.Client, jaeger v1.Jaeger)
...
 	j.Spec.AllInOne.Options = migrateDeprecatedOptions(j, j.Spec.AllInOne.Options, flagMapCollector)
 	j.Spec.Collector.Options = migrateDeprecatedOptions(j, j.Spec.Collector.Options, flagMapCollector)
 	j.Spec.Agent.Options = migrateDeprecatedOptions(j, j.Spec.Agent.Options, flagMapAgent)
 	j.Spec.Query.Options = migrateDeprecatedOptions(j, j.Spec.Query.Options, flagMapQuery)
 
-	return migrateAllDeprecatedOptions(jaeger, flagsAll), nil
+	return migrateCassandraVerifyFlag(jaeger), nil
+}
+
+func flagBoolValue(v interface{}) bool {
```
jpkrohling, line 43: You should either move this to a generic location or prefix it with the version number.

pkg/upgrade/v1_22_0.go
```diff
@@ -30,16 +31,45 @@ func upgrade1_22_0(ctx context.Context, client client.Client, jaeger v1.Jaeger)
...
+			return true
+		}
+		return false
+	}
+
+	boolValue := v.(bool)
+	return boolValue
+}
+
+func migrateCassandraVerifyFlag(j v1.Jaeger) v1.Jaeger {
```
jpkrohling, line 56: Same as the previous comment: prefix this with the version that is using this function.

pkg/upgrade/v1_22_0.go
```diff
@@ -30,16 +31,45 @@ func upgrade1_22_0(ctx context.Context, client client.Client, jaeger v1.Jaeger)
...
+	j.Spec.Collector.Options = updateCassandraVerifyHostFlag(j.Spec.Collector.Options)
+	j.Spec.Storage.Options = updateCassandraVerifyHostFlag(j.Spec.Collector.Options)
+	j.Spec.Ingress.Options = updateCassandraVerifyHostFlag(j.Spec.Collector.Options)
+	return j
+}
+
+func updateCassandraVerifyHostFlag(options v1.Options) v1.Options {
```
jpkrohling, line 63: Ditto

pkg/upgrade/v1_22_0_test.go
```diff
@@ -107,6 +107,35 @@ func TestCassandraRemovedFlags(t *testing.T) {
 	persisted := &v1.Jaeger{}
 	assert.NoError(t, cl.Get(context.Background(), nsn, persisted))
 	assert.Equal(t, latestVersion, persisted.Status.Version)
+	assert.Len(t, persisted.Spec.Collector.Options.Map(), 0)
```
jpkrohling, line 110: You can have a table test for this, instead of two test functions.

pkg/upgrade/utils.go
```diff
@@ -0,0 +1,16 @@
+package upgrade
+
+import "strings"
+
+func flagBoolValue(v interface{}) bool {
+	strValue, isString := v.(string)
+	if isString {
+		if strings.EqualFold(strValue, "true") {
```
jpkrohling, line 8: You can just return the result of EqualFold:

```
if isString {
  return strings.EqualFold(strValue, "true")
}
```

But I wonder why you just don't use `strconv.ParseBool("true")`

jpkrohling (commented):
Other than the strconv comment, looks good to me.

pkg/upgrade/utils.go
rubenvp8510: Done.

jpkrohling (dismissed):

CHANGELOG.md
```diff
@@ -1,10 +1,16 @@
 Changes by Version
 ==================
 
-1.21.3 (2021-02-09)
+1.22.0 (2021-03-16)
 -------------------
 
+* simplest example file should be as simplest ([#1404](https://github.com/jaegertracing/jaeger-operator/pull/1404), [@jkandasa](https://github.com/jkandasa))
```
jpkrohling, line 7: This is now missing #1413

rubenvp8510: Updated.

jpkrohling (approved):

PR #1413, opened 2021-03-17
Add ability to indicate PriorityClass for collector and query

majidazimi:
Note: 

it seems `make generate` introduced some unwanted changes in `deploy/jaegertracing.io_jaegers_crd.yaml`. Not sure whether these changes are required.

### DIFF
--- a/deploy/crds/jaegertracing.io_jaegers_crd.yaml
+++ b/deploy/crds/jaegertracing.io_jaegers_crd.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
+    controller-gen.kubebuilder.io/version: v0.2.4
   creationTimestamp: null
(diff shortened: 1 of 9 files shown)


jpkrohling:
>it seems make generate introduced some unwanted changes in deploy/jaegertracing.io_jaegers_crd.yaml. Not sure whether these changes are required.

That does look odd. Looks like we generated it with v0.3.0 of the controller-gen, but we are installing an earlier version: 
https://github.com/jaegertracing/jaeger-operator/blob/d8fbe6c0fdc9d795ea1c6f1316f6882fe86f8e17/Makefile#L396

Could you try removing the controller-gen and installing it using the following command (if you are using Go 1.16 or higher)?

```
go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4
```

jpkrohling (approved):
Looks great! Thanks for the contribution!

majidazimi:
So using `controller-gen@v0.3.0` passes all e2e tests.

jpkrohling:
cc @rubenvp8510 , this needs to be added to the changelog for 1.22

PR #1418, opened 2021-03-24
Allow configure custom certificates to collector

rubenvp8510:
In the agent (`sidecar` mode) we allow user to set their custom TLS configuration, if we detect that the user set the flat `reporter.tls.enable=true`, we skip the automatic configuration.

That logic is not present on the collector, or in the agent in `daemonset` mode.  This PR fixs it. 

Also added a couple of tests.

### DIFF
--- a/pkg/deployment/agent_test.go
+++ b/pkg/deployment/agent_test.go
@@ -200,30 +200,62 @@ func TestAgentArgumentsOpenshiftTLS(t *testing.T) {
 	viper.Set("platform", v1.FlagPlatformOpenShift)
 	defer viper.Reset()
 
-	jaeger := v1.NewJaeger(types.NamespacedName{
-		Name:      "my-instance",
-		Namespace: "test",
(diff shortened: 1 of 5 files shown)


jpkrohling:
@kevinearls would you mind reviewing this?

pkg/deployment/agent_test.go
```diff
@@ -197,33 +197,66 @@ func TestAgentCustomReporterPort(t *testing.T) {
 }
 
 func TestAgentArgumentsOpenshiftTLS(t *testing.T) {
+
```
kevinearls, line 200: nit - I don't think there should be whitespace here.

kevinearls (approved):
LGTM except for 1 very minor item.

pkg/deployment/agent_test.go
rubenvp8510: Removed.

Issue #1415, opened 2021-03-18
support repetitive arguments to operand


jkandasa:
For now, there is no way to supply repetitive arguments options to the operand

I am expecting the following arguments to query pod, how to specify this in jaeger CR?
```
--query.additional-headers=access-control-allow-origin:blerg
--query.additional-headers=whatever:thing
```

```yaml
spec:
  query:
    options:
      query:
         additional-headers: "whatever:thing"
```

AFAIK, in YAML we cannot supply the same key multiple times
https://github.com/jaegertracing/jaeger-operator/blob/v1.21.3/pkg/apis/jaegertracing/v1/options.go#L96~L99

rubenvp8510:
i think something like this could be an option:
```
...
spec:
  query:
    options:
      query:
         additional-headers:
              -  "whatever:thing"
              -  "access-control-allow-origin:blerg"
```

PR #1434, opened 2021-04-21
Add support repetitive arguments to operand

rubenvp8510:
Fixes https://github.com/jaegertracing/jaeger-operator/issues/1415


jpkrohling:
CI seems to be failing.

jpkrohling:
Let me know when this gets ready to be reviewed.

rubenvp8510:
This is ready for review

pkg/apis/jaegertracing/v1/options_test.go
```diff
@@ -148,3 +150,38 @@ func TestUpdate(t *testing.T) {
 	// verify
 	assert.Equal(t, o.opts["key"], "new")
 }
+
+func TestStringMap(t *testing.T) {
+	o := NewOptions(nil)
+	err := o.UnmarshalJSON([]byte(`{"firstsarg":"v1", "additional-headers":["whatever:thing", "access-control-allow-origin:blerg"]}`))
```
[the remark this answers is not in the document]
rubenvp8510: I'll add some real examples, 

This is testing a new method I introduced in this PR. `StringMap`, this method only returns a map with strings (mimic the old behavior when we didn't support arrays) , it omits repeated arguments, I did in this way because this map is used in a lot of places, and I wand to preserve that behaviour for those places which is used, also I don't want to do type asserting each time the option map is used.

pkg/apis/jaegertracing/v1/options.go
```diff
@@ -7,10 +7,28 @@ import (
 	"strings"
 )
 
+// Values represent and string or a slice of string
```
jpkrohling, line 10: This doesn't seem accurate. Perhaps you meant "Values hold a map, with string as the key and either a string or a slice of strings as the value"

In any case, the name for this struct isn't really telling what the struct is. Can you come up with a better name?

pkg/apis/jaegertracing/v1/options.go
```diff
@@ -69,18 +86,24 @@ func (o Options) MarshalJSON() ([]byte, error) {
 }
 
 func (o *Options) parse(entries map[string]interface{}) {
-	o.opts = make(map[string]string)
+	o.opts = make(map[string]interface{})
```
jpkrohling, line 89: Don't we have a structure with this signature already?

pkg/apis/jaegertracing/v1/options_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
```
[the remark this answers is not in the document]
jpkrohling: Looks like this hasn't been fixed.

pkg/apis/jaegertracing/v1/options.go
rubenvp8510: Better comment :) I'll change it, and rename this to a better name

pkg/apis/jaegertracing/v1/options.go
rubenvp8510: I can't find it, We have `FreeForm` but that is a different structure with` json *[]byte`

rubenvp8510:
@pavolloffay Could you please review?

Thank you!

pavolloffay (approved):
LGTM, one test job is failing

PR #1440, opened 2021-04-28
Allow TLS flags to be disabled

### LINKED ISSUE #1438: Disabling TLS in the collector keeps tls=true option
As reported by @kenfinnigan, setting the option `reporter.grpc.tls.enabled` to false on a CR that has it set to true (like in a CR on OpenShift), will cause the option to be set at both true and false.

rubenvp8510:
Fixes #1438 

### DIFF
--- a/pkg/deployment/all_in_one.go
+++ b/pkg/deployment/all_in_one.go
@@ -74,7 +74,7 @@ func (a *AllInOne) Get() *appsv1.Deployment {
 	// even though the agent is in the same process as the collector, they communicate via gRPC, and the collector has TLS enabled,
 	// as it might receive connections from external agents
 	if viper.GetString("platform") == v1.FlagPlatformOpenShift {
-		if len(util.FindItem("--reporter.grpc.tls.enabled=true", options)) == 0 {
+		if len(util.FindItem("--reporter.grpc.tls.enabled=", options)) == 0 {
 			options = append(options, "--reporter.grpc.tls.enabled=true")
 			options = append(options, fmt.Sprintf("--reporter.grpc.tls.ca=%s", ca.ServiceCAPath))
 			options = append(options, fmt.Sprintf("--reporter.grpc.tls.server-name=%s.%s.svc.cluster.local", service.GetNameForHeadlessCollectorService(a.jaeger), a.jaeger.Namespace))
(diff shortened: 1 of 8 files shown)


pkg/deployment/all_in_one.go
```diff
@@ -63,15 +63,17 @@ func (a *AllInOne) Get() *appsv1.Deployment {
 
 	configmap.Update(a.jaeger, commonSpec, &options)
 	sampling.Update(a.jaeger, commonSpec, &options)
-	tls.Update(a.jaeger, commonSpec, &options)
-	ca.Update(a.jaeger, commonSpec)
-	ca.AddServiceCA(a.jaeger, commonSpec)
+	if len(util.FindItem("--collector.grpc.tls.enabled", options)) == 0 {
+		tls.Update(a.jaeger, commonSpec, &options)
```
jpkrohling, line 67: This doesn't feel right: if the option is set, then tls.Update will be skipped (and ca, and ca.AddServiceCA). Why did we have this as part of all reconciliation loops before? What happens when on the first reconciliation we did not have this parameter, causing the operator to add `--reporter.grpc.tls.enabled=true`, and the user then manually sets this to `false`?

jpkrohling (commented):
Could you please confirm that you ran a manual test like the following?

- On OpenShift, create a simple CR. This should enable TLS by default
- Change the CR to disable TLS
- Check that the CR (and underlying Jaeger config) look sane and without extra options like paths to the certs and keys.

pkg/deployment/all_in_one.go
rubenvp8510: Good point, I need to test this case, 

I ran manually tests only for the first case, I'll check what happens in the case of an update to the CR. but it seems like you're right.

rubenvp8510:
I did some changes to only touch the flags in case of the TLS certificates, still creating CA on the reconciliation, regarding of the flags.

pkg/deployment/all_in_one.go
jpkrohling: Did you test this case?

jpkrohling (commented):
This looks good, but it's not still clear to me what happens when a collector is first provisioned with the TLS options, and then has the TLS options disabled. Did you test this case? Is it possible to simulate this with an e2e test?

pkg/deployment/all_in_one.go
rubenvp8510: I tested it with the last changes I did, it seems to work fine.

jpkrohling:
@rubenvp8510 could you please rebase this?

rubenvp8510:
> This looks good, but it's not still clear to me what happens when a collector is first provisioned with the TLS options, and then has the TLS options disabled. Did you test this case? Is it possible to simulate this with an e2e test?

The TLS will be disabled, but all the volumes will be still there.

jpkrohling (approved):

PR #1447, opened 2021-05-10
Pass secretName to cassandra dependencies job (#1162)

### LINKED ISSUE #1162: Allow storage.options.cassandra.password to have an `existingSecret` reference to pick up cassandra password from
HI,

There is ability to connect Cassandra storage with username and password through jaeger-operator, but there is no way to use a kubernetes secret for the password field and the Jaeger CR would need to have it in plaintext. This is an issue since we would store the Jaeger CR definition in git.

Even with templating through helm/kustomize, it'd mean the operator needs to store the secret somewhere in their own env. A much cleaner way would be able to refer to an existing kubernetes secret.

a `useExistingSecret: <kubernetes secret name>` would help here, since we would spin up cassandra with either the bitnami/cassandra chart or scylla-operator (among others), both of which support the `existingSecret` paradigm to create the clusters.

Gerrit-K:
Fixes #1162 

Signed-off-by: Gerrit Kieffer <kieffer.gerrit@gmail.com>


Gerrit-K:
A couple of notes on this:
* Depending on whether or not `secretName` is set, we need to unset the `CASSANDRA_*` environment variables because otherwise the values from the secret would be overridden. This unfortunately forced me to restructure the whole env handling here. I don't really like my solution, but I'm also not very familiar with go, so please let me know if you have any suggestions on how to improve this :)
* The other changes in the diff of the `Container` is simply spacing induced by `go fmt`
* The unit test currently only checks for the existence of `EnvFrom` in the final `Job` (similarly to [`all_in_one_test.go`](https://github.com/jaegertracing/jaeger-operator/blob/4693d119aabb550a7a651315cbfce1833bb786fc/pkg/deployment/all_in_one_test.go#L150-L159))

jpkrohling:
@rubenvp8510, please review this one.

rubenvp8510 (approved):
LGTM

pkg/storage/cassandra_dependencies.go
```diff
@@ -36,26 +36,64 @@ func cassandraDeps(jaeger *v1.Jaeger) []batchv1.Job {
...
 
-	username := jaeger.Spec.Storage.Options.Map()["cassandra.username"]
-	password := jaeger.Spec.Storage.Options.Map()["cassandra.password"]
+	envVars = append(envVars, corev1.EnvVar{
+		Name:  "KEYSPACE",
+		Value: keyspace,
+	})
+
+	var envFromSource []corev1.EnvFromSource
+	if len(jaeger.Spec.Storage.SecretName) > 0 {
```
rubenvp8510, line 79: There is a function on utils that we can reuse to create this 

https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/util/util.go#L280

Gerrit-K: Neat! Thanks for the hint, I'll replace it. And thanks for reviewing 👍

jpkrohling (approved):

rubenvp8510:
@Gerrit-K Could you rebase please?

Gerrit-K:
@rubenvp8510 apparently I forgot to reply after I rebased on master and then totally forgot about this, sorry ...
Is there anything else I can / should do? Do we need another update or is the branch not too stale yet?

rubenvp8510:
@Gerrit-K  sorry I didn't realized that you rebased and all test passed 

This PR LGTM,.

rubenvp8510 (approved):

Gerrit-K:
@rubenvp8510 Thanks for the approval! Unfortunately there's now a kuttl test that failed 😕 To me it looks like a timeout and it doesn't seem related to my changes (at least it seems to happen for several other recent builds as well), but I don't really know, neither can I restart the failed check. Could you take another look?

rubenvp8510:
Now all test are on green, @jpkrohling WDYT?

jpkrohling (approved):

PR #1456, opened 2021-05-19
Promote crd to apiextensions.k8s.io/v1

rubenvp8510:

This might break compatibility with k8s < 1.19, but is a necessary step to support k8s 1.22+ 

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22


jpkrohling (approved):
LGTM, but just to confirm: did you try this in OpenShift as well?

Makefile
```diff
@@ -393,7 +393,7 @@ install-tools:
 		golang.org/x/lint/golint \
 		golang.org/x/tools/cmd/goimports \
 		github.com/securego/gosec/cmd/gosec@v0.0.0-20191008095658-28c1128b7336 \
-		sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0 \
+		sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0 \
```
pavolloffay, line 396: Did bumping this version update the CRD yamls?

pavolloffay: or did you run any command to generate CRD yalms?

rubenvp8510: I ran the controller-gen command and it generates the new CRD with the version.

rubenvp8510:
> LGTM, but just to confirm: did you try this in OpenShift as well?

Tested with CRC and all-in-one strategy and it works .

jpkrohling:
I think I restarted the tests a couple of times already. If this fails again, it might be a real failure, @rubenvp8510

rubenvp8510:
> I think I restarted the tests a couple of times already. If this fails again, it might be a real failure, @rubenvp8510

Yes, I was investigating about the failures, it seems like I need to migrate the Ingress API too, and because of that I need to bump some versions of the libraries, specifically `k8s.io/api`. I'll request your review again once I finish.

pavolloffay (approved):
LGTM,

does this PR break compatibility with older k8s clusters?

.github/workflows/e2e-kubernetes.yaml
```diff
@@ -8,28 +8,28 @@ on:
...
     steps:
     - uses: actions/setup-go@v2.1.3
       with:
-        go-version: '1.14.4'
-    - uses: jpkrohling/setup-minikube@v1.0.2
+        go-version: '1.14.12'
+    - uses: actions/checkout@v2.3.4
+    - name: "setup docker"
+      run: ./.ci/setup-docker.sh
+    - uses: manusa/actions-setup-minikube@v2.4.1
```
pavolloffay, line 22: nice I didn't see this. In the future we could also consider using `kind`.

jpkrohling: Indeed, the new e2e tests are using kind and kuttl :-)

rubenvp8510:
> LGTM,
> 
> does this PR break compatibility with older k8s clusters?

Yes :( 1.19+ is required.

PR #1465, opened 2021-06-04
Add compatibility matrix

jpkrohling:
This change adds a compatibility matrix to the readme file. It also extends the kuttl e2e tests to be executed in the N-2 versions of Kubernetes.

﻿Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>


kind-1.19.yaml
```diff
@@ -0,0 +1,18 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  image: kindest/node:v1.19.11@sha256:07db187ae84b4b7de440a73886f008cf903fcf5764ba8106a9fd5243d6f32729
```
pavolloffay, line 5: why using hash?

pavolloffay (approved):

kind-1.19.yaml
jpkrohling: So that it won't try to get a new version from a remote repository if it has the hash locally. It also has a security aspect to it: a man-in-the-middle wouldn't be able to manipulate the image we pull.

It probably doesn't matter much in this case, but it's recommended by kind.

PR #1467, opened 2021-06-08
Remove OwnerReferences from CA configmaps

rubenvp8510:

This change is to avoid `ConfigMaps` which are not in the same namespace as the operator from being garbage collected.

As the documentation says:

> 
> Cross-namespace owner references are disallowed by design.
> Namespaced dependents can specify cluster-scoped or namespaced owners. A namespaced owner must exist in the same namespace as the dependent. If it does not, the owner reference is treated as absent, and the dependent is subject to deletion once all owners are verified absent.

https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

- This solves an issue that deletes CA/Service CA `ConfigMaps` from sidecars that are not in the same namespace as the operator.

- Removed the `OwnerReferences` from `ConfigMaps`
- Implemented a deletion logic for those `ConfigMaps` when a `Jaeger` instance is deleted.

**NOTE**: The clean up process consider that there is no more than one Jaeger instance with the same name on the cluster. I think we did this assumption for other parts of the code, so I don't think this will be a problem for now.


rubenvp8510:
Struggling to reach the desired coverage due the tests can't reach error blocks.

pkg/controller/jaeger/configmap.go
```diff
@@ -3,14 +3,18 @@ package jaeger
 import (
 	"context"
 
+	"k8s.io/apimachinery/pkg/selection"
```
jpkrohling, line 6: Import order.

pkg/controller/jaeger/configmap.go
```diff
@@ -63,3 +67,30 @@ func (r *ReconcileJaeger) applyConfigMaps(ctx context.Context, jaeger v1.Jaeger,
 
 	return nil
 }
+
+func (r *ReconcileJaeger) getSelectorsForConfigMaps(instanceName string, components []string) labels.Selector {
+	nameReq, _ := labels.NewRequirement("app.kubernetes.io/name", selection.DoubleEquals, []string{util.Truncate(instanceName, 63)})
```
jpkrohling, line 72: Document why you are swallowing the error.

jpkrohling (commented):
Looks good, just need to properly format the imports and document why errors are being swallowed.

pkg/controller/jaeger/configmap.go
rubenvp8510: Done

pkg/controller/jaeger/configmap.go
rubenvp8510: Ups.. my bad. Fixed

jpkrohling (approved):

PR #1484, opened 2021-06-24
Instrument instances types

rubenvp8510:

-  First approach of instrumenting Jaeger operator to know more about operands.

-  Use Prometheus exporter and OpenTelemety metrics API, created a `valueObserver` and query in every callback all the managed instances by this operator. The reason of using a `valueObserver` is because I cannot relies on reconciliation process and an `UpDownCounter` instrument because the reconciliation process can be execute multiple times for creation/update/delete and I can't distinguish when is an update or creation.


pkg/metrics/bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const meterName = "jaegertracing.io/jaeger"
+
+// Bootstrap prepares a new tracer to be used by the operator
+func Bootstrap(ctx context.Context, namespace string, client client.Client) {
+	tracer := otel.GetTracerProvider().Tracer(v1.CustomMetricsTracer)
+	ctx, span := tracer.Start(ctx, "bootstrap")
```
pavolloffay, line 30: is this span recording events/ reported? 

If the bootstrap creates tracer than the span might be noop/default if it is created before initialization.

pkg/metrics/bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+	processor "go.opentelemetry.io/otel/sdk/metric/processor/basic"
+	selector "go.opentelemetry.io/otel/sdk/metric/selector/simple"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const meterName = "jaegertracing.io/jaeger"
+
+// Bootstrap prepares a new tracer to be used by the operator
+func Bootstrap(ctx context.Context, namespace string, client client.Client) {
```
pavolloffay, line 28: should this return some kind of closer interface to flush the spans on termination event?

pavolloffay: When is this method called? Periodically or just once?

pkg/metrics/instances.go
```diff
@@ -0,0 +1,105 @@
...
+	if storage == "" {
+		storage = "memory"
+	}
+	return strings.ToLower(storage)
+
+}
+func (i *instancesMetric) strategy(jaeger v1.Jaeger) string {
+	strategy := string(jaeger.Spec.Strategy)
+	if strategy == "" {
+		return "allinone"
```
pavolloffay, line 65: if this is the convention if strategy is empty then we should define a method on the `Spec` object to return a strategy and set the allinone as default.

pavolloffay: The same for memory

pavolloffay: and agent sidecar

pkg/metrics/instances.go
```diff
@@ -0,0 +1,105 @@
...
+		groups: make(map[string]*instancesCounter),
+	}
+}
+
+func (i *instancesMetric) Setup(ctx context.Context) error {
+	tracer := otel.GetTracerProvider().Tracer(v1.CustomMetricsTracer)
+	ctx, span := tracer.Start(ctx, "setup-jaeger-instances")
+	defer span.End()
+	meter := global.Meter(meterName)
+	_, err := meter.NewInt64ValueObserver("operator_jaeger_instances", i.callback,
```
jpkrohling, line 42: With which frequency is the callback called?

pkg/metrics/bootstrap.go
rubenvp8510: This is called just once.

pkg/metrics/instances.go
rubenvp8510: are we using `Defaulters` in this version of the operator? I started to do it on v2 but IIRC we don't use it here.

pkg/metrics/bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+	controller "go.opentelemetry.io/otel/sdk/metric/controller/basic"
+	processor "go.opentelemetry.io/otel/sdk/metric/processor/basic"
+	selector "go.opentelemetry.io/otel/sdk/metric/selector/simple"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const meterName = "jaegertracing.io/jaeger"
+
+// Bootstrap prepares a new tracer to be used by the operator
```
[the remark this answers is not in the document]
rubenvp8510: not at all haha. sorry I'll change it.

pkg/metrics/bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+func Bootstrap(ctx context.Context, namespace string, client client.Client) {
+	tracer := otel.GetTracerProvider().Tracer(v1.CustomMetricsTracer)
+	ctx, span := tracer.Start(ctx, "bootstrap")
+	defer span.End()
+	tracing.SetInstanceID(ctx, namespace)
```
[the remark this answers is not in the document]
rubenvp8510: I see that function only add labels to the span with the instance operator information.

pkg/metrics/instances.go
rubenvp8510: IIRC this is called per collection interval.

pkg/metrics/instances.go
jpkrohling: True. We have a normalizer in this operator.

pkg/metrics/bootstrap.go
rubenvp8510: The tracer is already created when this bootstraping is called, this bootstraps the metrics for Open Telemetry,

rubenvp8510:
Address all comments, added an screen-shot with the visualization of the metrics.

pavolloffay:
> Address all comments, added an screen-shot with the visualization of the metrics.

If the metrics are exposed in p8s nedpoint. It might be better to paste the metrics here in the text form. At least for me it is the best way to understand what is being exposed.

rubenvp8510:
> > Address all comments, added an screen-shot with the visualization of the metrics.
> 
> If the metrics are exposed in p8s nedpoint. It might be better to paste the metrics here in the text form. At least for me it is the best way to understand what is being exposed.


```
# HELP operator_jaeger_instances Number of jaeger instances in cluster
# TYPE operator_jaeger_instances gauge
operator_jaeger_instances{agent="sidecar",service_name="unknown_service:jaeger-operator-local",storage="elasticsearch",strategy="production",telemetry_sdk_language="go",telemetry_sdk_name="opentelemetry",telemetry_sdk_version="0.20.0"} 1
operator_jaeger_instances{agent="sidecar",service_name="unknown_service:jaeger-operator-local",storage="elasticsearch",strategy="streaming",telemetry_sdk_language="go",telemetry_sdk_name="opentelemetry",telemetry_sdk_version="0.20.0"} 0
operator_jaeger_instances{agent="sidecar",service_name="unknown_service:jaeger-operator-local",storage="memory",strategy="allinone",telemetry_sdk_language="go",telemetry_sdk_name="opentelemetry",telemetry_sdk_version="0.20.0"} 1
```

pavolloffay:
The metrics look good. I would rename `agent` to `agent_strategy`

Are the following labels needed? What do they provide us?  

```
service_instance_id="default.jaeger-operator",service_name="jaeger-operator",service_namespace="default
```

rubenvp8510:
> The metrics look good. I would rename `agent` to `agent_strategy`
> 
> Are the following labels needed? What do they provide us?
> 
> ```
> service_instance_id="default.jaeger-operator",service_name="jaeger-operator",service_namespace="default
> ```

I  don't think we need it, we can get rid of it

WDYT? @jpkrohling

jpkrohling:
If we keep the name/namespace labels, we will have high cardinality here. In fact, I would even split the metrics into this:

```
jaeger_operator_instances_agent_modes{mode=sidecar} 1
jaeger_operator_instances_agent_modes{mode=daemonset} 1
jaeger_operator_instances_storage_types{type=memory} 1
jaeger_operator_instances_storage_types{type=elasticsearch} 1
jaeger_operator_instances_strategies{type=production} 1
jaeger_operator_instances_strategies{type=allinone} 1
jaeger_operator_instances_versions{version=1.23.0} 1
jaeger_operator_instances_versions{version=1.24.0} 1
```

With that, we only have a limited number of time series, the last one being the only one that can potentially grow "indefinitely".

rubenvp8510:
I changed the metrics with the suggested way:

```
# HELP jaeger_operator_instances_agent_modes Number of instances per agent strategy
# TYPE jaeger_operator_instances_agent_modes gauge
jaeger_operator_instances_agent_modes{agent_modes="sidecar"} 1
# HELP jaeger_operator_instances_storage Number of instances per storage type
# TYPE jaeger_operator_instances_storage gauge
jaeger_operator_instances_storage{storage="memory"} 1
# HELP jaeger_operator_instances_strategy Number of instances per strategy type
# TYPE jaeger_operator_instances_strategy gauge
jaeger_operator_instances_strategy{strategy="allinone"} 1

```

pkg/cmd/start/bootstrap.go
```diff
@@ -85,7 +86,10 @@ func bootstrap(ctx context.Context) manager.Manager {
 	serveCRMetrics(ctx, cfg, namespace)
 	createMetricsService(ctx, cfg, namespace)
 	detectOAuthProxyImageStream(ctx, mgr)
-
+	err = opmetrics.Bootstrap(ctx, namespace, mgr.GetClient())
+	if err != nil {
+		log.WithError(err).Error("Failed to initialize metrics")
```
jpkrohling, line 91: Errors in Go shouldn't start with a capital letter, except if they are the highest level (ie, at the main.go for the program).

pkg/metrics/instances.go
```diff
@@ -0,0 +1,156 @@
...
+		})
+	if err != nil {
+		return err
+	}
+	i.observations = append(i.observations, obs)
+	return nil
+}
+
+func isInstanceNormalized(jaeger v1.Jaeger) bool {
+	return !(jaeger.Spec.Storage.Type == "" || jaeger.Spec.Strategy == "")
```
jpkrohling, line 118: When is this the case? The instance should be normalized as soon as it's persisted.

pkg/apis/jaegertracing/v1/const.go
```diff
@@ -30,4 +30,7 @@ const (
 
 	// ReconciliationTracer is the OpenTelemetry tracer name for the reconciliation loops
 	ReconciliationTracer string = "operator/reconciliation"
+
+	// CustomMetricsTracer is the OpenTelemetry tracer name for the custom metrics setup
+	CustomMetricsTracer string = "operator/bootstrap-custom-metrics"
```
jpkrohling, line 35: Is there a reason to have this in a different tracer? Why not reuse the bootstrap one, given that this setup is indeed part of the main bootstrap routine?

pkg/metrics/instances.go
```diff
@@ -0,0 +1,156 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/global"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
+)
+
+const metricPrefix = "jaeger_operator_instances"
```
jpkrohling, line 17: This whole file needs to be testable. For instance, prove with tests that the counter increases when there's a new CR and that it decreases when a CR is removed.

pkg/metrics/instances.go
rubenvp8510: This case is when the reconciliation process haven't ran yet it and this function is called.

pkg/metrics/instances.go
rubenvp8510: I'll write tests for this.

rubenvp8510:
@jpkrohling  @pavolloffay Please give it another review, I added tests and a flag to enable this feature.

pkg/metrics/bootstrap.go
```diff
@@ -0,0 +1,62 @@
+package metrics
+
+import (
+	"context"
+
+	"github.com/spf13/viper"
```
jpkrohling, line 6: Import order.

pkg/cmd/start/main.go
```diff
@@ -57,6 +57,7 @@ func NewStartCommand() *cobra.Command {
 	cmd.Flags().Int32("cr-metrics-port", 8686, "The metrics port for Operator and/or Custom Resource based metrics")
 	cmd.Flags().String("jaeger-agent-hostport", "localhost:6831", "The location for the Jaeger Agent")
 	cmd.Flags().Bool("tracing-enabled", false, "Whether the Operator should report its own spans to a Jaeger instance")
+	cmd.Flags().Bool("operand-metrics-enabled", false, "Whether the Operator should report metrics about the operands.")
```
jpkrohling, line 60: Why do we need this? The PR was so closed to get merged without this feature... Remember that smaller PRs get reviewed/merged faster.

pkg/metrics/bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
+	"github.com/jaegertracing/jaeger-operator/pkg/tracing"
+)
+
+const meterName = "jaegertracing.io/jaeger"
+
+// Bootstrap configures the OpenTelemetry meter provides with the prometheus exporter
```
jpkrohling, line 28:
```suggestion
// Bootstrap configures the OpenTelemetry meter provider with the Prometheus exporter.
```

pkg/metrics/bootstrap.go
```diff
@@ -0,0 +1,62 @@
...
+func Bootstrap(ctx context.Context, namespace string, client client.Client) error {
+	if viper.GetBool("operand-metrics-enabled") {
```
jpkrohling, line 30: If you invert the check, you can just return `nil` at the beginning of the function, letting the remaining of the func to reside outside the conditional, simplifying the understanding of this func in the long term.

pkg/cmd/start/main.go
rubenvp8510: Well, I tried to maintain a similar option like one that we have for tracing.

rubenvp8510: Also, I don't want to fetch the instances and do all the metrics logic if it's not requires, so people don't want to do it have this option to disable.

jpkrohling: It's not required, and it serves a different purpose as the tracing flag. The tracing flag serves as debugging help for the operator, while the metrics provide an insight into the setup over time.

In fact, I think we can even consider removing the tracing-enabled flag. Could you create a new issue to discuss that?

rubenvp8510: Removed! I'll create the issue.

rubenvp8510:
I address all comments, removed the flag, I think this is ready for another review and hopefully a merge =)

pkg/metrics/bootstrap.go
```diff
@@ -0,0 +1,57 @@
...
+	defer span.End()
+	tracing.SetInstanceID(ctx, namespace)
+
+	config := prometheus.Config{
+		Registry: metrics.Registry.(*prometheusclient.Registry),
+	}
+	c := controller.New(
+		processor.New(
+			selector.NewWithHistogramDistribution(
+				histogram.WithExplicitBoundaries(config.DefaultHistogramBoundaries),
```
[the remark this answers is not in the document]
rubenvp8510: None of that is applicable to us. for now. That is why I just set default values.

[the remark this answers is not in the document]
jpkrohling: What are the default values? If we don't have histograms, why do we configure them? Can't we just leave this out?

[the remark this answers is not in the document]
rubenvp8510: I can't I need to pass and aggregation selector to the New method.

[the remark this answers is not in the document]
rubenvp8510:
```
// NewWithHistogramDistribution returns a simple aggregator selector
// that uses histogram aggregators for `ValueRecorder` instruments.
// This selector is a good default choice for most metric exporters.
```

From this comment I'm assuming this is the best "default" configuration.

jpkrohling (approved):

PR #1490, opened 2021-06-25
Fix overwritten default labels in label selectors of `Service`

### LINKED ISSUE #1479: Service selector does not match custom Deployment labels
**Describe the bug**
When creating a `Jaeger` object with custom `labels`, they are propagated to the `Deployment` (and hence the `Pod`) correctly, but unfortunately not to the `Service`(s). This is especially causing issues if you (intentionally or by accident) try to override the default labels, such as `app.kubernetes.io/name`, because then the `Service`s do not match the `Pod` labels anymore.

It might be debatable that a user should be allowed to overwrite those default labels, but it should at least be consistent (i.e. either override them everywhere, including the `Service`s, or disallow that in the `Deployment` as well) in order to not break the overall application.

**To Reproduce**
1. Deploy a `Jaeger` instance with `spec: { labels: { "app.kubernetes.io/name": "jaeger-with-custom-name" } }` (be sure to  use a different name than for the `Jaeger` object itself)

**Expected behavior**
The `Service` selectors should match the respective `Deployment`, even when using custom labels.

 - OS: Linux nodes on AKS
 - Jaeger version: 1.22.1 (Operator image version)
 - Deployment: Kubernetes

rudeigerc:
Signed-off-by: Yuchen Cheng <rudeigerc@gmail.com>

- Resolves #1479 

- Fix overwritten default labels in label selectors of `Service`.
- Remove reserved labels in the normalize operation.

### DIFF
--- a/pkg/deployment/all_in_one_test.go
+++ b/pkg/deployment/all_in_one_test.go
@@ -93,6 +93,42 @@ func TestAllInOneLabels(t *testing.T) {
 	assert.Equal(t, "false", dep.Spec.Selector.MatchLabels["another"])
 }
 
+func TestAllInOneOverwrittenDefaultLabels(t *testing.T) {
+	jaeger := v1.NewJaeger(types.NamespacedName{Name: "TestAllInOneOverwrittenDefaultLabels"})
+	jaeger.Spec.Labels = map[string]string{
+		"name":                   "operator",
+		"hello":                  "jaeger",
+		"app.kubernetes.io/name": "my-jaeger", // Override default labels
+	}
+	jaeger.Spec.AllInOne.Labels = map[string]string{
+		"hello":   "world", // Override top level annotation
+		"another": "false",
+	}
+
+	allinone := NewAllInOne(jaeger)
+	dep := allinone.Get()
+
+	assert.Equal(t, "operator", dep.Spec.Template.Labels["name"])
+	assert.Equal(t, "world", dep.Spec.Template.Labels["hello"])
+	assert.Equal(t, "false", dep.Spec.Template.Labels["another"])
+	assert.Equal(t, "my-jaeger", dep.Spec.Template.Labels["app.kubernetes.io/name"])
+
+	// Deployment selectors should be the same as the template labels.
+	assert.Equal(t, "operator", dep.Spec.Selector.MatchLabels["name"])
+	assert.Equal(t, "world", dep.Spec.Selector.MatchLabels["hello"])
+	assert.Equal(t, "false", dep.Spec.Selector.MatchLabels["another"])
+	assert.Equal(t, "my-jaeger", dep.Spec.Selector.MatchLabels["app.kubernetes.io/name"])
+
+	// Service selectors should be the same as the template labels.
+	services := allinone.Services()
+	for _, svc := range services {
+		assert.Equal(t, "operator", svc.Spec.Selector["name"])
+		assert.Equal(t, "world", svc.Spec.Selector["hello"])
+		assert.Equal(t, "false", svc.Spec.Selector["another"])
+		assert.Equal(t, "my-jaeger", svc.Spec.Selector["app.kubernetes.io/name"])
+	}
+}
+
 func TestAllInOneHasOwner(t *testing.T) {
 	name := "TestAllInOneHasOwner"
 	a := NewAllInOne(v1.NewJaeger(types.NamespacedName{Name: name}))
--- a/pkg/strategy/controller.go
+++ b/pkg/strategy/controller.go
@@ -79,14 +79,20 @@ func normalize(ctx context.Context, jaeger *v1.Jaeger) {
 	}
 
 	// remove reserved labels
-	delete(jaeger.Spec.JaegerCommonSpec.Labels, "app.kubernetes.io/instance")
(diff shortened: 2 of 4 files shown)


rubenvp8510 (dismissed):
LGTM

pkg/deployment/all_in_one_test.go
```diff
@@ -91,6 +91,42 @@ func TestAllInOneLabels(t *testing.T) {
...
+		"another": "false",
+	}
+
+	allinone := NewAllInOne(jaeger)
+	dep := allinone.Get()
+
+	assert.Equal(t, "operator", dep.Spec.Template.Labels["name"])
+	assert.Equal(t, "world", dep.Spec.Template.Labels["hello"])
+	assert.Equal(t, "false", dep.Spec.Template.Labels["another"])
+	assert.Equal(t, "my-jaeger", dep.Spec.Template.Labels["app.kubernetes.io/name"])
```
jpkrohling, line 112: Should we really allow this? Don't we rely on this for the reconciliation?

rudeigerc: Sorry I have missed your comment in the original issue.
Maybe we could reorder the spec during `util.Merge` so that we could ensure the reserved labels not be overwritten?

jpkrohling: I'm really not sure what's the desired behavior. If we are not using this field internally, we should allow people to override it. The only labels are using are these:

https://github.com/jaegertracing/jaeger-operator/blob/b24e2a716cd4f4efbba617fb29025f6edcfab24d/pkg/controller/jaeger/account.go#L21-L27

@rubenvp8510, what do you think?

rubenvp8510: I was thinking on this carefully, I think the correct behaviour should be not allowed to override the labels we relies on for the reconciliation process. All other labels can be overrides in a consistent way.

jpkrohling: Would you be able to update this PR then, @rudeigerc?

rudeigerc: @jpkrohling Sure. So only those two labels mentioned above should not be overwritten?

jpkrohling: Yes. Actually, does it require a change at all? This is already the behavior of this PR, I suppose.

jpkrohling: Or removing the reserved entries from the CR in the normalize operation?

https://github.com/jaegertracing/jaeger-operator/blob/050e95a671d5cfceda848055d61af0968f5f68a3/pkg/strategy/controller.go#L54-L55

rudeigerc: Your solution sounds better. 😂
I will take a look at it. Thanks for your guidance.

rudeigerc: @rubenvp8510 @jpkrohling Hi, I've made an update. Please have a look at it.

pkg/strategy/controller.go
```diff
@@ -78,6 +78,16 @@ func normalize(ctx context.Context, jaeger *v1.Jaeger) {
 		jaeger.Spec.Storage.Type = v1.JaegerMemoryStorage
 	}
 
+	// remove reserved labels
+	delete(jaeger.Spec.JaegerCommonSpec.Labels, "app.kubernetes.io/instance")
```
jpkrohling, line 82: I don't like silently removing things that users explicitly added, but I don't think we have much of a choice here. Perhaps at least add an event to the span when the item existed?

rudeigerc: Sure. I've added loggings when removing reserved labels.

pkg/strategy/controller.go
```diff
@@ -79,14 +79,20 @@ func normalize(ctx context.Context, jaeger *v1.Jaeger) {
...
-
-	delete(jaeger.Spec.Query.JaegerCommonSpec.Labels, "app.kubernetes.io/instance")
-	delete(jaeger.Spec.Query.JaegerCommonSpec.Labels, "app.kubernetes.io/managed-by")
+	for _, labels := range []map[string]string{
+		jaeger.Spec.JaegerCommonSpec.Labels,
+		jaeger.Spec.AllInOne.JaegerCommonSpec.Labels,
+		jaeger.Spec.Query.JaegerCommonSpec.Labels,
+	} {
+		if _, ok := labels["app.kubernetes.io/instance"]; ok {
+			jaeger.Logger().Warnf("The reserved label 'app.kubernetes.io/instance' is overwritten. Falling back to %s", jaeger.Name)
```
jpkrohling, line 88: Sorry for being picky, but it really belongs as a span event. A span is created on line 58, you can use it here. Unfortunately, we do not have a good way to send back this kind of information to the user who created the CR (event, status message, ...) in this code, as we have in the upgrade package, and leaving log messages at the operator stdout would likely only pollute the logs. Logging to the span would allow people looking at the trace for a reconciliation to see this in connection to the bigger context.

jpkrohling (approved):

jpkrohling:
Thanks for the PR!

rudeigerc:
@rubenvp8510 @jpkrohling Thanks for your kind guidance and assistance!

PR #1491, opened 2021-06-26
 Fix ingress tests for kuttl 

rubenvp8510:
-  Fixes ingress tests for kuttl

-  The `ingress-nginx`  controller in the `master` branch  has a `ValidatingAdmissionWebhook` for `networking.k8s.io/v1` but the validation aren't migrated yet on ingress controller _0.47.0_ which is the image used by the` deploy.yaml `on `master`  (the validation is already migrated on 1.0.0. See [commit](https://github.com/kubernetes/ingress-nginx/commit/78afe7e389a3b54f5b3b733635c9dc098909ee41#diff-994d32c26c5c0bfa793222a8a5171ed8e4ea78fc850353bb98602b76751f4c2dR47)  ). This produces an error when `jaeger-operator` tried to create an Ingress CR using `networking.k8s.io/v1` API.  It fails with the following message in the log:

`Error: server.go:84] "failed to process webhook request" err="rejecting admission review because the request does not contain an Ingress resource but networking.k8s.io/v1, Kind=Ingress with name`

The solution for now is to use branch `helm-chart-3.33.0` which still haven't has the validation for `networking.k8s.io/v1`. We can move to 1.0 when it is released.

There was also another detail regarding to Ingress:

We forgot one thing when the migration of API work was done, we need to specify the pathType, because is mandatory for 1.22 (https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122)

> pathType is now required for each specified path. Options are Prefix, Exact, and ImplementationSpecific. To match the undefined v1beta1 behavior, use ImplementationSpecific

### DIFF
--- a/pkg/controller/jaeger/configmap.go
+++ b/pkg/controller/jaeger/configmap.go
@@ -10,7 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
+	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
 	"github.com/jaegertracing/jaeger-operator/pkg/inventory"
 	"github.com/jaegertracing/jaeger-operator/pkg/tracing"
 	"github.com/jaegertracing/jaeger-operator/pkg/util"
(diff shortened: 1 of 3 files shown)


jpkrohling:
To clarify, this is caused by a change in the dependent resource, right? Nothing that we did on our side actively broke this.

pkg/controller/jaeger/configmap.go
```diff
@@ -10,7 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
+	v1 "github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1"
```
jpkrohling, line 13: Why is this change here?

jpkrohling (approved):
LGTM, just an unnecessary change is in the PR.

pkg/controller/jaeger/configmap.go
rubenvp8510: not sure , I just did make fmt  and this changed :/

PR #1501, opened 2021-07-07
Include OIDC plugin in binary

esnible:

On OIDC-secured clusters, such as IBM Cloud, the `make run` target fails with `FATA[0000] no Auth Provider found for name "oidc"`

This happens when _.kube/config_ contains `users[*].user.auth-provider.config.name` == `oidc`.  When running within the cluster the OIDC provider isn't needed (on my cloud, with my configuration), but when running outside the cluster it is needed -- otherwise the go-client can't talk to the API server.

Include "k8s.io/client-go/plugin/pkg/client/auth/oidc" in binary.  Even though we don't use the include, because it is there [this line](https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/oidc/oidc.go#L50) allows the code to use OIDC if needed.

I did not create a test case because I don't know how to mock auth providers.

### DIFF
--- a/pkg/cmd/start/bootstrap.go
+++ b/pkg/cmd/start/bootstrap.go
@@ -21,6 +21,9 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/rest"
+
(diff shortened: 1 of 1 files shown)


rubenvp8510 (approved):
This change looks good to me.

jpkrohling (approved):

PR #1508, opened 2021-07-14
Sidecar not removed

### LINKED ISSUE #1497: Sidecar is not removed when annotation is changed to false
**Describe the bug**
Operator doesn't remove sidecar based on annotation, only adds it.

**To Reproduce**
1. Create a deployment with sidecar.jaegertracing.io/inject set to true
2. Spin it up
3. Apply changed deployment with sidecar.jaegertracing.io/inject set to false

**Expected behavior**
Expect the operator to spin down the agent side car

 - Jaeger version: 1.23
 - Deployment:  Kubernetes

**What troubleshooting steps did you try?**
I was able to make it work only when I completely delete the deployment and create it again with the false value. If I do a rollout or delete the pod, it will always spin up with a side car.

**Additional context**

Even when the annotation value is blank or the annotation is removed it still keeps the side car.
on https://www.jaegertracing.io/docs/1.23/operator/#install-modes It says 
"watch all deployments, to inject or remove sidecars based on the sidecar.jaegertracing.io/inject annotation"
It doesn't remove it, so I want my money back >:(

mfz85:
- Resolves #1497 

- validates annotation with "false" value on the deployment in the needed method, and adds a method for removing the sidecar on the deployment controller if it exists and is not needed


pkg/controller/deployment/deployment_controller.go
```diff
@@ -117,6 +117,10 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 
 	if !inject.Desired(dep, ns) {
 		// sidecar isn't desired for this deployment, skip remaining of the reconciliation
```
jpkrohling, line 119: Could you update this comment, to reflect the new logic?

pkg/controller/deployment/deployment_controller.go
```diff
@@ -170,6 +174,25 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 	return reconcile.Result{}, nil
 }
 
+func removeSideCar(r *ReconcileDeployment, dep *appsv1.Deployment, ctx context.Context) {
```
jpkrohling, line 177: Isn't it better to make `removeSidecar` to be an instance method, instead of receiving an instance as a parameter? Also, Go conventions dictate that the context should be the first parameter. This would be the final signature:

```suggestion
func (r *ReconcileDeployment) removeSidecar(ctx context.Context, dep *appsv1.Deployment) {
```

pkg/controller/deployment/deployment_controller.go
```diff
@@ -170,6 +174,25 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
...
+	jaegerInstance, hasLabel := dep.Labels[inject.Label]
+	if hasLabel {
```
jpkrohling, line 179: I feel that if this function has been called, we should remove the sidecar no matter the state of the label. For instance, we might have the label at the namespace level.

pkg/inject/sidecar.go
```diff
@@ -80,10 +80,10 @@ func Desired(dep *appsv1.Deployment, ns *corev1.Namespace) bool {
 		"namespace":  dep.Namespace,
 		"deployment": dep.Name,
 	})
-	_, depExist := dep.Annotations[Annotation]
+	annotationValue, depExist := dep.Annotations[Annotation]
 	_, nsExist := ns.Annotations[Annotation]
 
-	if depExist {
+	if depExist && !strings.EqualFold(annotationValue, "false") {
```
jpkrohling, line 86: What if the annotation is at the namespace level? Isn't this logic already present in the `inject.Desired`?

jpkrohling (commented):
Thank you for the PR! I'll ask @rubenvp8510 to review it as well, but I believe there are a couple of changes necessary. In particular, we should account for the fact that annotations might be at the namespace level.

pkg/inject/sidecar.go
mfz85: inject.Desired just test if the annotation is present, but if the annotation exists with a "false" value it still returns "true".

Should i validate the same condition for the namespace?

jpkrohling: I'm confident we have a logic somewhere that determines whether a deployment should have a sidecar based on annotations from both the deployment itself and the namespace.

mfz85: it's present in "inject.select", but doesn't it make more sense for this to be validated in "inject.desired" ?, it is a bit confusing that desired returns "true" when it is not really desired.

jpkrohling: True. Would you be able to create an issue to refactor that? Also, it would be great to have an e2e test to assert the behavior of this PR here. It doesn't have to be done as part of this PR, though.

I'd still like to get @rubenvp8510's review before merging, but looks good to me!

mfz85: > True. Would you be able to create an issue to refactor that? Also, it would be great to have an e2e test to assert the behavior of this PR here. It doesn't have to be done as part of this PR, though.

count on it

rubenvp8510: This covers both cases,  because `inject.Needed ` is used when we modify the namespace to determine if the deployments that belongs to namespace needs an injection. Same case for individual deployments.

Except that we already have the cleaning logic on namespace  but not on individual deployments (which is something this PR adds )

I think this modification to the desired behaviour makes sense , but it is a little bit confusing the difference between needed and desired,  Indeed `inject.Needed` calls` inject.Desired.` so I think is enough to expose `Needed` and use it on the controllers.

pkg/controller/deployment/deployment_controller.go
```diff
@@ -116,7 +116,15 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 
 	if !inject.Desired(dep, ns) {
```
rubenvp8510, line 118: I would add this cleaning logic on the else statement of the `if inject.Needed(dep, ns) {`

Here: https://github.com/jaegertracing/jaeger-operator/pull/1508/files#diff-c553184997fa4c9f63e697af2f39eb2e6fb43eeff64a0ef66c9e955f61347b02R172

pkg/controller/deployment/deployment_controller.go
```diff
@@ -170,6 +178,23 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileDeployment) removeSidecar(ctx context.Context, dep *appsv1.Deployment) {
```
rubenvp8510, line 181: Is this the same logic that we used for the namespace case? 

Is there a way to do a single function and reuse it for both case? Not need to be part of this PR, but we can have another PR with a re-factoring of this logic.

rubenvp8510:
A couple of comments and one change, I would try to use Needed for all logic on the controller.

This is the logic I think both controllers should have (this is pseudo-code)

```
If inject.Needed(dep, ns) {
  jaeger = inject.Select(....) 
  if jaeger != nil {
   <inject logic> 
  } else {
     log("No suitable instance found...")
  }
} else {
  agent = hasAgent(dep)
  if (agent) {
      cleanSideCar()
  }
}
```

pkg/inject/sidecar.go
rubenvp8510: BTW, I think we shouldn't use select for validating these cases, inject.Select  is just when we are make sure we need to inject something but we need to find which instance we need to inject (and if it exists)

mfz85: the refactor is done, can you review @rubenvp8510 ?

rubenvp8510 (approved):
LGTM!

PR #1517, opened 2021-07-23
Add support for GRPC storage plugin

### LINKED ISSUE #1042: Operator for storage plugin
Hey. I want to deploy Jaeger with Logz.io storage [plugin](https://github.com/logzio/jaeger-logzio), using Kubernetes operator. I can't find a way to do it. Is it supported? if so can you provide any guidance on how to deploy the operator with storage plugin?

pavolloffay:
Resolves #1042 
Supersedes #1045 

How does it work: 

User has to specify `grpc-plugin` as `storage.type` and image with the plugin binary in `storage.grpcPlugin.image`. The image has to copy plugin binary into `/plugin/` directory. At the moment this is not configurable. If there will be requests we could make that configurable in the CR.  
 
The image is used for init container. At jaeger startup the init container copies the plugin binary into volume. This makes the binary visible to Jaeger.

The flags for the storage plugin (bibary, configuration-file, log-level) are provided in `storage.options`, to be consistent with other storage implementations. The configuration file for the storage plugin has to be explicitly configured in storage volumes as a config map (the user is responsible for creating it).

```
kind: Jaeger
metadata:
  name: clickhouse-grpc-plugin
spec:
  storage:
    type: grpc-plugin
    grpcPlugin:
      image: ghcr.io/pavolloffay/jaeger-clickhouse:0.4.1
    options:
      grpc-storage-plugin:
        binary: /plugin/jaeger-clickhouse
        configuration-file: /plugin-config/config.yaml
        log-level: debug
  volumeMounts:
    - name: plugin-config
      mountPath: /plugin-config
  volumes:
    - name: plugin-config
      configMap:
        name: jaeger-clickhouse
```

This PR does not contain e2e tests. The e2e tests will be added in a separate PR.


rubenvp8510 (dismissed):

rubenvp8510:
I think we should add documentation on how this works. (something similar to the description of this PR)

pkg/deployment/ingester_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/require"
```
rubenvp8510, line 8: I think the import order here is not correct

pavolloffay:
re docs- yes definitely I will add them to docs repo.

pkg/deployment/ingester_test.go
pavolloffay: it passes CI check. We should perhaps use different tooling to catch this on CI

PR #1521, opened 2021-08-03
Add gRPC port for jaeger-query into its service resource

### LINKED ISSUE #1520: Add gRPC port for jaeger-query into its Service resource
## Requirement - what kind of business use case are you trying to solve?

I am trying to set up integration via GRPC with Kiali and Jaeger within an Istio-mesh K8s cluster.

- kiali – v1.37.0
- istio – v1.10.3
- jaeger-operator– v1.24.0
  - The CR is a simple AllInOne with no addtional config.

## Problem - what in Jaeger blocks you from solving the requirement?

Kiali from v1.37.0 expects the communication port with Jaeger via RPC (16685).  In Kiali-operator [PR#370](https://github.com/kiali/kiali-operator/pull/370/files). 

It was a pain to debug this but here is the list of problems:

- Pod/SVC of jaeger do not expose 16685 by default.
  ```bash
  NAME                      READY   STATUS    RESTARTS   AGE
  jaeger-64bf49b557-ddtlm   1/1     Running   0          89m
  
  NAME           TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)     AGE
  jaeger-query   ClusterIP   10.101.138.58   <none>        16686/TCP   88m
  ```
- The CRDs do not support adding additional ports for both SVC and Deployment of Jaeger. At least from what I can see in here: [pkg.go.dev/jaegertracing/jaeger-operator](https://pkg.go.dev/github.com/jaegertracing/jaeger-operator/pkg/apis/jaegertracing/v1?utm_source=godoc#pkg-overview)
- There is no easy way to reference a full list of supported CR. And #671 should be the priority for new people to actually see what are available. And not by going through a go code.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

- The easiest way to fix it is: **Creating another SVC to expose 16685. Although Pod do not expose 16685, but it will work.**. I kind of remember that from reading the K8s docs.
- Adding more granular configuration to all the SVC, Deployment of Jaeger.
- Or support kustomize patch manifests folder with raw k8s yaml.

## Any open questions to address

N/A

rubenvp8510:

- Fixes #1520 

-  Added GRPC port to query service

### DIFF
--- a/deploy/crds/jaegertracing.io_jaegers_crd.yaml
+++ b/deploy/crds/jaegertracing.io_jaegers_crd.yaml
@@ -6149,6 +6149,9 @@ spec:
                       type: string
                     nullable: true
                     type: object
+                  grpcNodePort:
+                    format: int32
+                    type: integer
                   image:
                     type: string
                   labels:
--- a/pkg/apis/jaegertracing/v1/jaeger_types.go
+++ b/pkg/apis/jaegertracing/v1/jaeger_types.go
@@ -267,6 +267,10 @@ type JaegerQuerySpec struct {
 	NodePort int32 `json:"nodePort,omitempty"`
 
 	// +optional
+	// NodePort represents the port at which the NodePort service to allocate
+	GRPCNodePort int32 `json:"grpcNodePort,omitempty"`
+
+	// +optional
 	// TracingEnabled if set to false adds the JAEGER_DISABLED environment flag and removes the injected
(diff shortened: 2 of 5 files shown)


jpkrohling (approved):

rafilkmp3:
We need a example how to use this, because jaeger query dont have grpc port enabled by default

rafilkmp3:
image

PR #1544, opened 2021-08-31
Add E2E testing for ES Rollover feature

iblancasa:
## Which problem is this PR solving?
Add E2E testing to the Rollover ES feature implemented in #267

## Short description of the changes
* Add E2E Test for the Rollover ES feature
* Add `index` job assert to work with ES indices
* Add more capabilities to the `reporter` assert job to being able to report more spans in different ways
* Add a cmd to wait until a job scheduled by a cronjob is executed successfully `tests/cmd-utils/wait-cronjob/`. I tried to detect this using the events from the Kubernetes API and Kuttl but [there is a problem](https://github.com/kudobuilder/kuttl/issues/86) and the matching between objects is not correct
* Small changes in `kuttl-test.yaml` to avoid errors trying to recreate the `jaeger-operator-system` Kubernetes namespace

PR #1548, opened 2021-09-03
Add E2E testing for applications running outside the cluster

iblancasa:
Add testing for the use case when there are applications running outside the cluster. The collector needs to be exposed.

* Add E2E testing for the use case where there are applications that need to report spans to the Jaeger instance running in the cluster

*Note:* there is a change introduced in #1544 that can avoid call the `sleep` command in `01-check-collector.yaml`.


tests/e2e/outside-cluster/01-check-collector.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: "JAEGER_SERVICE_NAME=my-test-service OPERATION_NAME=my-little-op JAEGER_ENDPOINT=http://localhost/collector/api/traces go run ../../assert-jobs/reporter/main.go"
+  - script: "sleep 5" # This sleep is needed to ensure the traces from the previous step are available
```
iblancasa, line 5: This will not be needed after #1544 is merged

iblancasa:
@rubenvp8510 can you take a look to this PR? All the tests are passing but I don't know if more changes are required. Thanks! :)

rubenvp8510 (approved):
LGTM!

Issue #1431, opened 2021-04-16
Failed to list namespaces error when not using cluster-wide mode


ediezh:
Deployed the operator 1.22 following these instructions

```
kubectl create namespace observability
kubectl create -n observability -f https://raw.githubusercontent.com/jaegertracing/jaeger-operator/master/deploy/crds/jaegertracing.io_jaegers_crd.yaml
kubectl create -n observability -f https://raw.githubusercontent.com/jaegertracing/jaeger-operator/master/deploy/service_account.yaml
kubectl create -n observability -f https://raw.githubusercontent.com/jaegertracing/jaeger-operator/master/deploy/role.yaml
kubectl create -n observability -f https://raw.githubusercontent.com/jaegertracing/jaeger-operator/master/deploy/role_binding.yaml
kubectl create -n observability -f https://raw.githubusercontent.com/jaegertracing/jaeger-operator/master/deploy/operator.yaml
```

The operator watches only the observability namespace. But I'm seeing these errors in the log:
```
E0416 08:20:50.253032       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.2/tools/cache/reflector.go:125: Failed to list *v1.Namespace: namespaces is forbidden: User "system:serviceaccount:observability:jaeger-operator" cannot list resource "namespaces" in API group "" at the cluster scope
```

jpkrohling:
@ediezh could you please provide the deployment that is being used by the operator?

    kubectl get deployment -n observability jaeger-operator -o yaml

I'm especially interested in the value of the `WATCH_NAMESPACE`.

ediezh:
@jpkrohling here is the deployment.
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubectl.kubernetes.io/restartedAt: "2021-04-16T08:10:16Z"
    kubernetes.io/psp: eks.privileged
    prometheus.io/path: /stats/prometheus
    prometheus.io/port: "15020"
    prometheus.io/scrape: "true"
    sidecar.istio.io/status: '{"version":"8e6e902b765af607513b28d284940ee1421e9a0d07698741693b2663c7161c11","initContainers":["istio-init"],"containers":["istio-proxy"],"volumes":["istio-envoy","istio-data","istio-podinfo","istio-token","istiod-ca-cert"],"imagePullSecrets":null}'
  creationTimestamp: "2021-04-16T08:10:16Z"
  generateName: jaeger-operator-799c7fbbb5-
[...]

PR #1551, opened 2021-09-08
Only reconcile namespaces matching WATCH_NAMESPACE

esnible:

This fixes the problem I brought up in https://github.com/jaegertracing/jaeger-operator/issues/1431#issuecomment-915574074

It doesn't fix the problem of #1431 .  It fixes the problem where the operator reconciles more namespaces than it should, if it has the RBAC permission to do so.

I don't know how to test operators and therefor didn't write any tests.  Please suggest how to test this.

### DIFF
--- a/pkg/controller/namespace/namespace_controller.go
+++ b/pkg/controller/namespace/namespace_controller.go
@@ -2,9 +2,9 @@ package namespace
 
 import (
 	"context"
-	"os"
 	"strings"
 
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	"go.opentelemetry.io/otel"
@@ -34,7 +34,17 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileNamespace{client: mgr.GetClient(), scheme: mgr.GetScheme(), rClient: mgr.GetAPIReader()}
+	watchedNamespace, _ := k8sutil.GetWatchNamespace()
+	watchedNamespaces := []string{}
+	if watchedNamespace != "" {
+		watchedNamespaces = strings.Split(watchedNamespace, ",")
+	}
+	return &ReconcileNamespace{
+		client:            mgr.GetClient(),
+		scheme:            mgr.GetScheme(),
+		rClient:           mgr.GetAPIReader(),
+		watchedNamespaces: watchedNamespaces,
+	}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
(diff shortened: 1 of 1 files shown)


pkg/controller/namespace/namespace_controller.go
```diff
@@ -84,11 +86,28 @@ func (r *ReconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Res
 	ctx, span := tracer.Start(ctx, "reconcileNamespace")
 	defer span.End()
 
+	// Don't reconcile namespace unless unless namespace matches $WATCH_NAMESPACE
+	watchedNamespaces := os.Getenv("WATCH_NAMESPACE")
```
rubenvp8510, line 90: We can parse this on the bootstrap

rubenvp8510: Also I think we should use` k8sutil.GetWatchNamespace()`

pavolloffay:
This might have been described in https://github.com/jaegertracing/jaeger-operator/issues/1431#issuecomment-915574074 but I am asking here again for the explanation. What issue does this PR fixes?

I would assume that `Reconcile` is called only for namespaces that the operator is watching. 


> I don't know how to test operators and therefor didn't write any tests. Please suggest how to test this.

Have a look at https://github.com/jaegertracing/jaeger-operator/tree/master/test/e2e. The test could setup the operator to watch a single namespace and assert that the issues this PR fixes does not exist.

esnible:
> This might have been described in [#1431 (comment)](https://github.com/jaegertracing/jaeger-operator/issues/1431#issuecomment-915574074) but I am asking here again for the explanation. What issue does this PR fixes?

@pavolloffay The namespace controller watches all the namespaces.  This is unexpected behavior if the Jaeger Operator has been started with WATCH_NAMESPACE != ''
https://github.com/jaegertracing/jaeger-operator/blob/master/pkg/controller/namespace/namespace_controller.go#L52

I tested with the command line and see that the operator will reconcile all the namespaces on the cluster.  This includes restarting pods (!), which is unexpected.

```bash
WATCH_NAMESPACE=observability make run-debug
kubectl annotate namespace default sidecar.jaegertracing.io/inject=true
```

esnible:
@rubenvp8510 I have made the changes you asked for, I think.  Let me know if I moved the code correctly.

The call you requested, `k8sutil.GetWatchNamespace()` ... I don't believe it is present in the current Operator SDK.  I believe we are on a year-old operator.  Is that safe?  Is there an issue to get on the current code?

esnible:
@pavolloffay Do we have instructions on how to run/modify the e2e tests?

The e2e test I need to write would
1. Create two namespaces, test-a and test-b, both with `sidecar.jaegertracing.io/inject=true`
2. Create a Deployments in each namespace (neither with a sidecar)
3. Start the operator with WATCHED_NAMESPACE=test-a
4. Verify the test-a gets the sidecar, but test-b does not.
5. Cleanup

I can see how to make the K8s calls, with `fw.Client.Create()` and `e2eutil.WaitForDeployment()`, but how do I start the operator with `WATCH_NAMESPACE=test-a`?

PR #1555, opened 2021-09-14
Add matrix for test operator with k8s 1.22

rubenvp8510:


.ci/minikube-ingress.yaml
```diff
@@ -0,0 +1,639 @@
+# Copyright 2021 The Kubernetes Authors All rights reserved.
```
pavolloffay, line 1: Why do we need to install the ingress?

rubenvp8510: The reason is because the ingress add-on now requires to have all ingress annotated,  but the `nginx` ingress controller (which is the ingress `minikube` is using) provides a way to disable that behaviour in the recent versions. unfortunately `minikube` ingress controller is still not updated, and there is no way to use that feature, the only way I found is provisioning my own ingress controller.

https://kubernetes.github.io/ingress-nginx/#i-have-only-one-instance-of-the-ingresss-nginx-controller-in-my-cluster-what-should-i-do

I filed an issue here: https://github.com/kubernetes/minikube/issues/12511

pavolloffay (approved):

PR #1569, opened 2021-10-04
Allow sidecar injection for query pod from other Jaeger instances

pavolloffay:
This PR allows injecting jaeger-agent sidecar from another Jaeger instance to jaeger-query. 

If the tracing is enabled in jaeger-query deployment the operator adds an annotation `sidecar.jaegertracing.io/inject=<current-jaeger-instance>` to a created common spec object. The annotation is merged with annotations from the CR which allows overriding the instance name e.g.:

```yaml
kind: Jaeger
metadata:
  name: jaeger-tenant1
  namespace: rhosdt-tenant1
spec:
  strategy: production
  query:
    tracingEnabled: true
    annotations:
      sidecar.jaegertracing.io/inject: telemeter-jaeger
  storage:
    type: elasticsearch
    options:
      es:
        max-span-age: 168h # should match numberOfDays*24
    esIndexCleaner:
      enabled: true
      numberOfDays: 7
      schedule: "55 23 * * *"
    elasticsearch:
      nodeCount: 1
````

### DIFF
--- a/pkg/strategy/production.go
+++ b/pkg/strategy/production.go
@@ -120,9 +120,6 @@ func newProductionStrategy(ctx context.Context, jaeger *v1.Jaeger) S {
 	// prepare the deployments, which may get changed by the elasticsearch routine
 	cDep := collector.Get()
 	queryDep := inject.OAuthProxy(jaeger, query.Get())
-	if jaeger.Spec.Query.TracingEnabled == nil || *jaeger.Spec.Query.TracingEnabled == true {
-		queryDep = inject.Sidecar(jaeger, queryDep)
-	}
 	c.dependencies = storage.Dependencies(jaeger)
 
 	// assembles the pieces for an elasticsearch self-provisioned deployment via the elasticsearch operator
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/strategy/production.go [hunk 1]
```diff
@@ -120,9 +120,6 @@ func newProductionStrategy(ctx context.Context, jaeger *v1.Jaeger) S {
 	// prepare the deployments, which may get changed by the elasticsearch routine
 	cDep := collector.Get()
 	queryDep := inject.OAuthProxy(jaeger, query.Get())
-	if jaeger.Spec.Query.TracingEnabled == nil || *jaeger.Spec.Query.TracingEnabled == true {
```
rubenvp8510, line 123: Why did you remove this `TracingEnabled` if?

pavolloffay: now the injection will be done purely via the deployment controller. So by removing this check I have to allow injection to `query` component in the deployment controller.

If this check is here the agent gets injected here with the current jaeger instance. I need to go throguh the deploymnt injection logic to allow injection via arbittrary jaeger instance.

rubenvp8510: ok, make sense.

rubenvp8510 (approved):

PR #1587, opened 2021-10-18
Migrate examples 2 E2E test to use KUTTL

iblancasa:
- Migrate the E2E examples 2 tests to use KUTTL.


rubenvp8510:
Is this ready for review?

Makefile
```diff
@@ -542,7 +595,7 @@ prepare-e2e-kuttl-tests: build docker build-assert-job
 kuttl-e2e: prepare-e2e-kuttl-tests start-kind run-kuttl-e2e
 
 .PHONY: run-kuttl-e2e
-run-kuttl-e2e:
+run-kuttl-e2e: kafka
```
rubenvp8510, line 598: I think this should be  part of test prepare target.

iblancasa: Sure! I fixed this for #1598 but I didn't have the chance to apply the change here

tests/e2e/examples-simple-streaming/00-install.yaml
```diff
@@ -0,0 +1,5 @@
+# Start Kafka
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: "pwd && cd ../../.. && make kafka"
```
rubenvp8510, line 5: This test if fialing I think we need to change this to 

`  - script: " KAFKA_NAMESPACE=$NAMESPACE make kafka"`

rubenvp8510 (changes requested):

rubenvp8510 (approved):

PR #1606, opened 2021-11-02
Migrate ElasticSearch E2E tests to KUTTL

iblancasa:
- Migrate the ElasticSearch E2E tests to KUTTL


tests/e2e/es-index-cleaner/02-generate-spans.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: Job
+metadata:
+  name: report-span-1
+spec:
+  template:
+    spec:
+      containers:
+        - name: report-span-1
+          image: local/asserts:e2e
+          command: ["./reporter",
```
rubenvp8510, line 12: nit: I think is more readable to have this in another line, something like this:
```
[
   "./reporter",
   "--days", "5",
   "--services", "1",
    "--verbose"
]
```

tests/e2e/es-index-cleaner/06-check-indices.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: Job
+metadata:
+  name: assert-indexes-no-prefix
+spec:
+  template:
+    spec:
+      containers:
+        - name: assert-indexes-no-prefix
+          image: local/asserts:e2e
+          command: ["./index",
```
rubenvp8510, line 12: Same comment as above.

tests/e2e/es-index-cleaner/es-index.template
```diff
@@ -0,0 +1,5 @@
+      index-prefix: ""
```
rubenvp8510, line 1: As far as I understand you use this for adding more things at the end of another `YAML` file, 

Can we use conditions on our templates? I think that would make things more readable. I did something like that in this template:

https://github.com/jaegertracing/jaeger-operator/blob/master/tests/templates/cassandra-jaeger-install.yaml.template#L10

rubenvp8510 (changes requested):

rubenvp8510 (approved):

PR #1608, opened 2021-11-03
Use CRDs to detect features in the cluster

pavolloffay:


pkg/autodetect/main.go
```diff
@@ -109,16 +108,11 @@ func (b *Background) autoDetectCapabilities() {
...
 
-func (b *Background) availableAPIs(ctx context.Context) (*metav1.APIGroupList, error) {
-	apiList, err := b.dcl.ServerGroups()
-	if err != nil {
-		return nil, err
-	}
-
-	return apiList, nil
+func (b *Background) availableAPIs(_ context.Context) ([]*metav1.APIResourceList, error) {
+	return b.dcl.ServerPreferredResources()
```
pavolloffay, line 112: `b.dcl.ServerGroups()` returns only the API groups and for instance `logging.openshift.io` has multiple kinds: `elasticsearch`, `kibana` ... Hence we should enable ES provisioning only if the `elasticsearch` kind is in the cluster.

rubenvp8510:
Code Looks good, just one question

Why for this particular GroupVersion `logging.openshift.io` we need to look fo `elasticsearch` and in the others, e.g Kafka, we don't need it?

rubenvp8510 (approved):

pavolloffay:
bc the vanilla OCP cluster already has some Kids in `logging.openshift.io`.

PR #1610, opened 2021-11-03
Improve robustness of E2E Golang tests

iblancasa:
- Improve robustness of E2E Golang tests


test/e2e/all_in_one_test.go
```diff
@@ -95,38 +92,11 @@ func (suite *AllInOneTestSuite) TestAllInOneWithIngress() {
...
 
-		for _, v := range resp.Data {
-			if v == "jaeger-query" {
-				return true, nil
-			}
-		}
-
-		return false, nil
-	})
+	// We just need to check there ere not errors in the HTTP response and the REST API is available
```
rubenvp8510, line 99:
```suggestion
	// We just need to check there are not errors in the HTTP response and the REST API is available
```

test/e2e/wait_util.go
```diff
@@ -232,3 +237,82 @@ func WaitForSecret(secretName, secretNamespace string) {
...
+
+	err = wait.Poll(retryInterval, timeout*2, func() (done bool, err error) {
+		res, err := httpClient.Do(req)
+
+		if res != nil {
+			logrus.Infof("response status code: %d", res.StatusCode)
+		}
+
+		if err != nil {
+			if strings.Contains(err.Error(), "Timeout exceeded") {
```
rubenvp8510, line 262: I think there is a better way to do this, checking the error type
You can do:

```
if e,ok := err.(net.Error); ok && e.Timeout() {
    // This was a timeout
} else if err != nil {
    // This was an error, but not a timeout
}
```
https://pkg.go.dev/net#Error

test/e2e/all_in_one_test.go
iblancasa: Good catch!

test/e2e/wait_util.go
iblancasa: Oh, really interesting. I'll apply this. Thanks!

rubenvp8510 (approved):

PR #1635, opened 2021-11-18
Update readme to point to jaeger docs for install instructions

rubenvp8510:


pavolloffay (commented):
Any reason for adding the `deploy/jaeger-operator.yaml` I would prefer to generate the file only on release and upload it to the release artifacts.

rubenvp8510:
> Any reason for adding the `deploy/jaeger-operator.yaml` I would prefer to generate the file only on release and upload it to the release artifacts.

The only reason for do it for  now is because we haven't release any new version yet

pavolloffay (approved):

PR #1645, opened 2021-11-29
Add operator-sdk scorecard bundle test to CI. Closes #1565

### LINKED ISSUE #1565: Add operator-sdk scorecard bundle test on CI
As part of our CI we should be testing SDK scorecard. 

However, first we should update the SDK to the latest version.

iblancasa:

- Fixes #1565


iblancasa:
Hi @pavolloffay, could you take a look to into this PR? Thanks!

The CI is failing because some tests are not passing so I think we have 3 options:
1. Remove the tests that are not passing and add a note explaining why do we just run some  tests
2. Remove the tests that are not passing and create issues to make them pass
3. Block this PR until all the tests pass

Sorry, I don't have a lot of background about this tests and if we should fix them or not 😅  Thanks!

pavolloffay (commented):
Overall it looks good to me! Thanks for picking this up ❤️ 

A couple of notes:
* The PROJECT file is missing scorecard plugin https://github.com/jaegertracing/jaeger-operator/blob/master/PROJECT#L5
* I would suggest not blocking this PR and merge it with passing the most important tests: `olm-crds-have-validation-test` ,` olm-bundle-validation-test`, `basic-check-spec-test`. The basic test should really pass. It concerns me that we didn't migrate the project to 1.13 correctly.

rubenvp8510:
> Overall it looks good to me! Thanks for picking this up heart
> 
> A couple of notes:
> 
>     * The PROJECT file is missing scorecard plugin https://github.com/jaegertracing/jaeger-operator/blob/master/PROJECT#L5
> 
>     * I would suggest not blocking this PR and merge it with passing the most important tests: `olm-crds-have-validation-test` ,` olm-bundle-validation-test`, `basic-check-spec-test`. The basic test should really pass. It concerns me that we didn't migrate the project to 1.13 correctly.

Why do you think we didn't migrate the project correctly?

rubenvp8510:
I fixed some metadata things from the CSV, now the scoreboard tests are passing.

.github/workflows/base-checks.yaml
```diff
@@ -8,13 +8,13 @@ on:
 
 jobs:
   basic-checks:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
rkukura, line 11: Why did we need to pin the version here, but use ubuntu-latest below in sdk-scorecard.yaml?

rubenvp8510: I think we need to use the same version for all our GH actions, My understanding is that it is usually a good practice to pin the versions. We can do it in a separate PR.

rkukura: I see there is currently a mix of ubuntu-latest and ubuntu-20.04 in various files. We can make these consistent in a separate PR.

rkukura (approved):
LGTM

rubenvp8510 (approved):

PR #1647, opened 2021-11-30
Enable operator prometheus metrics

rubenvp8510:

- Create Service / ServiceMonitor for the exposed metrics.

-  With the operator-sdk upgrade, we are now have to create the service and service monitor for our Prometheus metrics endpoints. This has to be done as part of the manifest.

In this PR I'm enabling the configurations for do that


config/rbac/prometheus_role.yaml
```diff
@@ -0,0 +1,19 @@
+
```
iblancasa, line 1: This line is not needed, right?

kuttl-test.yaml
```diff
@@ -4,6 +4,7 @@ crdDir: ./tests/_build/crds/
 artifactsDir: ./tests/_build/artifacts/
 commands:
   - script: kubectl create namespace jaeger-operator-system 2>&1 | grep -v "already exists" || true
+  - command: kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/release-0.9/manifests/setup/prometheus-operator-0servicemonitorCustomResourceDefinition.yaml
```
iblancasa, line 7: Instead of doing this, would it make sense to create a Make target called `deploy-prometheus-operator`? Something similar to what we do with `kafka`

config/rbac/kustomization.yaml
```diff
@@ -16,3 +16,5 @@ resources:
 - auth_proxy_role.yaml
 - auth_proxy_role_binding.yaml
 - auth_proxy_client_clusterrole.yaml
+- prometheus_role.yaml
+- prometheus_role_binding.yaml
```
iblancasa, line 20: Missing new line at the end of the file

iblancasa (changes requested):

kuttl-test.yaml
rubenvp8510: Well, I don't required the operator here, only the CRD, It will make sense when we have tests about the prometheus use case.

config/rbac/prometheus_role.yaml
rubenvp8510: not at all

kuttl-test.yaml
iblancasa: Yes but doing this we:
- Allow the developers to install the Prometheus Operator in their environments if they need it (just for development purposes, not running the E2E tests)
- Allow skipping the installation of the Prometheus Operator if we want to use the provided by OLM

iblancasa (approved):

pavolloffay (approved):

PR #1650, opened 2021-12-03
Disable metrics on generated manifest

rubenvp8510:

I had to disable metrics by default because if not the users trying to install the operator will be forced to install prometheus operator.


rkukura:
LGTM

PR #1673, opened 2021-12-15
Register oschema for openshift resources 

rubenvp8510:
- During the migration to operator-sdk 1.13 we miss to add some resource schemas. this PR fixes that

- Register openshift , images and console link resources

@rkukura  Could you please review? Thanks!


rkukura (approved):
LGTM

Issue #1232, opened 2020-10-02
Inject the sidecard directly in Pod resources


fdimuccio:
Hello,

we are using the jaeger-operator together with [kubefed](https://github.com/kubernetes-sigs/kubefed) and we are facing an reconciliation loop issue when using `sidecar.jaegertracing.io/inject` annotation. Actually the issues are two but they stem from the same root cause. If I'm not mistaken, jeager-operator modify the Deployment resource where the annotation is defined by adding the sidecar container alongside the already defined containers. If another controller is managing the Deployment (like in our case the kubefed controller) the resource will be reconciled by removing the sidecar injected by jaeger-operator, ending up in a loop where both operators modify the resource in turns.

Other systems like [istio](https://istio.io/) instead inject the sidecard directly into the Pod object (without modifying the Deployment) in this way they avoid clashing with other controllers managing the same resource. Note that also the annotation definition is moved to the Pod template definition within the Deployment `spec.template.metadata.annnotations`.

I see that there two other github issues potentially connected with this one (#171 and #750), but it's not clear to me if this change is already implied by switching to a Mutating Webhook Admission Controller.

jpkrohling:
> it's not clear to me if this change is already implied by switching to a Mutating Webhook Admission Controller.

Yes, once we move to a webhook, we'll inject sidecars directly into pods.

vassilvk:
Just wanted to drop in a note here too - while #171 is being worked on, an alternative non-destructive injection solution is now available through KubeMod (as [described in my comment](https://github.com/jaegertracing/jaeger-operator/issues/171#issuecomment-711422072)).

Issue #171, opened 2019-01-18
Inject sidecar via Mutating Webhook Admission Controller


jpkrohling:
As hinted by @gregoryfranklin on #60, we should aim to inject the sidecars via [Mutating Webhook Admission Controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook). I believe I tried this during the PoC and couldn't get it to work, but it would be good to try again and, if not possible, document the reasons why we have it the way it is right now.

gregoryfranklin:
If it helps, various examples can be found here: https://github.com/kubernetes/kubernetes/tree/master/test/images/webhook

The biggest issue I've personally run into when working with Webhooks is managing the SSL certs.  The webhook must listen on port 443 and the apiserver must trust the cert (MutatingWebhookConfiguration allows you to provide a ca cert).  Both cert-manager and istio deal with this problem by generating their own ca chain (ie self-signed certs) and having a job that patches the MutatingWebhookConfiguration (see https://github.com/munnerz/apiextensions-ca-helper).

jpkrohling:
The operator-sdk is finally adding support for admission webhooks, which should allow us to finally work on this in an appropriate manner.

https://github.com/operator-framework/operator-sdk/pull/2729

JessieAMorris:
Any word on this? We're getting weird interactions with the jaeger sidecar injector and our LimitRange for default resources. Overall the current injector doesn't seem to work very well and we're running into issues with it.

jpkrohling:
I'll see if we can improve the current injector, to make it more reliable. It would help us immensely if you could describe concretely what you are facing.

The sidecar injection via mutating webhooks has been implemented in the opentelemetry-operator, which was mainly done to test waters and see if we can apply the same technique here. The good news is that it got merged today (https://github.com/open-telemetry/opentelemetry-operator/pull/52), the bad news is that the changes required to get it working were quite big: basically, the operator-sdk 0.19.x is a major rewrite from 0.18.x that we are using, and requires changing pretty much the whole operator's foundations. I'm not sure that's feasible to accomplish in the short-term.

jpkrohling:
@JessieAMorris are you using the operator v1.20? There were some changes made by @rubenvp8510 that should have made it more stable.

vassilvk:
Hi @JessieAMorris, @jpkrohling,

[...]

PR #1679, opened 2021-12-17
Use mutating webhook to inject sidecar

lujiajing1126:
- #171 
- Also #1232 

Based on [the idiom of istio injection policy](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy), 

| Resource  | Label | Enabled value | Disabled value |
| ------------- | ------------- | ------------- | ------------- |
| Namespace | jaeger-injection | enabled | disabled |
| Pod | sidecar.jaertracing.io/inject | "true" | "false" |

This idiom could safely prevent from [bootstrapping problem](https://book.kubebuilder.io/cronjob-tutorial/running-webhook.html#the-bootstrapping-problem) supported from Kubernetes v1.15+

- `CertManager` should be installed to inject CA certification as is sugggested by [`kubebuilder-book`](https://book.kubebuilder.io/cronjob-tutorial/cert-manager.html)

## Tasks

- [x] Add webhook and minifests
- [x] E2E tests: TBD

## To be discussed

1. Shall we totally remove namespace controller and deployment controller or just introduce this feature as an experimental thing in the next few releases keeping two kinds of injection approaches in parallel?
2. Annotation based injection is not currently supported by this webhook.


lujiajing1126:
Though more test cases are needed and issue #1692 has to be resolved first, generally ready to be reviewed.

Request review @jpkrohling @rubenvp8510

jpkrohling:
@rubenvp8510 and @pavolloffay would be the appropriate reviewers for this.

rubenvp8510:
I would like to get rid of namespace and deployment controllers at some point, but the only concern I have with this is it seems to be a breaking change from what the users are doing now to inject the sidecar.

Any reason to move it to Pod?  Is because it provides more flexibility?

.github/workflows/e2e.yaml
```diff
@@ -199,6 +199,33 @@ jobs:
         env:
           KUBE_VERSION: ${{ matrix.kube-version }}
         run: make run-e2e-tests-outside-cluster
+  sidecard-e2e-tests:
+    name: Smoke E2E tests
```
rubenvp8510, line 203: Is this name correct? Should we call this Smoke sidecar e2e tests?

config/webhook/component_inject_patch.yaml
```diff
@@ -0,0 +1,16 @@
...
+    objectSelector:
+      matchExpressions:
+        - key: app
+          operator: In
+          values:
+            - "jaeger"
+        - key: app.kubernetes.io/component
+          operator: NotIn
+          values:
+            - "query"
```
rubenvp8510, line 16: new line

config/webhook/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - manifests.yaml
+  - service.yaml
+
+configurations:
+  - kustomizeconfig.yaml
+
+patchesStrategicMerge:
+  - namespace_inject_patch.yaml
+  - pod_inject_patch.yaml
+  - component_inject_patch.yaml
```
rubenvp8510, line 11: add new line at the end

config/webhook/namespace_inject_patch.yaml
```diff
@@ -0,0 +1,20 @@
...
+          values:
+            - "enabled"
+    objectSelector: # also skip pods that have been already injected (prevent from being injected twice by multiple operators)
+      matchExpressions:
+        - key: sidecar.jaegertracing.io/inject
+          operator: NotIn
+          values:
+            - "false"
+        - key: sidecar.jaegertracing.io/injected # set by deployment controller
+          operator: DoesNotExist
```
rubenvp8510, line 20: new line

lujiajing1126:
> I would like to get rid of namespace and deployment controllers at some point, but the only concern I have with this is it seems to be a breaking change from what the users are doing now to inject the sidecar.

I cannot agree more about this, "namespace and deployment controllers" have to be removed.

After migrating to the pod mutating webhook, the current injected sidecar will not be removed. Mutating webhook just provides a new way to do so which is orthogonal to the previous one. As I understand, it is the backwards compatibility guarantee.

> Any reason to move it to Pod? Is because it provides more flexibility?

I suppose there are multiple reasons, the most important of which is stable. We have met several issues in the production cluster with deployment/namespace reconciliation,

1. Issue #1696 is caused by deployment automatic reconciliation. This issue made surge in our production cluster twice, leading to thousands of pods creation within a small interval.
2. Issue #1318 is also caused by deployment reconciliation which had been fixed later.

Another issue is the upgrade: performing upgrade for the operator has to be extremely careful since it will update all deployments to inject new sidecar. It would be very scary in a relatively large cluster.

rubenvp8510:
@lujiajing1126  Sorry for been late on this, but this PR: https://github.com/jaegertracing/jaeger-operator/pull/1718 has been merged already, I think you needed in order to test this, right?

frzifus:
hi @lujiajing1126, would you like to continue working on this topic? alternatively, i would like to continue your work.

lujiajing1126:
> hi @lujiajing1126, would you like to continue working on this topic? alternatively, i would like to continue your work.

I am too busy to proceed on this PR. Sure please go ahead!

config/default/webhookcainjection_patch.yaml
```diff
@@ -0,0 +1,15 @@
+# This patch add annotation to admission webhook config and
+# the variables $(CERTIFICATE_NAMESPACE) and $(CERTIFICATE_NAME) will be substituted by kustomize.
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: mutating-webhook-configuration
+  annotations:
+    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
+#---
```
iblancasa, line 9: Is it OK to keep this commented code?

pkg/inject/sidecar.go
```diff
@@ -19,10 +19,17 @@ import (
...
+	// Only "enabled" is the valid value for namespace-scoped injection
+	NamespaceLabel = "jaeger-injection"
+	// Annotation is the annotation name to look for when deciding whether to inject
 	Annotation = "sidecar.jaegertracing.io/inject"
-	// Label is the label name the operator put on injected deployments.
+	// Label is the label name the operator put on injected deployments or pods.
 	Label = "sidecar.jaegertracing.io/injected"
+	// PodLabel is the label name on the pod level to look for when deciding whether to inject
+	// This is required and recommended approach to make use of mutating-webhook
+	// TODO: support sidecar injection via annotation-level declaration like istio
```
iblancasa, line 31: Create an issue to address it in the future and add here the issue ID.

iblancasa (changes requested):

rubenvp8510:
I'll close this PR , as was superseded by https://github.com/jaegertracing/jaeger-operator/pull/1828

Thanks for this PR which was used as the base for the webhook work!

Issue #1684, opened 2021-12-20
Jaeger operator is spamming logs with "Waited for Xs due to client-side throttling, not priority and fairness"


rakesh-garimella:
**Describe the bug**

I tried to upgrade the Jaeger operator to the version 1.29.1 on kubernetes. I used the deployment stratergy of `allinone`. When I check the logs of the jaeger operator I see that following logs appear almost evert second
```
I1217 12:10:20.274563       1 request.go:668] Waited for 1.540277s due to client-side throttling, not priority and fairness, request: GET:https://10.43.0.1:443/apis/apiextensions.k8s.io/v1?timeout=32s
I1217 12:10:30.274983       1 request.go:668] Waited for 1.5175147s due to client-side throttling, not priority and fairness, request: GET:https://10.43.0.1:443/apis/networking.k8s.io/v1?timeout=32s
I1217 12:10:40.325410       1 request.go:668] Waited for 1.577647s due to client-side throttling, not priority and fairness, request: GET:https://10.43.0.1:443/apis/node.k8s.io/v1?timeout=32s
I1217 12:10:54.740491       1 request.go:668] Waited for 1.0375806s due to client-side throttling, not priority and fairness, request: GET:https://10.43.0.1:443/apis/events.k8s.io/v1?timeout=32s
```

I am upgrading Jaeger operator from 1.18.1 to 1.29.1. This behaviour was not seen in jaeger operator 1.18.1.

I think the autodetect feature of jaeger operator is trying to look for all the crds present on the kubernetes cluster. May be this [PR](https://github.com/jaegertracing/jaeger-operator/pull/1608/files) might have caused the issue. 

Another difference to the charts I see is now jaeger [crd(version 1.29.1)](https://github.com/jaegertracing/helm-charts/blob/jaeger-operator-2.27.1/charts/jaeger-operator/crds/crd.yaml) is quite exhaustive (compared to [crd](https://github.com/jaegertracing/helm-charts/blob/jaeger-operator-2.16.0/charts/jaeger-operator/crds/crd.yaml) of 1.18) can this also lead to this issue ?

Is there someway to supress these messages as they dont look something serious.

**To Reproduce**
1. Install jaeger operator with 1.18.1 (strategy (`allinone`))
2. Upgrade to latest Jaeger operator (1.29.1)
3. Check the logs of the jaeger operator

**Expected behavior**
The operator logs should be clean.

**Screenshots**

**Version (please complete the following information):**
[...]

PR #1712, opened 2022-01-19
Avoid calling k8s api for each resource kind on the cluster

rubenvp8510:

- This PR should fixes [Issue-1684](https://github.com/jaegertracing/jaeger-operator/issues/1684) or at least alleviate the problem, The message of _"Waited for Xs due to client-side throttling, not priority and fairness"_ is caused because a lot of API calls to K8s. It seems like calling  `PreferedServerResources` does a lot of calls to the API when the cluster has a lot of different resource Kinds.

- To avoid this, in this PR we ask for `ServerGroups`  and then only get the resources for our interested Groups.

Note: I still see one line in the log with the error "Waited Xs.." but one single time, I don't see the same error repeating in logs each N seconds. so I think this can be considered a fix.

### DIFF
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -22,7 +22,7 @@ import (
 	"github.com/jaegertracing/jaeger-operator/pkg/inject"
 )
 
-var listenedGroupsMap = map[string]struct{}{"logging.openshift.io": struct{}{}, "kafka.strimzi.io": struct{}{}, "route.openshift.io": struct{}{}}
+var listenedGroupsMap = map[string]bool{"logging.openshift.io": true, "kafka.strimzi.io": true, "route.openshift.io": true}
 
 // Background represents a procedure that runs in the background, periodically auto-detecting features
 type Background struct {
@@ -113,8 +113,7 @@ func (b *Background) autoDetectCapabilities() {
 }
 
 func (b *Background) isInListenedGroups(group metav1.APIGroup) bool {
-	_, exists := listenedGroupsMap[group.Name]
-	return exists
+	return listenedGroupsMap[group.Name]
 }
 
 func (b *Background) availableAPIs(_ context.Context) ([]*metav1.APIResourceList, error) {
--- a/pkg/autodetect/main.go
+++ b/pkg/autodetect/main.go
@@ -21,6 +21,8 @@ import (
 	"github.com/jaegertracing/jaeger-operator/pkg/inject"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/autodetect/main.go
```diff
@@ -109,8 +111,29 @@ func (b *Background) autoDetectCapabilities() {
 	b.cleanDeployments(ctx)
 }
 
+func (b *Background) isInListenedGroups(group *metav1.APIGroup) bool {
+	for _, groupName := range groupsListened {
```
pavolloffay, line 115: `map` might be more efficient here. The code checks if a given group should be handled or not.

pkg/autodetect/main.go
```diff
@@ -109,8 +111,29 @@ func (b *Background) autoDetectCapabilities() {
 	b.cleanDeployments(ctx)
 }
 
+func (b *Background) isInListenedGroups(group *metav1.APIGroup) bool {
```
pavolloffay, line 114: passing group by value seems better here

pkg/autodetect/main.go
```diff
@@ -109,8 +111,29 @@ func (b *Background) autoDetectCapabilities() {
...
+
 func (b *Background) availableAPIs(_ context.Context) ([]*metav1.APIResourceList, error) {
-	return b.dcl.ServerPreferredResources()
+	apiLists := []*metav1.APIResourceList{}
+	groupList, _ := b.dcl.ServerGroups()
+	for _, sg := range groupList.Groups {
+		if b.isInListenedGroups(&sg) {
+			groupAPIList, err := b.dcl.ServerResourcesForGroupVersion(sg.PreferredVersion.GroupVersion)
+			if err != nil {
+				return apiLists, err
```
pavolloffay, line 130: I would either return empty `apiList` or accumulate errors and return the error with gathered groups once all calls are done.

pkg/autodetect/main.go
```diff
@@ -21,6 +22,8 @@ import (
 	"github.com/jaegertracing/jaeger-operator/pkg/inject"
 )
 
+var listenedGroupsMap = map[string]struct{}{"logging.openshift.io": struct{}{}, "kafka.strimzi.io": struct{}{}, "route.openshift.io": struct{}{}}
```
pavolloffay, line 25: `map[string]bool` would be better here 

```
func (b *Background) isInListenedGroups(group metav1.APIGroup) bool {
	return listenedGroupsMap[group.Name]
}
```

pavolloffay (approved):

PR #1713, opened 2022-01-19
First call of autodetect should be synchronous

rubenvp8510:

- Other initialization routines relies on platform value to do certain detection, in this particular case the oauth image was not detected correctly.

- We need to make sure to detect the platform before doing any other initialization. In order to do that I made the first call of auto-detect synchronous (other calls will happens as usual, each 5 sec) . I think we should not have problems with this, but if this is a problem we can at least change the platform detection to be sync on the first call.


rubenvp8510:
@pavolloffay Could you please review? it is a little bit hard to reach the coverage. Let me know if this is an issue I can try to fix it.

pavolloffay (approved):

Issue #1556, opened 2021-09-16
Add the ability to the KUTTL tests to be run in a real cluster


iblancasa:
## Requirement - what kind of business use case are you trying to solve?
Add the ability to the [KUTTL](https://kuttl.dev/) tests to be run in a real cluster

## Problem - what in Jaeger blocks you from solving the requirement?
With the current automation in the `Makefile`, it is not possible to run the KUTTL tests in a real cluster. The KUTTL tests are run in a [kind](https://kind.sigs.k8s.io/) cluster but it would be really nice if we allow running them in other clusters like [OpenShift](https://www.redhat.com/es/technologies/cloud-computing/openshift) and [minikube](https://minikube.sigs.k8s.io/).

Some things to solve:
* Add a way to push the `asserts:e2e` Docker image to a Docker registry to allow the cluster to pull it
* Allow the modification of the `YAML` files that are part of the KUTTL tests to point to the correct images (for instance, point to the pushed `asserts:e2e` Docker image)
* Avoid running the `start-kind` `make` target when running the tests again a real cluster

## Proposal - what do you suggest to solve the problem or improve the existing situation?
It would be needed to add a new `make` target to push the `asserts:e2e` Docker image to a Docker registry.

Also, I think it would be a good idea to templatize the `KUTTL` test `YAML`
files. This can be really helpful because we have some situations where it can be interesting:
* Changing easily where the Docker images are hosted to pull them
* We run the same tests for different configurations. Since there is not a way to create "`functions`" in `YAML`, this solution would avoid duplicating `YAML` files
* Copy the files that are the same for all the configurations (like an Elasticsearch configuration, for instance)
* We can add logic to generate a test or not depending on the platform (`OpenShift`, `Kubernetes`, `minikube`). I was not able to find a way to implement this logic directly with KUTTL.

iblancasa:
Please, I would like to work on this task :)

jpkrohling:
kuttl does support running the tests in different envs:

https://kuttl.dev/docs/testing/test-environments.html

iblancasa:
@jpkrohling I know but, currently, our automation doesn't. It is prepared to just run the test in the `kind` cluster. The idea is to allow easily to run in a different environment. One of the reasons is the `asserts:e2e` Docker image is not pushed anywhere and is loaded directly in the `kind` cluster. So, doing the modifications I suggest in the first comment, we can run the testes in all the environments supported by KUTTL.

PR #1715, opened 2022-01-19
Port E2E KUTTL tests to OpenShift

iblancasa:
- #1556: Port some E2E tests to work with OpenShift

- Add macros to render common operations in the tests
- Use a real name for the container image of the Jaeger Operator to allow pushing it to registries
- The `istio` test was missing some steps
- Add the `yq` CLI as a dependency to modify YAML files [(link)](https://github.com/mikefarah/yq)
- Increase E2E assert jobs robustness
- Change the base image of the `E2E assert jobs` image to have some tools like `curl`
- Render the templates under a `_build` folder. This reduces the problems associated to overwrite files and allows skipping tests if they are not supported
- Port some tests to work with OpenShift:
    - Create a way to generate or get secrets to be able to query the `query` endpoint when using OpenShift
    - Create new templates for the OpenShift specific objects
- Avoid rebuilding some of the Docker images if not needed


frzifus:
Hi @iblancasa, could we squash your commits a bit? Would make it esier to follow what changed. :)

iblancasa:
Sure! @frzifus

tests/e2e/Makefile
```diff
@@ -1,37 +1,64 @@
...
 .PHONY: prepare-e2e-images
-prepare-e2e-images:
-prepare-e2e-images: docker build-assert-job
-	$(VECHO)docker pull jaegertracing/vertx-create-span:operator-e2e-tests
-	$(VECHO)docker pull docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.6
+prepare-e2e-images: docker push build-assert-job docker-e2e-upgrade-image set-assert-e2e-img-name
+ifeq ($(USE_KIND_CLUSTER),true)
+	$(MAKE) cert-manager
+	$(ECHO) Loading the Jaeger Operator Docker image in the KIND cluster...
+	$(VECHO)kind load docker-image ${IMG}
```
frzifus, line 40:
```suggestion
	$(VECHO) $(KIND) load docker-image ${IMG}
```

tests/e2e/allinone/kuttl-test.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kuttl.dev/v1beta1
```
frzifus, line 1: Should be removed ^^

tests/e2e/examples/kuttl-test.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kuttl.dev/v1beta1
```
frzifus, line 1: this too :)

frzifus (commented):
As discussed, ive marked a few things that may be repeated. But i would recommend that someone else has a look too. 

Maybe @rubenvp8510 ?

tests/assert-jobs/index/main.go
```diff
@@ -28,25 +26,26 @@ const (
...
 	flagJaegerService      = "jaeger-service"
+	flagCertificatePath    = "certificate-path"
 	flagVerbose            = "verbose"
 )
 
 func filterIndices(indices *[]elasticsearch.EsIndex, pattern string) ([]elasticsearch.EsIndex, error) {
 	regexPattern, err := regexp.Compile(pattern)
 	if err != nil {
-		return nil, fmt.Errorf(fmt.Sprintf("There was a problem with the pattern: %s", err))
+		return nil, fmt.Errorf("There was a problem with the pattern: %s", err)
```
frzifus, line 36: Do we use capitalized in the rest of the repo? Otherwise i would recommend to stick to lowercase versions. 

Found this here: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

What do you think?

tests/assert-jobs/index/main.go
```diff
@@ -111,37 +113,45 @@ func initCmd() error {
...
-		fmt.Println(err)
-		os.Exit(1)
+		logrus.Fatalln(err)
 	}
 
-	log = *logger.InitLog(viper.GetBool(flagVerbose))
+	if viper.GetBool(flagVerbose) == true {
+		logrus.SetLevel(logrus.DebugLevel)
+	}
+	logrus.SetOutput(os.Stdout)
```
frzifus, line 122: doesn't logrus use `os.Stdout` by default? And if not, would it make sense to place this call on top of the main function?

frzifus (commented):
looking forward to bulletproof tests :muscle:  only marked a few minor things i noticed.

Dockerfile.asserts
```diff
@@ -18,7 +18,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o ./reporter
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o ./query -a ./tests/assert-jobs/query/main.go
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o ./index -a ./tests/assert-jobs/index/main.go
 
-FROM scratch
+FROM golang:1.16
```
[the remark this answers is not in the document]
frzifus: I would like it to understand why this change was made. But i think its not necessary. up to you :)

tests/assert-jobs/utils/testhttp.go
```diff
@@ -19,35 +20,93 @@ func TestGetHTTP(url string, params *TestParams, testFn func(response *http.Resp
...
+			logrus.Warn("Something failed while executing the test function: ", err.Error())
 			return false, err
 		}
 
-		if len(body) == 0 {
-			return false, fmt.Errorf("empty body")
+		if err == nil {
+			logrus.Warn("The condition of the test function was not accomplished")
+		} else {
+			logrus.Warn("There test function returned an error:", err.Error())
```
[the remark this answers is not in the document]
iblancasa: If you take a look to the previous lines you will see there is a check for the `ok` variable:
```go
		ok, err := testFn(res, body)
		if ok {
			return true, nil
		}
		retries++

		if retries > maxRetries {
			logrus.Warn("Something failed while executing the test function: ", err.Error())
			return false, err
		}

		if err == nil {
			logrus.Warn("The condition of the test function was not accomplished")
		} else {
			logrus.Warn("There test function returned an error:", err.Error())
		}
```

tests/assert-jobs/index/main.go
iblancasa: I didn't know about this. I'll fix it. Thanks!

tests/assert-jobs/index/main.go
iblancasa: That's true! Thanks

tests/assert-jobs/utils/testhttp.go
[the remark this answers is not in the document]
frzifus: oh you are right :+1:

frzifus (approved):
good stuff, up to the others :+1:

iblancasa:
@rubenvp8510 please, take a look to this PR :) Thanks!

frzifus (approved):

pavolloffay (approved):

Issue #1726, opened 2022-01-26
Split the E2E tests into different GitHub Actions workflows


iblancasa:
## Requirement - what kind of business use case are you trying to solve?
Make faster to recover from CI flaky E2E tests.

## Problem - what in Jaeger blocks you from solving the requirement?
Currently, when something fails in the E2E tests, it is needed to rerun the full set of E2E tests. The idea is to split the GitHub workflow in a way it is not needed to rerun tests that are working (or, at least, reduce the number of tests to rerun).

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Split the E2E test into different GitHub Actions workflows.

frzifus:
Issue can be closed?

iblancasa:
Yes

PR #1727, opened 2022-01-26
Split each E2E test suite in a different workflow

iblancasa:
- #1726

- Split the test suites in workflows
- Add a local GitHub action to avoid copying pasting all the steps each time we want to add a new suite

## Notes
* Each test suite will be run in a different workflow. Each workflow will have 4 jobs (1 per Kubernetes release we want to test). If needed, we can split this more to have each `test suite-Kubernetes version` pair in its own workflow (this will require some automation to generate all the workflow files automatically instead of copying pasting everything)
* The GitHub workflows use a local GitHub Action to run the tests. For this purpose, the GitHub Actions reusable workflow feature would be a better fit. The reusable workflow feature didn't provide a way to use a local workflow without a reference (like the branch, the commit...) [until yesterday](https://github.blog/changelog/2022-01-25-github-actions-reusable-workflows-can-be-referenced-locally/) but I saw [some issues to make them work](https://github.community/t/allow-reusable-workflows-to-be-located-at-arbitrary-locations-and-be-local/212745/7). Using a GitHub Action covers what we need.


.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,309 @@
+name: "End-to-end tests"
```
frzifus, line 1: It seems that the file was added again during the merge of the master branch

frzifus (commented):
LGTM, only one thing. Don't we have to delete [.github/workflows/e2e.yaml](https://github.com/iblancasa/jaeger-operator/blob/feature/1726-split-gh-workflows/.github/workflows/e2e.yaml), otherwise the tests will be executed twice?

frzifus (approved):
perfect :)

rubenvp8510 (approved):

PR #1737, opened 2022-02-07
Automatically set num shards and replicas from referenced OCP ES

### LINKED ISSUE #1720: Watch ES CR for changes and corretly set number of shards and replicas
Related to #1707 

The #1707 changed the way how Elasticsearch certificates are provisioned and managed. The PR also made possible to provision ES externally and make Jaeger just connect to it. This however needs further improvements. The jaeger operator should watch referenced ES CR and update shards and replicas accordingly. 

The updated shards should be used in collector, query and ES management scripts.

pavolloffay:
Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

Resolves #1720 

Notable changes:
* defaulting webhook - set ES number of nodes from the referenced ES instance
* validating webhook - validate if references ES instance exists
* ES controller - watch for changes and update ES nodes in Jaeger CR accordingly.

Test data:
```
kubectl apply -f - <<EOF
apiVersion: logging.openshift.io/v1
kind: Elasticsearch
metadata:
  annotations:
    logging.openshift.io/elasticsearch-cert-management: "true"
    logging.openshift.io/elasticsearch-cert.jaeger-shared-es: user.jaeger
  name: shared-es
spec:
  managementState: Managed
  nodeSpec:
    resources:
      limits:
        memory: 1Gi
      requests:
        cpu: 200m
        memory: 1Gi
  nodes:
    - nodeCount: 1
      proxyResources: {}
      resources: {}
      roles:
        - master
        - client
        - data
      storage: {}
EOF

kubectl apply -f - <<EOF
kind: Jaeger
metadata:
  name: tenant1-jaeger
spec:
  strategy: production
  query:
    strategy:
      type: RollingUpdate
  collector:
    strategy:
      type: RollingUpdate
  ingress:
    openshift:
      sar: "" # allow login for all OCP users for testing
  storage:
    type: elasticsearch
    options:
      es:
        index-prefix: tenant1
    elasticsearch:
      name: shared-es
      doNotProvision: true
      nodeCount: 1
      resources:
        requests:
          cpu: 200m
          memory: 1Gi
        limits:
          memory: 1Gi
EOF
```

### DIFF
--- a/apis/v1/jaeger_webhook.go
+++ b/apis/v1/jaeger_webhook.go
@@ -1,8 +1,15 @@
 package v1
 
 import (
+	"context"
+	"fmt"
+
+	esv1 "github.com/openshift/elasticsearch-operator/apis/logging/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
@@ -12,10 +19,14 @@ const (
 )
 
 // log is for logging in this package.
-var jaegerlog = logf.Log.WithName("jaeger-resource")
+var (
+	jaegerlog = logf.Log.WithName("jaeger-resource")
+	cl        client.Client
+)
 
 // SetupWebhookWithManager adds Jaeger webook to the manager.
 func (j *Jaeger) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	cl = mgr.GetClient()
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(j).
 		Complete()
@@ -31,6 +42,19 @@ func (j *Jaeger) Default() {
 
 	if j.Spec.Storage.Elasticsearch.Name == "" {
 		j.Spec.Storage.Elasticsearch.Name = defaultElasticsearchName
+	}
+
+	if ShouldInjectOpenShiftElasticsearchConfiguration(j.Spec.Storage) && j.Spec.Storage.Elasticsearch.DoNotProvision {
+		// check if ES instance exists
+		es := &esv1.Elasticsearch{}
+		err := cl.Get(context.Background(), types.NamespacedName{
+			Namespace: j.Namespace,
+			Name:      j.Spec.Storage.Elasticsearch.Name,
+		}, es)
+		if errors.IsNotFound(err) {
+			return
+		}
+		j.Spec.Storage.Elasticsearch.NodeCount = OpenShiftElasticsearchNodeCount(es.Spec)
 	}
 }
 
@@ -48,6 +72,18 @@ func (j *Jaeger) ValidateCreate() error {
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (j *Jaeger) ValidateUpdate(_ runtime.Object) error {
 	jaegerlog.Info("validate update", "name", j.Name)
+
+	if ShouldInjectOpenShiftElasticsearchConfiguration(j.Spec.Storage) && j.Spec.Storage.Elasticsearch.DoNotProvision {
+		// check if ES instance exists
+		es := &esv1.Elasticsearch{}
+		err := cl.Get(context.Background(), types.NamespacedName{
+			Namespace: j.Namespace,
+			Name:      j.Spec.Storage.Elasticsearch.Name,
+		}, es)
+		if errors.IsNotFound(err) {
+			return fmt.Errorf("elasticsearch instance not found: %v", err)
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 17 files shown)


apis/v1/jaeger_webhook.go
```diff
@@ -12,10 +19,14 @@ const (
 )
 
 // log is for logging in this package.
-var jaegerlog = logf.Log.WithName("jaeger-resource")
+var (
+	jaegerlog = logf.Log.WithName("jaeger-resource")
+	cl        client.Client
```
pavolloffay, line 24: I don't like introducing a global state, but another option would be to implement the webhook "manually", which I like even less for validating and defaulting webhooks.

controllers/elasticsearch/elasticsearch_controller.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/jaegertracing/jaeger-operator/pkg/controller/elasticsearch"
+)
+
+// ElasticsearchReconciler reconciles a Deployment object
+type ElasticsearchReconciler struct {
+	reconcilier *elasticsearch.ReconcileElasticsearch
+}
+
+// NewElasticsearchReconciler creates a new deployment reconciler controller
+func NewElasticsearchReconciler(client client.Client, clientReader client.Reader) *ElasticsearchReconciler {
```
pavolloffay, line 19: To be honest I am not sure whether we should use controller or webook to catch changes on the ES CR. I didn't find any good docs when to use webhook over controller.

apis/v1/jaeger_webhook.go
frzifus: Can we register another object to handle the webhook instead of Jaeger and embed Jaeger?

<details>
  <summary>Just an idea (untested)</summary>
  
```golang
type JaegerWebhook struct {
	Jaeger

	cl client.Client
}

type Jaeger struct {
	...
}
```
</details>

pavolloffay: that unfortunately does not work. We could add the client to the `Jaeger` but that is even worse.

controllers/elasticsearch/elasticsearch_controller.go
rubenvp8510: I think for this case it is OK to use a controller, I haven't found a reference for when to use webhooks or controllers, but I tend to thing that webhooks are for validation or mutating the resource, which is not the case here.  Also this allow us to reenqueue the reconciliation in case of failure.

pavolloffay: I think the controller is fine here, the webhooks should be more used for validation or mutating. 

I had to make one more important change. The controller can be installed only if `Elasticsearch` kind is installed in the cluster (OpenShift Elasticserach operator).

pavolloffay: It seems this is already covered in docs

https://docs.openshift.com/container-platform/4.9/distr_tracing/distr_tracing_install/distr-tracing-installing.html#distr-tracing-jaeger-operator-install_install-distributed-tracing

> If you require persistent storage, you must also install the OpenShift Elasticsearch Operator before installing the Red Hat OpenShift distributed tracing platform Operator.

cc) @JStickler

apis/v1/jaeger_webhook.go
rubenvp8510: I think this is the "less worse" option we have, I don't like it either but I don't see any other possibility so far.

pkg/controller/elasticsearch/elasticsearch_controller.go
```diff
@@ -0,0 +1,103 @@
...
+	// Jaeger instance will be updated in cluster and Jaeger reconciliation will be triggered to
+	// update Jaeger deployments (e.g. --es.num-shards).
+	for i := 0; i < len(jaegers.Items); i++ {
+		jaeger := &jaegers.Items[i]
+		if v1.ShouldInjectOpenShiftElasticsearchConfiguration(jaeger.Spec.Storage) {
+			if jaeger.Spec.Storage.Elasticsearch.Name == es.Name && jaeger.Spec.Storage.Elasticsearch.NodeCount != esNodeCount {
+				logger.WithField("jaeger", jaeger.Name).WithField("old-es-node-count", jaeger.Spec.Storage.Elasticsearch.NodeCount).WithField("new-es-node-count", esNodeCount).Info("Updating Jaeger CR because OpenShift ES number of nodes changed")
+				jaeger.Spec.Storage.Elasticsearch.NodeCount = esNodeCount
+				if err := r.client.Update(ctx, jaeger); err != nil {
+					log.WithError(err).Error("failed to update deployment with sidecar")
```
rubenvp8510, line 95: Is this log error accurate?

apis/v1/jaeger_webhook.go
```diff
@@ -56,3 +92,21 @@ func (j *Jaeger) ValidateDelete() error {
...
+func OpenShiftElasticsearchNodeCount(spec esv1.ElasticsearchSpec) int32 {
+	nodes := int32(0)
+	for i := 0; i < len(spec.Nodes); i++ {
+		nodes += spec.Nodes[i].NodeCount
+	}
+	return nodes
+}
+
+// ShouldInjectOpenShiftElasticsearchConfiguration returns true if OpenShift Elasticsearch is used and its configuration should be used.
+func ShouldInjectOpenShiftElasticsearchConfiguration(s JaegerStorageSpec) bool {
```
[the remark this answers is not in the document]
pavolloffay: The referenced one moved here to the `v1` package.

apis/v1/jaeger_webhook.go
frzifus: While I was working on something else I became aware of the custom validator[0].

What do you think about using it? The method calls differ only slightly. In the `SetupWebhookWithManager` the locale client could be passed to the validator.

```golang
func (r *something) SetupWebhookWithManager(mgr ctrl.Manager) error {
	return ctrl.NewWebhookManagedBy(mgr).
		For(r).
		WithValidator(newSomethingValidator(mgr.GetClient())).
		Complete()
}
```

0: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/validator_custom.go#L30-L35

pavolloffay:
@rubenvp8510 could you please review?

rubenvp8510 (approved):
LGTM

Issue #1649, opened 2021-12-03
Jaeger deployment issues in Kubernetes


venkataphaniv:
Hi,

 I'm working to deploy Jaeger in my test env in AWS EKS. The pods to be monitored are in default namespace.

I've created the Jeager Operator (v1.28) in observability namespace, and the jaeger-operator deployment & pod are getting created successfully.

 When creating the jaeger instance from the operator of type production, the jaeger is created however neither the jaeger pod, nor the ingress are getting created. 

**Version:**
 - OS: Amazon Linux
 - Jaeger version: 1.28
 - Deployment: Kubernetes

## Yaml snippet for tracer deployment with production strategy, storage - cassandra (jtprod.yaml)
--------------------------
```
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: jaegertracer
  namespace:
spec:
  strategy: production
  sampling:
    options:
      default_strategy:
        type: probabilistic
        param: 0.5
  storage:
    type: cassandra
    options:
      cassandra:
        servers: internal-a1-1179306.ap-southeast-1.elb.amazonaws.com
        keyspace: tccc_jaeger
        username: cassandra
        env:
        - name: CASSANDRA_PASSWORD
          valueFrom:
            secretKeyRef:
              key: password
              name: cassandra-secret
        - name: TRACE_TTL
          value: 2629800
        - name: DEPENDENCIES_TTL
          value: 2629800
      cassandraCreateSchema:
        enabled: false
        datacenter: "dev"
        mode: "test"
  collector:
    autoscale: false
    maxReplicas: 1
    resources:
      limits:
        cpu: 30m
        memory: 128Mi
```

## Results:
----------------------

**$ kubectl apply -f  jtprod.yaml**
jaeger.jaegertracing.io/jaegertracer created

**$ kubectl get po --all-namespaaces | grep jaeger**
NAMESPACE   NAME           STATUS   VERSION   STRATEGY     STORAGE     AGE
observability   jaeger-operator-5786ff487b-xbj92          1/1     Running   0          3h32m

**Note:** _jaegertracer pod is not created_

**$ kubectl get jaegers**
NAMESPACE   NAME           STATUS   VERSION   STRATEGY     STORAGE     AGE
default     jaegertracer                      production   cassandra   24s

**$ kubectl get ingress**
No resources found

**$ kubectl -n observability get deployment jaeger-operator -o jsonpath="**
{.spec.template.spec.containers[0].env}"
[...]

PR #1738, opened 2022-02-08
fix panic caused by an invalid type assertion

### LINKED ISSUE #1733: CRD validation does not catch invalid options
**Describe the bug**
Creating a Jaeger instance with Cassandra storage and setting an object as env key causes a panic.

**To Reproduce**
1. Add a Cassandra storage to a spec
2. Use an object as `ENV` key

**Expected behavior**
An error message when creating an instance with invalid parameters.

 - OS: Linux
 - Jaeger version: 1.28 and above
 - Deployment: Kubernetes

**What troubleshooting steps did you try?**
See: https://github.com/jaegertracing/jaeger-operator/issues/1649#issuecomment-1001271023

**Additional context**
See: https://github.com/jaegertracing/jaeger-operator/issues/1649

frzifus:

Closes #1733 

- option fields not of type string, [make the operator panic](https://github.com/jaegertracing/jaeger-operator/issues/1649#issuecomment-1001271023). 

- Check type before making a type association in `(*Options).UnmarshalJSON` method.
- Return error if the elements of the argument list are not of type `string`.
- If `Options` cant be decoded [error 400](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/http_test.go#L78) is returned.

```bash
controller-runtime.webhook.webhooks	received request	{"webhook": "/mutate-jaegertracing-io-v1-jaeger", "UID": "6c77a050-a28d-4172-abc4-56f61670baef", "kind": "jaegertracing.io/v1, Kind=Jaeger", "resource": {"group":"jaegertracing.io","version":"v1","resource":"jaegers"}}
controller-runtime.webhook.webhooks	wrote response	{"webhook": "/mutate-jaegertracing-io-v1-jaeger", "code": 400, "reason": "", "UID": "6c77a050-a28d-4172-abc4-56f61670baef", "allowed": false}
```

### DIFF
--- a/apis/v1/options.go
+++ b/apis/v1/options.go
@@ -63,7 +63,9 @@ func (o *Options) UnmarshalJSON(b []byte) error {
 	if err := d.Decode(&entries); err != nil {
 		return err
 	}
-	o.parse(entries)
+	if err := o.parse(entries); err != nil {
+		return err
+	}
 	o.json = &b
 	return nil
 }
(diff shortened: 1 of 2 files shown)


apis/v1/options.go
```diff
@@ -85,30 +87,43 @@ func (o Options) MarshalJSON() ([]byte, error) {
...
+			entries, err = entry(entries, fmt.Sprintf("%s.%v", key, k), v)
+			if err != nil {
+				return nil, err
+			}
 		}
-	case []interface{}:
-		values := make([]string, 0, len(value.([]interface{})))
-		for _, v := range value.([]interface{}) {
-			values = append(values, v.(string))
+	case []interface{}: // NOTE: explain why we dont look for []string
```
frzifus, line 112: wait i did that :o

rubenvp8510 (approved):

pavolloffay:
Didn't we talk about using validating webhook to catch this?

apis/v1/options.go
```diff
@@ -63,7 +63,9 @@ func (o *Options) UnmarshalJSON(b []byte) error {
 	if err := d.Decode(&entries); err != nil {
 		return err
 	}
-	o.parse(entries)
+	if err := o.parse(entries); err != nil {
+		return err
```
pavolloffay, line 67: is the return error propagated to the user at object creation?

frzifus:
> Didn't we talk about using validating webhook to catch this?

@pavolloffay you are right. In course of doing this, i noticed that the `Jeager` object is already a unmarshalled version of the spec.

apis/v1/options.go
frzifus: Not this particular one. But an error that says your body of your request is invalid.
Its the `"code": 400` in the webhook response.

pavolloffay: If the error is wrapped here would that be propagated to user? 

We should tell the user what was wrong in the CR if that is possible.

frzifus: oh seems i mixed things up. When you create Jeager with invalid `Options` it tells you whats wrong. 

```bash
$ k apply -f invalid.yaml                                                     
Error from server: error when creating "invalid.yaml": admission webhook "mjaeger.kb.io" denied the request: v1.Jaeger.Spec: v1.JaegerSpec.Strategy: Storage: v1.JaegerStorageSpec.Type: Options: unmarshalerDecoder: invalid option type, expect: string, got: map[string]interface {}, error found in #10 byte of ...|":"test"}},"type":"c|..., bigger context ...|"datacenter":"dev","enabled":false,"mode":"test"}},"type":"cassandra"},"strategy":"production"}}|...
```

Looking at the logs, we could explain error `400` a bit better. But placing a description in the  `"reason": ""` field needs to be done in `controller-runtime.webhook.webhooks`.

PR #1748, opened 2022-02-11
fix(config): add missing start

johanneswuerbach:

When using the jaeger operator without metrics / the auth proxy, starting the image fails due to a missing `start` command.

Add `start` similar to https://github.com/jaegertracing/jaeger-operator/blob/main/config/default/manager_auth_proxy_patch.yaml#L23-L26


rubenvp8510:
Nice catch, thanks for your contribution!

rubenvp8510 (approved):

frzifus (approved):

Issue #1700, opened 2022-01-06
Add METRICS_STORAGE_TYPE support


albertteoh:
## Requirement - what kind of business use case are you trying to solve?

For context: https://cloud-native.slack.com/archives/CGG7NFUJ3/p1640003084352300

Add ability for jaeger-query to opt-into querying aggregated span metrics from prometheus.

## Problem - what in Jaeger blocks you from solving the requirement?

This is enabled by setting the `METRICS_STORAGE_TYPE` env var to `prometheus`.

However, it's not possible to define environment variables in jaeger-operator.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Follow a similar pattern to how `SPAN_STORAGE_TYPE` is defined in jaeger-operator via the `spec.storage.type` path, for example, [this is how jaeger-query defines it](https://github.com/jaegertracing/jaeger-operator/blob/91e3b69ee5c8761bbda9d3cf431400a73fc1112a/pkg/deployment/query.go#L138).

From a brief glance, the proposal is to support a `spec.metrics.type` path in both [query](https://github.com/jaegertracing/jaeger-operator/blob/91e3b69ee5c8761bbda9d3cf431400a73fc1112a/pkg/deployment/query.go#L136) and [all-in-one](https://github.com/jaegertracing/jaeger-operator/blob/91e3b69ee5c8761bbda9d3cf431400a73fc1112a/pkg/deployment/all_in_one.go#L146) that will assign this value to the `METRICS_STORAGE_TYPE` env var.

JaredTan95:
and `PROMETHEUS_SERVER_URL ` also need to be added, right?

albertteoh:
> and PROMETHEUS_SERVER_URL also need to be added, right?

_I think_ we won't need to from what I gather in [@jpkrohling's comment](https://cloud-native.slack.com/archives/CGG7NFUJ3/p1640595672034500?thread_ts=1640003084.352300&cid=CGG7NFUJ3):

> if this can't be configured via flags (which the operator does support), the operator has to be changed to support it

because we can configure this via the `--prometheus.server-url` flag.

JaredTan95:
I am interested in this feature

PR #1755, opened 2022-02-17
add METRICS_STORAGE_TYPE for metrics query

JaredTan95:
- Resolves https://github.com/jaegertracing/jaeger-operator/issues/1700
Add ability for jaeger-query to option to querying aggregated span metrics from prometheus.

-


JaredTan95:
@albertteoh would you help me to review this PR?

config/crd/bases/jaegertracing.io_jaegers.yaml
```diff
@@ -6551,6 +6551,11 @@ spec:
                 additionalProperties:
                   type: string
                 type: object
+              metrics:
```
albertteoh, line 6554: This functionality is only implemented in the jaeger-query component, no other component should be aware of querying metrics. 

Given that, do you think it would make more sense to move it under `query`? I think it may have implications for `allInOne` as well. 

I'm pretty new to this repo, so I'm not across what are the best practices for the Jaeger CRD design, just thought to bring it up as a consideration as it's not the same as span storage where the collector also writes to a backing store.

Having said that, I believe @pavolloffay mentioned there's a chance the otel collector's exporters could be brought into Jaeger as part of the v2 effort, meaning the possibility of writing to metrics storage, in which case it's okay as it is here.

apis/v1/jaeger_types.go
```diff
@@ -138,6 +138,9 @@ type JaegerSpec struct {
 	// +optional
 	Storage JaegerStorageSpec `json:"storage,omitempty"`
 
+	// +optional
+	Metrics JaegerMetricsStorageSpec `json:"metrics,omitempty"`
```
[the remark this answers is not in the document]
JaredTan95: make sense.i will update it.

config/crd/bases/jaegertracing.io_jaegers.yaml
JaredTan95: I'm pretty new to this repo, either. This file is generated by `make bundle`, So, I'm not sure this is the right changes.

JaredTan95: Thanks for your review, This file is generated by `make bundle` command, I moved `JaegerMetricsStorageSpec` under `query` and `allInOne` Spec. and I re-run `make bundle`. That's may works right.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Changes by Version
 ==================
 
+* add METRICS_STORAGE_TYPE for metrics query
```
albertteoh, line 4: I don't think you need this, there's a Makefile target to generated the changelog from git commits: https://github.com/jaegertracing/jaeger-operator/blob/main/RELEASE.md#generating-the-changelog

albertteoh (commented):
@JaredTan95 did you get a chance to test a local deployment with these changes?

It also looks like you need to run `make bundle`: https://github.com/jaegertracing/jaeger-operator/runs/5389687115?check_suite_focus=true.

albertteoh:
@JaredTan95  Just curious, how did you test these changes (except for updating unit tests)? Did you try it out on a local k8s cluster?

JaredTan95:
> @JaredTan95 Just curious, how did you test these changes (except for updating unit tests)? Did you try it out on a local k8s cluster?

Hi, @albertteoh 
My test steps are as follows:
1. create a local k8s cluster using `kind create cluster` command with `kind` installed.
2. install `cert-manager` in `kind` cluster.
3. make test operator image using ` make docker` command provided by `jaeger-operator` makefile, and load this image to `kind` by using `kind load docker-image my-jaeger-operator:pr-1755`
4. create k8s namespace 'observability' and deploy my jaeger-operator image by using `IMG=my-jaeger-operator:pr-1755 make deploy`
5. create jaeger `all-in-one` instance with following config yaml:
```yaml
kind: "Jaeger"
metadata:
  name: "my-jaeger"
spec:
  strategy: allInOne
  allInOne:
    image: jaegertracing/all-in-one:1.31
    options:
      log-level: debug
      query:
        base-path: /jaeger
      prometheus:
        server-url: "http://prometheus:9090"
    metricsStorage:
      type: prometheus
  storage:
    options:
      memory:
        max-traces: 100000
``` 
6. last, we can get the jaeger instance and check the system env variables as following commands:
```bash
kubectl get po
kubectl exec -it my-jaeger-76fb5fbd98-7t79f -- /bin/sh
env
```
7. finally, we can see that `METRICS_STORAGE_TYPE=prometheus` has been injected into pod.
image

albertteoh (approved):

PR #1773, opened 2022-02-22
Migrate CronJob from batch/v1beta1 to batch/v1

### LINKED ISSUE #1585: Migrate CronJob from batch/v1beta1 to batch/v1
```
W1018 09:49:01.504879 1398543 warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
W1018 09:49:01.750729 1398543 warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
```

https://kubernetes.io/releases/_print/#release-history

We should clarify since batch/v1 is supported in k8s and OCP.

kevinearls:

- Resolves #1585

- Update Cronjob to use batch/v1 instead of batch/v1beta1

### DIFF
--- a/pkg/controller/jaeger/cronjob.go
+++ b/pkg/controller/jaeger/cronjob.go
@@ -5,7 +5,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"go.opentelemetry.io/otel"
-	batchv1beta1 "k8s.io/api/batch/v1beta1"
+	batchv1 "k8s.io/api/batch/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	v1 "github.com/jaegertracing/jaeger-operator/apis/v1"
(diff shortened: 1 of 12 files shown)


frzifus (commented):
like it, seems to be straightforward. But iam not sure if we need to consider something else to be backward compatible?

up to the others :)

pavolloffay:
@kevinearls could you please dig into the k8s docs and post here since what version `batch/v1` was added and also when `batch/v1beta1` was deprecated (if it is)?

kevinearls:
Yes @pavolloffay I started doing that yesterday but have not been able to find it yet.  Will continue this morning.

kevinearls:
@pavolloffay batch/v1 was not added until 1.21 as show in this changelog: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md  

What should I do here?  Close this PR and wait until 1.21 is the oldest version we support?

rubenvp8510:
> @pavolloffay batch/v1 was not added until 1.21 as show in this changelog: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md
> 
> What should I do here? Close this PR and wait until 1.21 is the oldest version we support?

Is there a way to detect and use the correct version in order to maintain backward compatibility? If not I think we need to wait

pavolloffay:
Correct, it seems v1 was added only on 1.21 and we have to support 1.19 for OCP 4.6.

The betav1 will be removed in k8s 1.25 https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125, so is at that time we will have to still support OCP 4.6 then we will have to dynamically switch to v1.

pavolloffay:
August 9 should be released k8s 1.25 https://kubernetes.io/blog/2021/07/20/new-kubernetes-release-cadence/ and OCP 4.6 (k8s 1.19) support ends October 27. 

This means that there will be period of time that we will have to support batch v1beta1 and v1.

PR #1774, opened 2022-02-22
Reduce the size of the E2E asserts image

iblancasa:
- During #1715, I change the base image used by the E2E asserts container image. This change was done to avoid pulling a new image during the build and being able to use tools like `curl` in the E2E tests. The size of the container increased a lot, doing the `kind load` slower. Also, this size increment would affect if we want to cache the Docker image in the GitHub Actions workflows.


frzifus (approved):

Dockerfile.asserts
```diff
@@ -18,7 +18,9 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o ./reporter
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o ./query -a ./tests/assert-jobs/query/main.go
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o ./index -a ./tests/assert-jobs/index/main.go
 
-FROM golang:1.16
+# Use the curl container image to ensure we have curl installed. Also, it is a
+# minimal container image
+FROM curlimages/curl:7.81.0
```
pavolloffay, line 23: How is this image used? 

Is it published/pushed to an external docker registry? I doubt we do it on the CI. Do you have a link to existing CI run that shows build or pull time of this image? 

I am not against the change, I just want to understand impact of this PR.

iblancasa: It is something we use just for the E2E tests. For the CI, we use `kind` to create a Kubernetes cluster. One of the `kind` commands is `load docker-image`. So, we load it directly into the cluster (without needing any external container registry). If you want to run it on something like OpenShift, you need to push that image to a registry.

rubenvp8510 (approved):
LGTM

PR #1775, opened 2022-02-22
Use only the latest and older supported Kubernetes version for E2E tests

iblancasa:

- The CI is too slow. Use just the latest Kubernetes supported version to test


iblancasa:
@pavolloffay, @kevinearls can you take a look into this PR?

pavolloffay:
@rubenvp8510 could you please review this one?

kevinearls (approved):
LGTM.  At some point soon we should discuss running on other versions

iblancasa:
Thanks for your review @kevinearls :)

rubenvp8510:
This LGTM, but I agree we need to discuss how to test the other versions, specially because we have a compatibility matrix and need to make sure to don't break things on older versions.

rubenvp8510 (dismissed):

pavolloffay:
I am a bit concerned about this. 

We should run against k8s 1.9 as well to make sure the project works well with OCP 4.9. The job can run against all PRs/nightly or perhaps during release (to block it). I would start with all PRs to see if there is improvement and if no then go further.

rubenvp8510:
> I am a bit concerned about this.
> 
> We should run against k8s 1.9 as well to make sure the project works well with OCP 4.9. The job can run against all PRs/nightly or perhaps during release (to block it). I would start with all PRs to see if there is improvement and if no then go further.

I shared this concern, I would like to see something like you mentioned, running full tests on a release or when the PRs got merged to `main`

kevinearls:
Two versions would certainly better.  In the short term running against 4 versions has been causing big backups with PRs.

iblancasa:
This change concerns me too because I think we are gonna miss a lot of corner cases. I talked to @pavolloffay and we think we should keep the newest and the oldest Kubernetes versions in the tests.

Also, @rubenvp8510  since I'm removing some workflows here, I think you need to reset something in the repository. Right? Thanks!

rubenvp8510:
> This change concerns me too because I think we are gonna miss a lot of corner cases. I talked to @pavolloffay and we think we should keep the newest and the oldest Kubernetes versions in the tests.
> 
That is a better approach.

> Also, @rubenvp8510 since I'm removing some workflows here, I think you need to reset something in the repository. Right? Thanks!

Yes!

iblancasa:
Now, I think it should be OK. The non complete jobs are the ones I removed.

pavolloffay (approved):

Issue #1778, opened 2022-02-22
Random kind configuration parsing error in CI


iblancasa:
**Describe the bug**
Sometimes, the CI fails due to the following error:
```sh
>>>> Starting KIND cluster...
/usr/local/bin/kind create cluster --config kind-1.19.yaml 2>&1 | grep -v "already exists" || true
/usr/local/bin/kind: 3: Syntax error: newline unexpected
```

[I created a bug in the `kind` repository](https://github.com/kubernetes-sigs/kind/issues/2642).

iblancasa:
It seems the problem is actually related to an error downloading the binary.

Issue #1750, opened 2022-02-13
add retry mechanism for e2e test workflows


frzifus:
## Requirement - what kind of business use case are you trying to solve?

It seems that the e2e sometimes fail. Instead of having to re-trigger the complete ci logic manually, an automatic retry behavior would be nice.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

So far, I have not dealt with it much. I found [this action](https://github.com/marketplace/actions/retry-step) online. Alternatively, someone [here](https://github.community/t/how-to-retry-a-failed-step-in-github-actions-workflow/125880) describes how to implement a retry logic itself.

## Any open questions to address

What are your thoughts?

---

// cc : @rubenvp8510 @iblancasa

iblancasa:
Thanks for opening this issue! Yes, I agree we are having some problems related to the CI. 🤔  The most of the issues I've seen in the last days are related to downloading resources from the Internet like the Golang dependencies and pulling container images. I'm preparing a PR to speedup the build of the E2E environments and I think most of those problems will disappear. It is based on #1715. But yes, we need something extra. I'll take care of this. 😄 Thanks!

frzifus:
if we can't download dependencies to build / test, maybe it would make sense to put them in a container image?
Our Build image could then call our make rules in the cl.
Just an idea:
```bash
make_ci_test_with_dep:
    docker run --rm -it build_img_with_deps bash -c "make test_with_dep"
---
make ci_test_with_dep 
```

Thanks for taking care :)

iblancasa:
I added some retry mechanisms and I think we can close this issue. What do you think @frzifus?

frzifus:
sure :+1:

PR #1779, opened 2022-02-23
Improve reliability of some scripts to download third party tools

iblancasa:
- #1778
- #1750 

- Move all the `install-<tool>.sh` scripts to the same place
- Update the `Makefile` rules to use those scripts
- Add some common functions to `hack/install/install-utils.sh` to reuse across those scripts
- Avoid downloading those tools if they are not needed


kevinearls (approved):

hack/install/install-utils.sh
```diff
@@ -0,0 +1,93 @@
...
+#
+# Some bash functions to install third party software.
+#
+
+if [[ "$(basename -- "$0")" = "retry.sh" ]]; then
+    echo "Don't run $0, source it" >&2
+    exit 1
+fi
+
+# Retry the given command until 5 times.
```
frzifus, line 11: i like retries

frzifus (approved):

rubenvp8510 (approved):

PR #1810, opened 2022-03-10
Add junit reports

iblancasa:
- Add jUnit generation report to the E2E tests

- Add `undeploy-cert-manager` make rule to undeploy cert manager
- Remove not existing `undeploy-es-operator` make dependency from `clean` rule
- Create `run-e2e-tests.sh` script to handle the logic to run the E2E easier
- Add integration with `junitcli` to fix the jUnit report output provided by KUTTL


iblancasa:
@rubenvp8510 @frzifus please, could you take a look to this PR? Thanks!

hack/run-e2e-tests.sh
```diff
@@ -0,0 +1,69 @@
...
+set +e
+
+$KUTTL test $KUTTL_OPTIONS --report xml
+exit_code=$?
+
+set -e
+
+# The output XML needs some work because it adds "artifacts" as a test case.
+# Also, the suites doesn't have a name so, we need to add one.
+go install github.com/iblancasa/junitcli/cmd/junitcli@latest
```
frzifus, line 61: would it make sense to pin this to a specific version like [v1.0.1](https://github.com/iblancasa/junitcli/releases/tag/v1.0.1)?

frzifus (commented):
looks good, only one thing :)

hack/run-e2e-tests.sh
iblancasa: I though about it but, since what this CLI does is to just format the jUnit XML output, I think it makes sense to maintain with the latest version. Otherwise, if somebody needs to build an old version of the operator or something, can find the output will be different to the expected.

iblancasa: But there is just one accepted output format. What we want is to use the "good" format. If a fix is done in this package, we want it backwards

rubenvp8510: My only concern here is not about the output format, but what happens if the cli flags change or other breaking changes in the cli tool.

iblancasa: Ok, done

frzifus (approved):
Nice :)

rubenvp8510 (approved):

PR #1824, opened 2022-03-23
Make RHOL Elasticsearch cert-management feature optional

pavolloffay:
The reason is that the cert-management is not supported on OCP 4.6.
The feature was added in RHOL 5.2 shipped on OCP 4.7.


rubenvp8510 (approved):
LGTM

PR #1804, opened 2022-03-04
Use mutating webhook to inject sidecar

frzifus:
This PR is based on #1679.

## TODOs

- [ ] fix new e2e tests

## Which problem is this PR solving?

- Pods belonging to a deployment with annotation to inject sidecars are created directly with a sidecar. (prevents pod from being terminated and replaced directly after startup).

## Short description of the changes

- added Annotation support for `sidecar.jaertracing.io/inject=true` on pod level. (NOTE: If this annotation is only specified on pod level, it will not be considered by the reconciler. This means that pods must be restarted manually by the user.)
- added Pod webhook. If a pod has the annotation `sidecar.jaertracing.io/inject=true` a sidecar is added at startup.
- added Deployment webhook. If a deployment has the annotation `sidecar.jaertracing.io/inject=true` it will be added to the pod if the annotation is not explicitly specified there.

## Additional

Note that the webhook thing will work at the end. But this patch is more a hack like turning 3 times to the left to go right. 

- No breaking change
- note... maybe the given commit history is a bit trashy (its a mess of cherry-picks from different test branches..)
- In the follow-up, the behavior of pod annotations must be documented.
- Details of the decision can be found in this[0] document.

<details>
  <summary>Example of valid resources.</summary>
  
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: myapp-pod-anno-standalone
  annotations: # is not considered by reconciler
    "sidecar.jaegertracing.io/inject": "true"
  labels:
    app: myapp-pod-anno-standalone
spec:
  containers:
  - name: myapp
    image: nginx:latest

---

apiVersion: apps/v1
kind: Deployment
metadata:
  name: myapp-deploy-anno
  annotations: # is considered by reconciler after startup
    "sidecar.jaegertracing.io/inject": "true"
spec:
  selector:
    matchLabels:
      app: myapp-deploy-anno
  template:
    metadata:
      labels:
        app: myapp-deploy-anno
    spec:
      containers:
      - name: myapp
        image: nginx:latest

---

apiVersion: apps/v1
kind: Deployment
metadata:
  name: myapp-pod-anno
spec:
  selector:
    matchLabels:
      app: myapp-pod-anno
  template:
    metadata:
      annotations: # is not considered by reconciler
        "sidecar.jaegertracing.io/inject": "true"
      labels:
        app: myapp-pod-anno
    spec:
      containers:
      - name: myapp
        image: nginx:latest
```

</details>

---

[...]

PR #1828, opened 2022-03-25
replace deployment reconciler with webhook

frzifus:
- this pr prevents containers with valid annotation from being created without a sidecar and getting immediately replaced. This was caused by the reconciler determining that a sidecar was needed after the pod was already launched.
- Belongs to https://github.com/jaegertracing/jaeger-operator/pull/1804

- moved deployment reconciler logic into webhook


pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -352,3 +356,75 @@ func (r ReconcileJaeger) getSecretsForNamespace(secrets []corev1.Secret, namespa
 	}
 	return secretsForNamespace
 }
+
+// SyncOnJaegerChanges sync deployments with sidecars when a jaeger CR changes
+func (r *ReconcileJaeger) SyncOnJaegerChanges(object client.Object) []reconcile.Request {
```
rubenvp8510, line 361: As far as I understand this function is watching Jaeger CRs for changes, and triggers deployment/namespace reconciliations using a modified annotation.
Why do we need this,  Could be a part of the Jaeger CRs  reconciliation? what is the difference?

pkg/controller/jaeger/jaeger_controller.go
```diff
@@ -352,3 +356,75 @@ func (r ReconcileJaeger) getSecretsForNamespace(secrets []corev1.Secret, namespa
...
+		}
+
+		// if the namespace has the sidecar annotation, trigger a reconciliation
+		if ok := increaseRevision(ns.Annotations); ok {
+			nssupdate = append(nssupdate, ns)
+			continue
+		}
+	}
+	for _, dep := range deps {
+		if err := r.client.Update(context.Background(), &dep); err != nil {
```
rubenvp8510, line 407: I only have one concern with this approach, it could be the case that the deployment change its state between the time you get it and the time you updated it (for instance from the initial state to RUNNING) . In that case you will get an error like this:

`ERRO[1123] failed to update deployment with sidecar      error="Operation cannot be fulfilled on deployments.apps \"simple-prod-query\": the object has been modified; please apply your changes to the latest version and try again" jaeger=simple-prod jaeger-namespace=monitoring name=simple-prod-query namespace=monitoring`

Because the version is not the same, I think in a normal reconciliation process you can handle this by returning an error , which eventually will trigger a new reconciliation loop. but I'm not sure how are we handling this case with this approach.

pkg/controller/jaeger/jaeger_controller.go
frzifus: Sure we can move it up. the function exists only because I have tried to change as little as possible in the existing behavior. But it makes sense. especially considering the [potential error](https://github.com/jaegertracing/jaeger-operator/pull/1828#discussion_r837041904) when the deployment state has changed.

rubenvp8510 (approved):
LGTM

frzifus:
improved flaky unittest https://github.com/jaegertracing/jaeger-operator/pull/1828/commits/32eeb2afe9fe6af827c96f8f6ed8e4f337fbd0e6 and rebased branch onto main.

cc @rubenvp8510

rubenvp8510 (approved):

PR #1830, opened 2022-03-28
Schedule dependabot jobs to run after 3:00 AM CET

kevinearls:

- Currently dependabot jobs can run at any time of day, and as a result sometimes clog up the build queue.  This change should restrict them to running after 3:00AM CET, or after 9:00PM eastern time in the US.

- Add a time and timezone to the dependabot config file

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,12 +4,12 @@ updates:
     directory: "/"
     schedule:
       interval: daily
-      time: "01:00"
+      time: "03:00"
       timezone: "Europe/Berlin"
 
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: "daily"
-      time: "01:00"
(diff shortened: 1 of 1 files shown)


frzifus (dismissed):
Like the idea! What I was just thinking though, would it maybe make sense to run/allow dependabot only on weekends? What you think?

kevinearls:
@rubenvp8510 Please review.

@frzifus I suppose we could do that.  I just assumed after 1AM CET would normally be quiet enough.  Maybe we should discuss this at the next meeting?

frzifus:
Oh its also fine this way:+1:. But sure, we can probably ask the others for their opinion.

iblancasa (approved):

frzifus (approved):

frzifus:
oh, @kevinearls your commit is not singed off

PR #1839, opened 2022-03-31
fix: take namespace modifications into account

### LINKED ISSUE #1838: operator doesnt observe ns changes
**Describe the bug**

Annotation change does not trigger a namespace reconcile.

**To Reproduce**
1. Observe logs.
2. `k annotate namespaces default --overwrite sidecar.jaegertracing.io/inject=true`
3. `k annotate namespaces default --overwrite sidecar.jaegertracing.io/inject-`

**Expected behavior**
If the annotation of a namespace changes, the corresponding reconcile call is executed.
This checks whether deployments in the namespace need sidecars or need to be removed.

 - Jaeger version: >1.32

**Additional context**
I assume its caused by https://github.com/jaegertracing/jaeger-operator/pull/1828.

Previously the namespace reconciler was called by reconcile requests generated by the [JaegerOnSync](https://github.com/jaegertracing/jaeger-operator/blob/2167682f4c49d2668f68b7b57177f9f9c40923d9/pkg/controller/deployment/deployment_controller.go#L218) method. In https://github.com/jaegertracing/jaeger-operator/pull/1828 the JaegerOnSync method changes an annotation of the namespace to call the reconcile loop. But the reconciler [does not watch for namespaces](https://github.com/jaegertracing/jaeger-operator/blob/7f8c23f246d44822d90bb3371cd8872a347b93b8/controllers/appsv1/namespace_controller.go#L40).

frzifus:
- this pr removes the sidecar logic from the namespace reconciler. If an ns is changed, the reconciler now increments the revision number of the affected deployments. In the deployment webhook it is then evaluated if the PodSpec has to be modified or not.
- Belongs to https://github.com/jaegertracing/jaeger-operator/pull/1804
- Follow-up of https://github.com/jaegertracing/jaeger-operator/pull/1828
- Fixes #1838

- namespace reconciler is no longer adding or removing sidecars to deployments.
- namespace reconciler increases revision annotations of affected deployment by namespace change.


rubenvp8510 (approved):
LGTM Good Job!

PR #1841, opened 2022-04-04
Use go install instead of go-get-tool to install controller-gen and s…

kevinearls:
…etup-envtest

- make controller-gen and setup-envtest were not working properly on my laptop.  Instead of installing into the local ./bin director they were installing into ~/go/bin, even though I did not have GOBIN set

- Use go install to install these tools


frzifus (approved):

kevinearls:
@rubenvp8510 Please review

rubenvp8510 (approved):

PR #1850, opened 2022-04-08
Fix: webhook deadlock 

### LINKED ISSUE #1844: Deployment of jaeger-operator to minikube running k8s 1.19 is blocked by webhook
**Describe the bug**
I tried to deploy the jaeger-operator to a minikube instance running kubernetes version 1.19 and it failed

**To Reproduce**
1. Start minikube with **minikube start --memory 12288 --vm-driver hyperkit --addons "ingress,metrics-server" --kubernetes-version='1.19.0'**
2.make docker push deploy

**Expected behavior**
The operator should deploy

**Screenshots**
```
certificate.cert-manager.io/jaeger-operator-serving-cert created
issuer.cert-manager.io/jaeger-operator-selfsigned-issuer created
mutatingwebhookconfiguration.admissionregistration.k8s.io/jaeger-operator-mutating-webhook-configuration configured
validatingwebhookconfiguration.admissionregistration.k8s.io/jaeger-operator-validating-webhook-configuration configured
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "deployment.sidecar-injector.jaegertracing.io": Post "https://jaeger-operator-webhook-service.observability.svc:443/mutate-v1-deployment?timeout=10s": dial tcp 10.101.31.210:443: connect: connection refused
make: *** [deploy] Error 1
```

frzifus:
- fixes #1844

- change deployment webhook [failurePolicy](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy) to ignore.
- add common label [name](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels)
- patch webhook to ignore deployments with label `app.kubernetes.io/name: jaeger-operator`

## TODO
- add e2e test to verify the patch works as expected.

---
cc @esnible

### DIFF
--- a/config/webhook/deployment_inject_patch.yaml
+++ b/config/webhook/deployment_inject_patch.yaml
@@ -1,0 +1,12 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: mutating-webhook-configuration
+webhooks:
+  - name: deployment.sidecar-injector.jaegertracing.io
+    objectSelector: # Skip resources with the name jaeger-operator
+      matchExpressions:
+        - key: app.kubernetes.io/name
+          operator: NotIn
+          values:
+            - "jaeger-operator"
--- a/controllers/appsv1/deployment_webhook.go
+++ b/controllers/appsv1/deployment_webhook.go
@@ -37,7 +37,7 @@ func NewDeploymentInterceptorWebhook(c client.Client) webhook.AdmissionHandler {
 }
 
 // You need to ensure the path here match the path in the marker.
-// +kubebuilder:webhook:path=/mutate-v1-deployment,mutating=true,failurePolicy=fail,groups="apps",resources=deployments,sideEffects=None,verbs=create;update,versions=v1,name=deployment.sidecar-injector.jaegertracing.io,admissionReviewVersions=v1
+// +kubebuilder:webhook:path=/mutate-v1-deployment,mutating=true,failurePolicy=ignore,groups="apps",resources=deployments,sideEffects=None,verbs=create;update,versions=v1,name=deployment.sidecar-injector.jaegertracing.io,admissionReviewVersions=v1
 
 // +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=namespaces/status,verbs=get;update;patch
--- a/bundle/manifests/jaeger-operator-metrics-reader_rbac.authorization.k8s.io_v1_clusterrole.yaml
+++ b/bundle/manifests/jaeger-operator-metrics-reader_rbac.authorization.k8s.io_v1_clusterrole.yaml
@@ -3,6 +3,7 @@ kind: ClusterRole
 metadata:
   creationTimestamp: null
   labels:
+    app.kubernetes.io/name: jaeger-operator
     name: jaeger-operator
   name: jaeger-operator-metrics-reader
 rules:
--- a/bundle/manifests/jaeger-operator-metrics_v1_service.yaml
+++ b/bundle/manifests/jaeger-operator-metrics_v1_service.yaml
@@ -4,6 +4,7 @@ metadata:
   creationTimestamp: null
   labels:
     app.kubernetes.io/component: metrics
+    app.kubernetes.io/name: jaeger-operator
     name: jaeger-operator
   name: jaeger-operator-metrics
 spec:
@@ -13,6 +14,7 @@ spec:
     protocol: TCP
     targetPort: https
   selector:
+    app.kubernetes.io/name: jaeger-operator
     name: jaeger-operator
 status:
   loadBalancer: {}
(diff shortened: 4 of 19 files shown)


config/webhook/deployment_inject_patch.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: mutating-webhook-configuration
+webhooks:
+  - name: deployment.sidecar-injector.jaegertracing.io
+    objectSelector: # Skip resources with the name jaeger-operator
+      matchExpressions:
+        - key: app.kubernetes.io/name
+          operator: In
```
esnible, line 10: Wouldn't you want `NotIn` here?

esnible: For Deployments we offer a way to control auto-injection:

```yaml
  annotations:
    "sidecar.jaegertracing.io/inject": "true"
```

(See https://www.jaegertracing.io/docs/1.32/operator/#auto-injecting-jaeger-agent-sidecars )

Do we have a similar annotation on the Pod level?  I'd like to be able to attach

```yaml
  annotations:
    "sidecar.jaegertracing.io/inject": "false"
```

to pods (and pod-templates) that must not be attached to the sidecar.  Then I would like this webhook to check for that label, rather than only skipping mutation for the operator itself.

esnible: https://github.com/jaegertracing/jaeger-operator/pull/1804 is closed.

The `failurePolicy: Ignore` is great but it should be possible for someone who doesn't want the sidecar to annotate their pods so that the webhook doesn't fire-then-fail.

For example, Istio's `MutatingWebhookConfiguration` uses

```YAML
  namespaceSelector:
    matchExpressions:
    - key: istio-injection
      operator: DoesNotExist
    - key: istio.io/rev
      operator: In
      values:
      - {{istio-instance}}
  objectSelector:
    matchExpressions:
    - key: "sidecar.istio.io/inject"
      operator: NotIn
      values:
      - "false"
```

frzifus: > https://github.com/jaegertracing/jaeger-operator/pull/1804 is closed.

Y, its not supported. We agreed in not extending the Jaeger-Operator functionality. Instead we focus on the [OTEL](https://github.com/open-telemetry/opentelemetry-operator).

> In v1.15+, webhooks may optionally limit which requests are intercepted based on the **labels** of the objects they would be sent, by specifying an objectSelector. 

[Istio](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy) does this on a label basis using the [objectSelector](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-objectselector). Jaeger-Operator uses annotations instead. Unfortunately, these cannot be pre-filtered. That means all new deployments pass the webhook and [there is checked](https://github.com/jaegertracing/jaeger-operator/blob/main/controllers/appsv1/deployment_webhook.go#L101) for the corresponding annotation and value.

controllers/appsv1/deployment_webhook.go
```diff
@@ -37,7 +37,7 @@ func NewDeploymentInterceptorWebhook(c client.Client) webhook.AdmissionHandler {
 }
 
 // You need to ensure the path here match the path in the marker.
-// +kubebuilder:webhook:path=/mutate-v1-deployment,mutating=true,failurePolicy=fail,groups="apps",resources=deployments,sideEffects=None,verbs=create;update,versions=v1,name=deployment.sidecar-injector.jaegertracing.io,admissionReviewVersions=v1
+// +kubebuilder:webhook:path=/mutate-v1-deployment,mutating=true,failurePolicy=ignore,groups="apps",resources=deployments,sideEffects=None,verbs=create;update,versions=v1,name=deployment.sidecar-injector.jaegertracing.io,admissionReviewVersions=v1
```
pavolloffay, line 40: +1.

Do we still need the label exclusion conf in the webhook if we will be using `ignore` `failurePolicy`? 

```yaml
webhooks:
  - name: deployment.sidecar-injector.jaegertracing.io
    objectSelector: # Skip resources with the name jaeger-operator
      matchExpressions:
        - key: app.kubernetes.io/name
          operator: NotIn
          values:
            - "jaeger-operator"
```

frzifus: really need probably not. But doesn't it make sense to at least exclude the operator? [Isito excludes](https://github.com/istio/istio/blob/098cc47dfcbdca202208bd241d9dee35cd230a4d/manifests/charts/istio-control/istio-discovery/templates/mutatingwebhook.yaml) even more to avoid to many (in this case pods) pass their webhooks.

And should we then keep `app.kubernetes.io/name`?

What do you think?

pavolloffay (approved):

controllers/appsv1/deployment_webhook.go
pavolloffay: No objection, perhaps then we can remove `app.kubernetes.io/name` and use something more generic. Is istio using a generic label/annotation to bypass the injection?

frzifus: Everything without a label to inject or a label injected doesnt pass the webhook.

rubenvp8510 (approved):
This LGTM, if no objections I'll merge this.

Issue #1855, opened 2022-04-13
Makefiles needs a target for installing junitcli


kevinearls:
**Describe the bug**
make test uses the junitcli.  The Makefile should have a target to install this.

**To Reproduce**
1. make test

kevinearls:
I'm getting the following errors, even after installing junitcli and creating a reports directory

```
# Show a final report
junitcli --report reports
FATA[0000] no error suites found! No report printed
make: *** [run-e2e-tests] Error 1
```

The make test target should do whatever is required to run and report on tests

frzifus:
Its moved into the [hack folder](https://github.com/jaegertracing/jaeger-operator/blob/main/hack/run-e2e-tests.sh#L61) since https://github.com/jaegertracing/jaeger-operator/pull/1810.
try: `make tools`

iblancasa:
The problem is not produced in the `test` make target. It is in the `run-e2e-tests` make target. `junit` is installed as part of running `hack/run-e2e-tests.sh` (which is run for each test suite). It seems, for whatever reason, `Make` is not able to access that program from the path. I just created a PR fixing this issue.

rubenvp8510:
Is this already solved by https://github.com/jaegertracing/jaeger-operator/pull/1861 ? Can we close this issue?

iblancasa:
I think so @rubenvp8510

PR #1861, opened 2022-04-18
Fix call to junitcli

iblancasa:

- #1855

-  Use the proper path to use `junitcli`

### DIFF
--- a/tests/e2e/Makefile
+++ b/tests/e2e/Makefile
@@ -79,7 +79,7 @@ run-e2e-tests: list-test-suites
 		$(MAKE) run-e2e-tests-$$suite -i; \
 	done
 	# Show a final report
-	junitcli --report reports
+	$(shell go env GOPATH)/bin/junitcli --report reports
 	cd reports && ! grep -nr failed
 
 .PHONY: render-e2e-tests


tests/e2e/Makefile
```diff
@@ -79,7 +79,7 @@ run-e2e-tests: list-test-suites
 		$(MAKE) run-e2e-tests-$$suite -i; \
 	done
 	# Show a final report
-	junitcli --report reports
+	$(shell go env GOPATH)/bin/junitcli --report reports
```
frzifus, line 82: isn it `$(shell pwd)/bin`?

iblancasa: Since it is installed using `go get..` it is installed under `GOPATH`

frzifus: Would it make sense to install it into ${PWD}/bin ?

frzifus (approved):

tests/e2e/Makefile
iblancasa: If it is possible to install it using `go get...`, we are using that method instead of installing it manually.

frzifus (approved):

rubenvp8510 (approved):

Issue #1840, opened 2022-04-01
sidecar-namespace e2e test doesnt work as expected


frzifus:
**Describe the bug**
It seems that the namespace annotation has no effect in the end-to-end test. It is noticed by the fact that https://github.com/jaegertracing/jaeger-operator/pull/1828 passed the test successfully and https://github.com/jaegertracing/jaeger-operator/issues/1838 occurred.

**To Reproduce**
1. Disable ns reconciler
2. Run `KUTTL_OPTIONS="--test sidecar-namespace"  make run-e2e-tests-sidecar`

**Expected behavior**
Until https://github.com/jaegertracing/jaeger-operator/pull/1839 gets merged, the test should fail.

**Version (please complete the following information):**
 - Jaeger-operator version: >1.32

**Additional context**
The problem seems to be that in the [vertx template](https://github.com/jaegertracing/jaeger-operator/blob/e484d76c7040962d222ac5b36c520422c8fb2faa/tests/templates/vertex-install.yaml.template#L6) the deployment is already annotated with `true`. While fixing this, we should consider adding an extra step, setting the namespace annotation to `false` and verify that the sidecar gets removed.

---
cc @iblancasa @rubenvp8510

PR #1865, opened 2022-04-20
Multiple fixes to sidecar E2E tests

iblancasa:

* Fix sidecar-namespace to ensure the autoinjection is done when annotating the namespace #1840 
* Add missing README.md in sidecar-skip-webhook (missing in #1839)
* Fix sidecar-skip-webhook to not overwrite the logic when rendering (added in #1839)
* Check the sidecar is created when the label is removed and the autoinjection is created in sidecar-skip-webhook
* Add a check for the sidecar-deployment E2E test to ensure the sidecar is removed when the annotation is removed
* Add a check for the sidecar-namespace E2E test to ensure the sidecar is removed when the annotation is removed
* Join the sidecar-deployment and sidecar-agent tests 

- Remove the autoinjection from the vertx deployment template
- Remove the `annotate.sh` file to use a `kubectl` command instead
- Add missing readme for sidecar-skip-webhook
- Add logic to ensure the reconcilation works properly


tests/templates/vertex-install.yaml.template
```diff
@@ -2,8 +2,6 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: vertx-create-span-sidecar
-  annotations:
-    "sidecar.jaegertracing.io/inject": "true"
```
frzifus, line 6: :+1:

frzifus (approved):
Thanks @iblancasa, great job. Not sure if i overlooked something. So better @rubenvp8510 has a look again.

rubenvp8510 (approved):
LGTM Good Work!

PR #1893, opened 2022-05-19
Fix run-e2e-tests make target

iblancasa:
- Fix the `run-e2e-tests` make target. It was not reporting properly the exit code
- Fix some issues with the `miscellaneous/cassandra-spark` E2E test
- Skip some test not supported in OpenShift when running in OpenShift


frzifus (approved):

iblancasa:
@rubenvp8510, could you take a look to this PR? Thanks!

Makefile
```diff
@@ -60,7 +60,7 @@ LD_FLAGS ?= "-X $(VERSION_PKG).version=$(VERSION) -X $(VERSION_PKG).buildDate=$(
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.22
 # Options for kuttl testing
-KUBE_VERSION ?= 1.20
+export KUBE_VERSION ?= 1.20
```
rubenvp8510, line 63: why do we need this export?

iblancasa: Because doing this, we can use `KUBE_VERSION` as environment variable of the scripts called by `make`

Issue #1894, opened 2022-05-19
Additional info in documentation


tide34:

tide34:
I think it's impossible to start jaeger-operator in not cluster-wide mode when follow steps in documention.

tide34:
@yurishkuro hello. Will anyone help solve it?

frzifus:
@rubenvp8510 Do you have experience with it?

tide34:
no one help

rubenvp8510:
Let me try follow your steps, and will back to you with more information, we might need to update the documentation.

rubenvp8510:
Could you attach your manifest yaml file?

rubenvp8510:
Ok I tried to install it on restricted mode (only watching observability namespace) and it works for me.

A couple of points you can check. See if the Roles and RoleBindings are created in the respective namespace. I was able to reproduce the issue with the following command:

`kubectl apply -f manifest.yaml`

But if I do this:
`kubectl apply -f manifest.yaml -n observability`

The error of the namespace goes away, and the operator start successfully.

tide34:
Steps (use minikube):  
1) install cert-manager v1.6.3 (follow [this](https://cert-manager.io/v1.6-docs/installation/kubectl/#steps) guide)  
2) download [jaeger-operator](https://github.com/jaegertracing/jaeger-operator/releases/download/v1.34.0/jaeger-operator.yaml) manifest (from [documentation](https://www.jaegertracing.io/docs/1.34/operator/#installing-the-operator-on-kubernetes))  
3) as documentation told if i want to install jaeger-operator in non cluster wide mode (watch jaeger CR in specific namespaces ONLY, in my case for example namespace app-profile) i need to change the ClusterRole and ClusterBindingRole of the operator manifest to Role and RoleBinding, also set the WATCH_NAMESPACE environmebt variable on the jaeger operator Deployment. Have done it. Got next [manifest](https://github.com/tide34/jaeger-operator/blob/main/jaeger-operator.yaml) 
4) create [this](https://github.com/tide34/jaeger-operator/blob/main/additional.yaml) manifest in namespace app-profile (documentation [link](https://www.jaegertracing.io/docs/1.34/operator/#install-modes))  
  
---  
in log still see line:  
`time="2022-05-29T07:17:46Z" level=info msg="Consider running the operator in a cluster-wide scope for extra features"`  
Pod of jaeger-operator restarts

tide34:
[...]

PR #1914, opened 2022-05-29
Verify namespace permissions before adding ns controller

rubenvp8510:

- Fixes https://github.com/jaegertracing/jaeger-operator/issues/1894

-  Before adding namespace controller we need to check the permissions, if not namespace controller will be added and prevent the operator to start in non cluster-wide mode.

### DIFF
--- a/pkg/cmd/start/bootstrap.go
+++ b/pkg/cmd/start/bootstrap.go
@@ -356,9 +356,13 @@ func setupControllers(ctx context.Context, mgr manager.Manager) {
 		os.Exit(1)
 	}
 
-	if err := appsv1controllers.NewNamespaceReconciler(client, clientReader, schema).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "Namespace")
-		os.Exit(1)
+	if viper.GetBool(v1.ConfigEnableNamespaceController) {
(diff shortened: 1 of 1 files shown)


rubenvp8510:
@frzifus Could you review?

Thanks

frzifus (approved):

pkg/cmd/start/bootstrap.go
```diff
@@ -356,9 +356,13 @@ func setupControllers(ctx context.Context, mgr manager.Manager) {
...
-	if err := appsv1controllers.NewNamespaceReconciler(client, clientReader, schema).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "Namespace")
-		os.Exit(1)
+	if viper.GetBool(v1.ConfigEnableNamespaceController) {
+		if err := appsv1controllers.NewNamespaceReconciler(client, clientReader, schema).SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "Namespace")
+			os.Exit(1)
+		}
+	} else {
+		log.Trace("skipping reconciliation for namespaces, do not have permissions to list and watch namespaces")
```
frzifus, line 365: Should we print a warning instead of using trace level? what do you think?

rubenvp8510: This log level was taken from the original code, but yeah I think a warning level is better.

Issue #1896, opened 2022-05-24
es-index-cleaner does not pass TLS flags


loelu:
**Describe the bug**
Since v1.34.0 Jaeger the `tls.enabled=true` flag is mandatory if other `tls.*` flags are used: https://github.com/jaegertracing/jaeger/pull/3030
But the Operator does not seem to pass the `tls.enabled` flag from the es config to the es-index-cleaner CronJob

**To Reproduce**
1. Create Jaeger instance with a secured connection to Elasticsearch
```yaml
...
    storage:
      elasticsearch:
        name: elasticsearch
      esIndexCleaner:
        enabled: true
        numberOfDays: 10
        resources: {}
        schedule: 55 22 * * *
      options:
        es:
          num-replicas: 0
          num-shards: 5
          server-urls: https://jaeger-es-http:9200
          tls:
            ca: /es/certificates/tls.crt
            enabled: true
...
```
2. Wait for execution of es-index-cleaner Job and observe the following error:
```
> kubectl logs jaeger-instance1-es-index-cleaner-1653346500-9zsp2 -n instance1
Error: es.tls.* options cannot be used when es.tls.enabled is false
Usage:
  jaeger-es-index-cleaner NUM_OF_DAYS http://HOSTNAME:PORT [flags]

Flags:
      --archive                       Whether to remove archive indices. It works only for rollover
      --es.password string            The password required by storage
      --es.tls.ca string              Path to a TLS CA (Certification Authority) file used to verify the remote server(s) (by default will use the system truststore)
      --es.tls.cert string            Path to a TLS Certificate file, used to identify this process to the remote server(s)
      --es.tls.enabled                Enable TLS when talking to the remote server(s)
      --es.tls.key string             Path to a TLS Private Key file, used to identify this process to the remote server(s)
      --es.tls.server-name string     Override the TLS server name we expect in the certificate of the remote server(s)
      --es.tls.skip-host-verify       (insecure) Skip server's certificate chain and host name verification
      --es.username string            The username required by storage
  -h, --help                          help for jaeger-es-index-cleaner
      --index-date-separator string   Index date separator (default "-")
      --index-prefix string           Index prefix
      --rollover                      Whether to remove indices created by rollover
      --timeout int                   Number of seconds to wait for master node response (default 120)

2022/05/23 22:55:22 es.tls.* options cannot be used when es.tls.enabled is false

```
**Expected behavior**
[...]

PR #1921, opened 2022-05-30
fix: skip dependencies on openshift platform

frzifus:

- Prevent write to dependency store on openshift platform. (currently this does not work because permissions are limited)

- When using openshift, the dependency index rollover is disabled. 

---

cc @rubenvp8510

### DIFF
--- a/pkg/cronjob/es_rollover.go
+++ b/pkg/cronjob/es_rollover.go
@@ -213,6 +213,13 @@ func EsScriptEnvVars(opts v1.Options) []corev1.EnvVar {
 			envs = append(envs, corev1.EnvVar{Name: x.envVar, Value: val})
 		}
 	}
+
+	if val, ok := options["skip-dependencies"]; ok {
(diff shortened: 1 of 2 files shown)


pkg/cronjob/es_rollover.go
```diff
@@ -213,6 +213,13 @@ func EsScriptEnvVars(opts v1.Options) []corev1.EnvVar {
 			envs = append(envs, corev1.EnvVar{Name: x.envVar, Value: val})
 		}
 	}
+
+	if val, ok := options["skip-dependencies"]; ok {
```
rubenvp8510, line 217: Why do we need to add the environment var if the option is specified?

frzifus: For cronjobs, only selected parameters are passed as environment variables. Here we pass user settings if they exist. If no user settings are present and the operator runs on openshift, we set `skip-dependencies=true`.

Details: https://github.com/jaegertracing/jaeger-operator/issues/1896#issuecomment-1135894828

rubenvp8510 (approved):

PR #1930, opened 2022-06-02
Call cert-manager only when it is needed

iblancasa:

- Call the `cert-manager` `make` rule only when it is needed.


frzifus (approved):
:+1:

rubenvp8510 (approved):

PR #1931, opened 2022-06-02
Fix namespace deletion for es-multiinstance E2E test

iblancasa:

- The test was failing because the secondary namespace could not be removed


frzifus (approved):

tests/e2e/elasticsearch/es-multiinstance/05-delete.yaml
```diff
@@ -2,8 +2,5 @@
 # the test
 apiVersion: kuttl.dev/v1beta1
 kind: TestStep
-delete:
-- apiVersion: v1
```
rubenvp8510, line 6: Why this doesn't work?

iblancasa: Because KUTTL ony allows removing stuff from the same namespace

rubenvp8510 (approved):
LGTM

PR #1948, opened 2022-06-13
Expose OTLP collector and allInOne ports

rubenvp8510:

-  Jaeger 1.35 now add support for receiving OTLP on the collector, this PR adds support for the operator, it exposes the necessary ports so pods can send OTLP data directly to the collector.

@pavolloffay @frzifus  Should we make this optional with the posibility of enable this using a CR field? WDYT?


pavolloffay:
>  Should we make this optional with the posibility of enable this using a CR field? WDYT?

Jaeger collector by default does not enable Zipkin or OTLP. However, the operator enables Zipkin by default. We could take the  same approach for OTLP - enable it by default and disable if flag `--collector.otlp.enabled=false` is present. 

We should be careful not to collide with OTELcol objects for the OTLP service.

rubenvp8510:
> @rubenvp8510 could you provide an example CR for this feature?

The same, this is not modifying the CRD, and you don't need to add anything to your CR to enable it.

rubenvp8510:
@pavolloffay @frzifus please review.

rubenvp8510:
btw this cannot work with sidecar, the test I did report it directly to the collector. no agent

frzifus (commented):
> btw this cannot work with sidecar, the test I did report it directly to the collector. no agent

Should be fine, right? it only moved into the collector. https://github.com/jaegertracing/jaeger/pull/3710/files#diff-d30ac53f2fdb1f4d0ad0d571bdb4a3a4ebc9c353f682d909e219fba6714c8d1e

rubenvp8510:
> agent

Yep

frzifus (approved):
LGTM

pkg/cmd/start/main.go
```diff
@@ -34,6 +34,7 @@ func AddFlags(cmd *cobra.Command) {
 	cmd.Flags().String("es-provision", "auto", "Whether to auto-provision an Elasticsearch cluster for suitable Jaeger instances. Possible values: 'yes', 'no', 'auto'. When set to 'auto' and the API name 'logging.openshift.io' is available, auto-provisioning is enabled.")
 	cmd.Flags().String("kafka-provision", "auto", "Whether to auto-provision a Kafka cluster for suitable Jaeger instances. Possible values: 'yes', 'no', 'auto'. When set to 'auto' and the API name 'kafka.strimzi.io' is available, auto-provisioning is enabled.")
 	cmd.Flags().Bool("kafka-provisioning-minimal", false, "(unsupported) Whether to provision Kafka clusters with minimal requirements, suitable for demos and tests.")
+	cmd.Flags().Bool("collector-otlp-enabled", true, "Expose otlp ports on the collector.")
```
pavolloffay, line 37: I would say the flag is not needed, and rather the operator should check if the OTLP collector flag is defined in the CR options and disable it if the value is false. 

We also don't expose the Zipkin enable flag on the operator.

Issue #1952, opened 2022-06-17
oauth-proxy cannot be started after setting sar options


frzifus:
**Describe the bug**
The Jaeger Operator is not working on OpenShift when setting some options in order to allow accessing the Jaeger Query REST API.

**To Reproduce**
1. Create a new namespace: kubectl create namespace testing
2. Deploy this Jaeger CRD in the created namespace:
```yaml
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: simple-prod
spec:
  strategy: production
  storage:
    type: elasticsearch
    elasticsearch:
      nodeCount: 1
      resources:
        requests:
          cpu: 100m
          memory: 1Gi
        limits:
          memory: 1Gi
  query:
    options:
      query.bearer-token-propagation: true
  ingress:
    openshift:
      sar: '{"namespace": "testing", "resource": "pods", "verb": "get"}'
      delegateUrls: '{"/":{"namespace": "testing", "resource": "pods", "verb": "get"}}'
    options:
      pass-access-token: true
      pass-user-bearer-token: true
      scope: "user:info user:check-access"
      pass-basic-auth: false
```

**Version (please complete the following information):**
 - OS: Linux
 - Jaeger version: 1.34
 - Deployment: OpenShift 4.10

**What troubleshooting steps did you try?**

As a result, the Jaeger Query pod status is CrashLoopBackOff.

```bash
2022/06/08 21:16:36 provider.go:106: Defaulting client-id to system:serviceaccount:kuttl-test-expert-bulldog:simple-prod-ui-proxy
2022/06/08 21:16:36 provider.go:111: Defaulting client-secret to service account token /var/run/secrets/kubernetes.io/serviceaccount/token
2022/06/08 21:16:36 provider.go:290: Delegation of authentication and authorization to OpenShift is enabled for bearer tokens and client certificates.
2022/06/08 21:16:36 main.go:138: Invalid configuration:
  unable to load OpenShift configuration: unable to retrieve authentication information for tokens: the server could not find the requested resource
  ```

**Additional context**

Events:
```bash
  Normal   Pulled          43s                kubelet            Successfully pulled image "openshift/oauth-proxy:latest" in 361.354385ms
  Warning  Unhealthy       41s (x3 over 43s)  kubelet            Readiness probe failed: HTTP probe failed with statuscode: 503
  Warning  BackOff         37s (x3 over 42s)  kubelet            Back-off restarting failed container
```

[...]

PR #1955, opened 2022-06-20
fix: point to a newer openshift oauth image 4.12

frzifus:

- Closes https://github.com/jaegertracing/jaeger-operator/issues/1952

- change default settings of `openshift-oauth-proxy-image` from `openshift/oauth-proxy:latest` to `quay.io/openshift/origin-oauth-proxy:latest`.

## Additional
Do we want to keep the `latest` tag? For rollover etc. we also use `latest`. But this brings the risk that faulty external updates could break the operator functionality.

```
REPOSITORY                                                                     TAG                   IMAGE ID       CREATED        SIZE
quay.io/openshift/origin-oauth-proxy                                           latest                0b2b08dd4482   3 weeks ago    299MB
openshift/oauth-proxy                                                          latest                e7d312d68fe3   2 years ago    244MB
```


iblancasa:
Regarding your question:
> Do we want to keep the latest tag? For rollover etc. we also use latest. But this brings the risk that faulty external updates could break the operator functionality.
I think it would be good to pin to a version

PR #2020, opened 2022-08-09
Ensure log consistency between messages

### LINKED ISSUE #1817: log line inconsistency
**Describe the bug**
Seems we use plain `logrus` in some places instead of the propagated `logr.Logger` wrapper. on one hand this leads to bad readability, on the other hand i had problems to change the log level in all places at once.

**To Reproduce**
1. Deploy Operator
2. Deploy Jaeger
3. View Operator logs

**Expected behavior**

- Seeing the same format of logs in every line.
- Being able to change log-levels globally.

**Screenshots**
image

 - Jaeger version: 1.32.0

iblancasa:
- Resolves #1817

- Use the propagated `logr.Log` logger to log the messages
- Change all the calls to the log to use the new API


controllers/appsv1/deployment_webhook.go
```diff
@@ -132,7 +132,7 @@ func (d *deploymentInterceptor) Handle(ctx context.Context, req admission.Reques
 
 		const msg = "no suitable Jaeger instances found to inject a sidecar"
 		span.AddEvent(msg)
-		logger.Debug(msg)
+		logger.V(-1).Info(msg)
```
rubenvp8510, line 135: what -1 means?

iblancasa: With `logr` you don't have the semantic verbosity levels. You only have "numbers". When using `logr` with `zap`, they recommend to map the `zap` semantic with those numbers https://github.com/uber-go/zap/blob/e751939b68425b6aedd6bbf93704363c482880da/zapcore/level.go#L37-L52

frzifus (commented):
are there some constants in the `log` package that could be used instead of `.V({-1,0,1,2,3,4,..})`?

iblancasa:
It seems using constants is kind of an anti-pattern for the `logr` philosophy. I think so because I've been reading about it and everybody recommends to use the numbers directly instead of providing the verbosity levels a semantic meaning.

When I was implementing the code, I was thinking about creating some constants and use them but I was not sure since the recommendations done by other developers... I don't have a strong opinion about that.

frzifus:
mh.. they link [this](https://dave.cheney.net/2015/11/05/lets-talk-about-logging) blog post:
```
I want to take a contradictory position. I think that all logging libraries are bad because the offer too many features; a bewildering array of choice that dazzles the programmer right at the point they must be thinking clearly about how to communicate with the reader from the future; the one who will be consuming their logs.

I posit that successful logging packages need far less features, and certainly fewer levels.
```

I am surprised that at least a few aliases are not provided.

frzifus (dismissed):

iblancasa:
@frzifus could you enable auto-merge or something when the tests are passed? I don't know why but it was disabled :(

frzifus (approved):

PR #2026, opened 2022-08-11
Change verbosity level and message of the log for autoprovisioned kafka

### LINKED ISSUE #844: Log "Provisioning Kafka, this might take a while" happens on every reconcile loop
Which is misleading and should be probably logged when waiting on kafka to be up during the first deployment.

iblancasa:

- Resolves #844

- Changed message and verbosity
- The creation of the Kafka cluster will be logged when the Kafka cluster is created ([kafka.go](https://github.com/jaegertracing/jaeger-operator/blob/41df45b2bc1922ca99988ee613681c9ca45619b1/pkg/controller/jaeger/kafka.go#L51))


pkg/strategy/streaming.go
```diff
@@ -76,7 +76,7 @@ func newStreamingStrategy(ctx context.Context, jaeger *v1.Jaeger) S {
 	// we provision a Kafka when no brokers have been set, or, when we are not in the first run,
 	// when we know we've been the ones placing the broker information in the configuration
 	if (!pfound && !cfound) || provisioned {
-		jaeger.Logger().Info("Provisioning Kafka, this might take a while...")
+		jaeger.Logger().Debug("The Jaeger deployment uses autoprovisioned kafka. Kafka cluster will be deployed if not there before")
```
frzifus, line 79:
```suggestion
		jaeger.Logger().Debug("Kafka auto provisioning is enabled. A Kafka cluster will be deployed if it does not exist.")
```

frzifus (commented):
what do you think?

frzifus (dismissed):

frzifus (approved):

Issue #2032, opened 2022-08-15
[Feature]: Support container security context in Jaeger CRD


mjnagel:
### Requirement

In some environments an end user may want to enforce certain policies on all pods running in the cluster, such as requiring `drop: ALL` capabilities set in every container security context.

### Problem

Currently this is not supported in the CRD from what I can tell - while you can modify the pod level security context there is no ability to add a container level security context, which is where `capabilities` must be set.

### Proposal

Add a `containerSecurityContext` portion of the spec to support setting these options. This would live at the same level as the current `securityContext` portions, but provide the container specific options.

PR #2033, opened 2022-08-15
Add container security context support

mjnagel:
Signed-off-by: Micah Nagel <micah.nagel@parsons.com>

Resolves https://github.com/jaegertracing/jaeger-operator/issues/2032

- Adds support for `containerSecurityContext` in common spec
- `containerSecurityContext` used as the security context for containers in each of agent, all-in-one, collector, ingester, query

### DIFF
--- a/apis/v1/jaeger_types.go
+++ b/apis/v1/jaeger_types.go
@@ -220,6 +220,9 @@ type JaegerCommonSpec struct {
 	SecurityContext *v1.PodSecurityContext `json:"securityContext,omitempty"`
 
 	// +optional
+	ContainerSecurityContext *v1.SecurityContext `json:"containerSecurityContext,omitempty"`
+
+	// +optional
 	ServiceAccount string `json:"serviceAccount,omitempty"`
 
 	// +optional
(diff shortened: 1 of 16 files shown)


iblancasa (changes requested):
Please @mjnagel, could you take a look to the failed GitHub Actions Workflows?

frzifus (commented):
calling `make api-docs` should fix it.

mjnagel:
> calling `make api-docs` should fix it.

@frzifus thanks! Should be good to go now? Also rebased/squashed commits.

iblancasa (approved):

frzifus (commented):
Should we set and verfiy some `ContainerSecurityContext` fields in the agent/collector/... unittests to make sure everything works as expected?
What do you think @mjnagel ?

mjnagel:
> Should we set and verfiy some `ContainerSecurityContext` fields in the agent/collector/... unittests to make sure everything works as expected? What do you think @mjnagel ?

I should be able to add test(s) today if we want. I didn't think it was necessary at first since it's a direct 1:1 "passthrough" from the Jaeger spec -> deployment and issues would be exposed as build errors if there were a type mismatch. But more testing never hurts 😄

mjnagel:
@frzifus should be good to go now. I added 2 tests to each of agent, collector, ... that validate container security context is being set and that it overrides the default "global"/common set in the spec.

frzifus (approved):
Awesome!

PR #2034, opened 2022-08-16
Use golangci-lint for linting

esnible:

This adds golangci-lint checking and fixes all of the problems reported by golangci-lint.

The configuration file _.golangci.yml_ is nearly identical to the one we have been using for Jaeger.  The only difference is that we don't blacklist `io/ioutil`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,30 +1,15 @@
 run:
-  go: '1.17'
+  go: '1.18'
   timeout: 10m
-  skip-dirs:
-    - mocks
-    - thrift-0.9.2
-    - .*-gen
-  skip-files:
-    - ".*.pb.go$"
 
 linters-settings:
   goimports:
-    local-prefixes: github.com/jaegertracing/jaeger
-  gosec:
-    # To specify a set of rules to explicitly exclude.
-    # Available rules: https://github.com/securego/gosec#available-rules
-    excludes:
-      - G104
-      - G107
-      - G404
-      - G601
+    local-prefixes: github.com/jaegertracing/jaeger-operator
   gosimple:
     go: "1.18"
 
 linters:
   enable:
-    - contextcheck
     - depguard
     - gofmt
     - gofumpt
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,45 @@
+run:
+  go: '1.17'
+  timeout: 10m
+  skip-dirs:
+    - mocks
+    - thrift-0.9.2
+    - .*-gen
+  skip-files:
+    - ".*.pb.go$"
+
+linters-settings:
+  goimports:
+    local-prefixes: github.com/jaegertracing/jaeger
+  gosec:
+    # To specify a set of rules to explicitly exclude.
+    # Available rules: https://github.com/securego/gosec#available-rules
+    excludes:
+      - G104
+      - G107
+      - G404
+      - G601
+  gosimple:
+    go: "1.18"
+
+linters:
+  enable:
+    - contextcheck
+    - depguard
+    - gofmt
+    - gofumpt
+    - goimports
+    - gosec
+    - govet
+    - misspell
+    - bidichk
+  disable:
+    - errcheck
+
+issues:
+  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-rules:
+    # Exclude some linters from running on tests files.
+    - path: _test\.go
+      linters:
+        - gosec
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ PHONY: lint
 lint:
 	$(ECHO) Linting...
 	$(VECHO)GOPATH=${GOPATH} ./.ci/lint.sh
+	golangci-lint -v run
 
 .PHONY: vet
 vet: ## Run go vet against code.
--- a/Makefile
+++ b/Makefile
@@ -438,6 +438,7 @@ tools: kustomize controller-gen operator-sdk
 
 .PHONY: install-tools
 install-tools: operator-sdk
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.2
 	$(VECHO)${GO_FLAGS} ./.ci/vgot.sh \
 		golang.org/x/lint/golint \
 		golang.org/x/tools/cmd/goimports
--- a/controllers/appsv1/deployment_webhook.go
+++ b/controllers/appsv1/deployment_webhook.go
@@ -53,7 +53,7 @@ type deploymentInterceptor struct {
 // Handle adds a label to a generated pod if deployment or namespace provide annotaion
 func (d *deploymentInterceptor) Handle(ctx context.Context, req admission.Request) admission.Response {
 	tracer := otel.GetTracerProvider().Tracer(v1.ReconciliationTracer)
-	ctx, span := tracer.Start(ctx, "reconcileDeployment")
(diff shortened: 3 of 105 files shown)


esnible:
I modified the code coverage config file _.codecov.yml_ to allow changes that lower code coverage by <0.1%.

The linter had me remove simplify code.  The lines uncovered by tests didn't change, but the lines covered by tests were less due to the simplification.

iblancasa (changes requested):

iblancasa:
E2E tests were passed properly in OpenShift 4.10

.golangci.yml
```diff
@@ -0,0 +1,45 @@
+run:
+  go: '1.17'
```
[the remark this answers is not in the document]
esnible: Good point.  I will set it to use 1.18.  I will also clean up the _.golangci.yml_, which has things that are Jaeger specific and aren't needed for jaeger-operator.

Note that Jaeger itself uses 1.17 here.  When I set this to 1.18 I get two messages:
```
WARN [linters context] structcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649. 
WARN [linters context] contextcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649. 
```

We don't need those linters, but perhaps that is why Jaeger itself did not bump to Go version used for linting.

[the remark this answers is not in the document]
iblancasa: I see. We can enable right now just the very minimum and enable other stuff later, fixing the reported warnings

Makefile
```diff
@@ -437,6 +438,7 @@ tools: kustomize controller-gen operator-sdk
 
 .PHONY: install-tools
 install-tools: operator-sdk
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.2
```
frzifus, line 441: Does it make sense to install the linter maybe in our local `$(REPO)/bin` foulder? otherwise we overwrite the version installed in the `$GOPATH/bin`, right?

controllers/appsv1/deployment_webhook.go
```diff
@@ -53,7 +53,7 @@ type deploymentInterceptor struct {
 // Handle adds a label to a generated pod if deployment or namespace provide annotaion
 func (d *deploymentInterceptor) Handle(ctx context.Context, req admission.Request) admission.Response {
 	tracer := otel.GetTracerProvider().Tracer(v1.ReconciliationTracer)
-	ctx, span := tracer.Start(ctx, "reconcileDeployment")
+	_, span := tracer.Start(ctx, "reconcileDeployment")
```
frzifus, line 56: Does `tracer.Start` modify or extend the given context in some way? Iam asking because we might lose information by using the old ctx, right?

frzifus (commented):
awesome @esnible !

Makefile
esnible: I copied this from https://github.com/jaegertracing/jaeger/blob/main/Makefile#L392 ; assuming it was correct there.

Are you concerned about the Github builders, or are you concerned about people running `make install-tools` on their laptop?

controllers/appsv1/deployment_webhook.go
esnible: At least some implementations modify the context by `calling context.WithValue(...)`.

I believe the new code is better than the old code.  The old code ignored the value reported returned into `ctx`, but it wasn't obvious upon reading it.  The new code makes it clear the modified context gets thrown away.

If we need to re-use the context, we should do that in a follow-up, not in this PR which is only to introduce `golangci-lint`.

iblancasa (dismissed):

iblancasa (approved):

frzifus (approved):

controllers/appsv1/deployment_webhook.go
pavolloffay: > The new code makes it clear the modified context gets thrown away.

This is a large method and somebody editing the code might not notice that the current context is thrown away. In this case a nolint annotation would be better.

controllers/appsv1/deployment_webhook.go
```diff
@@ -165,7 +163,7 @@ func (d *deploymentInterceptor) InjectDecoder(decoder *admission.Decoder) error
 
 func reconcileConfigMaps(ctx context.Context, cl client.Client, jaeger *v1.Jaeger, dep *appsv1.Deployment) error {
 	tracer := otel.GetTracerProvider().Tracer(v1.ReconciliationTracer)
-	ctx, span := tracer.Start(ctx, "reconcileConfigMaps")
+	_, span := tracer.Start(ctx, "reconcileConfigMaps")
```
pavolloffay, line 166: The same as previous comment - the context should be overridden.

controllers/appsv1/deployment_webhook.go
```diff
@@ -187,7 +185,7 @@ func reconcileConfigMaps(ctx context.Context, cl client.Client, jaeger *v1.Jaege
 
 func reconcileConfigMap(ctx context.Context, cl client.Client, cm *corev1.ConfigMap, dep *appsv1.Deployment) error {
 	tracer := otel.GetTracerProvider().Tracer(v1.ReconciliationTracer)
-	ctx, span := tracer.Start(ctx, "reconcileConfigMap")
+	_, span := tracer.Start(ctx, "reconcileConfigMap")
```
pavolloffay, line 188: The same as previous comment - the context should be overridden.

controllers/appsv1/deployment_webhook.go
pavolloffay: The ctx down in this function is propagated to other calls which might generate spans.

pkg/cmd/start/bootstrap.go
```diff
@@ -69,12 +69,11 @@ func init() {
 }
 
 func bootstrap(ctx context.Context) manager.Manager {
-
 	namespace := getNamespace(ctx)
 	tracing.Bootstrap(ctx, namespace)
 
 	tracer := otel.GetTracerProvider().Tracer(v1.BootstrapTracer)
-	ctx, span := tracer.Start(ctx, "bootstrap")
+	_, span := tracer.Start(ctx, "bootstrap")
```
pavolloffay, line 76: The same as previous comment - the context should be overridden.

pavolloffay (commented):
The context of `tracer.Start` shouldn't be thrown away. It's needed for correct tace context propagation.

Makefile
frzifus: Discussed with @iblancasa - It seem we have several global software installations. We will address this later in a separate patch.

pkg/cmd/start/bootstrap.go
```diff
@@ -230,7 +229,7 @@ func detectNamespacePermissions(ctx context.Context, mgr manager.Manager) {
 
 func setOperatorScope(ctx context.Context, namespace string) {
 	tracer := otel.GetTracerProvider().Tracer(v1.BootstrapTracer)
-	ctx, span := tracer.Start(ctx, "setOperatorScope")
+	ctx, span := tracer.Start(ctx, "setOperatorScope") // nolint:ineffassign,staticcheck
```
pavolloffay, line 232: The other way to make the linter happy would be to define a span var and then use `=` on this line. This way we don't have to disable linting. 

I would maybe prefer the version that passes linter, but it's not great either. I will leave it up to you.

PR #2045, opened 2022-08-23
Upgrade golangci-lint, fix reported issues and remove security make rule

iblancasa:

- Ugraded `golangci-lint` from `1.46.2` to `1.48.0`
- Removed `security` unneeded make rule
- Fix new reported issues


tests/assert-jobs/utils/elasticsearch/elasticsearch.go
```diff
@@ -221,7 +221,7 @@ func executeEsRequest(es EsConnection, httpMethod, api string, body []byte) ([]b
 			Certificates: []tls.Certificate{es.Certificate},
 			RootCAs:      es.RootCAs,
 		}
-		tlsConfig.BuildNameToCertificate()
+
```
frzifus, line 224: What is the effect of this change?

iblancasa: It seems in newer versions is not needed anymore. I ran all the E2E tests in OpenShift and they passed

frzifus (approved):

PR #2051, opened 2022-08-24
Cleanup tools dependencies installation

iblancasa:
- Add some unification to how the tools are installed
- Fix `envtest` installation
- Download some tools (like `operator-sdk`) only if they were not downloaded
- Fix some make dependency rules
- Remove `golint` usage since we have `golangci-lint`
- Allow usage of tools installed in other paths via Makefile vars (like `KIND` or `KUSTOMIZE`)


Makefile
```diff
@@ -337,9 +341,10 @@ CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
 	$(VECHO)./hack/install/install-controller-gen.sh
 
-ENVTEST = $(shell pwd)/bin/setup-envtest
-envtest: ## Download envtest-setup locally if necessary.
-	$(VECHO) GOBIN=$(shell pwd)/bin go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+.PHONY: envtest
+envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
+$(ENVTEST): $(LOCALBIN)
+	test -s $(ENVTEST) || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
frzifus, line 347: would it make sense to use here a specific version?

iblancasa: I had the same thought. But look at [the instructions to upgrade to `operator-sdk` `1.23.0`](https://v1-23-x.sdk.operatorframework.io/docs/upgrading-sdk-version/v1.23.0/#gov3-fix-makefile-targets-to-not-download-binaries-when-they-are-found-already): it seems it's expected (or good practice or something) to use `latest`

frzifus (approved):

PR #2055, opened 2022-08-25
Migrate autoscaling v2beta2 to v2 for Kubernetes 1.26

### LINKED ISSUE #1965: Migrate autoscaling v2beta2 to v2 for Kubernetes 1.26
> The autoscaling/v2beta2 API version of HorizontalPodAutoscaler will no longer be served in v1.26.

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v126

> Migrate manifests and API clients to use the autoscaling/v2 API version, available since v1.23.

As we need to support k8s 1.19 I would suggest keep using v2beta2 for k8s 1.19-1.22 and then use v2 on newer versions.

iblancasa:
- Resolves #1965

- Migrate `autoscaling` API from `v2beta2` to `v2`


pkg/autodetect/main.go
```diff
@@ -131,6 +131,30 @@ func (b *Background) detectCronjobsVersion(ctx context.Context) {
 	)
 }
 
+func (b *Background) detectAutoscalingVersion(ctx context.Context) {
+	apiGroupVersions := []string{"autoscaling/v2", "autoscaling/v2beta2"}
```
frzifus, line 135: Does it make sense to use the `api/v1` constants here?

pkg/autodetect/main.go
```diff
@@ -131,6 +131,30 @@ func (b *Background) detectCronjobsVersion(ctx context.Context) {
...
+		if err != nil {
+			log.Log.V(-1).Info(
+				fmt.Sprintf("Error getting %s api list: %s", apiGroupVersion, err),
+			)
+			continue
+		}
+		for _, api := range groupAPIList.APIResources {
+			if api.Name == "horizontalpodautoscalers" {
+				viper.Set(v1.FlagAutoscalingVersion, apiGroupVersion)
+				log.Log.V(-1).Info(fmt.Sprintf("Found the horizontalpodautoscalers api in %s", apiGroupVersion))
```
frzifus, line 147: lets start error messages with a lower case letter :)

frzifus (commented):
mentioned a few things :+1: 

@kevinearls @rubenvp8510 could you have a look too, please?

pkg/autodetect/main.go
frzifus: @iblancasa there are a few more error messages starting with an upper case letter ^^

iblancasa: Let's see if now is correct

frzifus:
@kevinearls @rubenvp8510 could one of you have another look please? :)

kevinearls:
@frzifus I intended to take a look over the next few days.  Apparantly @iblancasa is as bad as you when it comes to understanding the meaning of "being on vacation" though.  :grin:  

I will take another look later today.

kevinearls (approved):
LGTM.  You might want to add an e2e test though.  I added one for similar changes to the opentelemetry-operator.  See https://github.com/open-telemetry/opentelemetry-operator/pull/1014

iblancasa:
@kevinearls I don't see any new E2E tests in the PR you referred. Only unit tests.

Also, as soon as we added the new Kubernetes versions to the testing, we'll be checking this feature.

frzifus:
ok, then we go ahead?

iblancasa:
@frzifus I think yes, we can go ahead.

frzifus:
perfect :)

PR #2077, opened 2022-09-26
Add liveness and readiness probes to injected sidecar

### LINKED ISSUE #1202: Readiness probes missing from the injected sidecar container

MacroPower:
Signed-off-by: Jacob Colvin <jacobcolvin1@gmail.com>

Fixes #1202

This PR is mainly just taking code from #1230 which was unfortunately abandoned.

I believe I've fixed all of the issues with #1230, which involved:
- Changing `Handler` to `ProbeHandler`.
- Skipping checks for conflicting ports on the agent container. (Without this, it seems that probes could not be added if a deployment was restarted/upgraded, because the port was already used on the deployment from the last injection.)

### DIFF
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -10,6 +10,7 @@ import (
 	"github.com/spf13/viper"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	v1 "github.com/jaegertracing/jaeger-operator/apis/v1"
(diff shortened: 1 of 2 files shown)


frzifus (approved):
thanks @MacroPower! LGTM what do you think @rubenvp8510 ?

pkg/inject/sidecar.go
```diff
@@ -309,6 +310,31 @@ func container(jaeger *v1.Jaeger, dep *appsv1.Deployment, agentIdx int) corev1.C
...
+
+	if isContainerPortAvailable(adminPort, dep) {
+		containerDefinition.LivenessProbe = &corev1.Probe{
+			ProbeHandler: corev1.ProbeHandler{
+				HTTPGet: &corev1.HTTPGetAction{
+					Path: "/",
+					Port: intstr.FromInt(int(adminPort)),
+				},
+			},
+			InitialDelaySeconds: 5,
```
rubenvp8510, line 322: Why this values? where it comes from?

rubenvp8510:
small question, overall looks good.

pkg/inject/sidecar.go
MacroPower: @rubenvp8510, all those values were carried over from the previous PR, so I'm not completely sure of why they were originally set that way. I tested on a few clusters and they seemed to be fine, so I didn't really see a need to mess with them. But if you have any suggestions, I'd be happy to explore further.

PR #2078, opened 2022-09-29
Fix Kafka tests for old Kafka operator version

iblancasa:

- Newer Kafka Operator versions deploy `StrimziPodSet` instead of `StatefulSet`. This can be a problem if we want to do something with an older version of the Kafka Operator

### DIFF
--- a/tests/e2e/render-utils.sh
+++ b/tests/e2e/render-utils.sh
@@ -866,6 +866,17 @@ if [ ! -z "$output" ]; then
     IS_OPENSHIFT=true
 fi
 
+export KAFKA_USE_CUSTOM_PODSET
+if [ -z "$KAFKA_VERSION" ]; then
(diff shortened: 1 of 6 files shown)


tests/e2e/render-utils.sh
```diff
@@ -866,6 +866,17 @@ if [ ! -z "$output" ]; then
     IS_OPENSHIFT=true
 fi
 
+export KAFKA_USE_CUSTOM_PODSET
+if [ -z "$KAFKA_VERSION" ]; then
```
frzifus, line 870: Can we get rid of the nested if with something like this?
```suggestion
if [ -z "$KAFKA_VERSION" ] || [ version_gt $KAFKA_VERSION "0.25.0" ]; then
```

pavolloffay (approved):

PR #1950, opened 2022-06-16
Add E2E tests for OTLP ingestion feature

iblancasa:
## Which problem is this PR solving?
- Jaeger can now process OTLP traces. #1948 introduces this feature in the Jaeger Collector. This PR adds E2E tests for this new feature

## Short description of the changes
- Added E2E tests to ensure the OTLP feature works on Jaeger collector. 4 new tests are run:
- - Jaeger deployed as `allInOne` using `http` for the OTEL exporter
- - Jaeger deployed as `production` using `http` for the OTEL exporter
- - Jaeger deployed as `allInOne` using `grcp` for the OTEL exporter
- - Jaeger deployed as `production` using `grcp` for the OTEL exporter
- Other fixes:
- - Added dependency `kind` to `load-operator-image` make rule to ensure the `KIND` make variable is populated

# Blocked by:
- #1949: it depends on Jaeger 1.35.x in order to work
- #1948: it depends on the feature in order to pass the tests

PR #2084, opened 2022-10-09
Update otel packages to v1.10.0+v0.32.1

### LINKED ISSUE #1966: Upgrade otel packages to the latest version
## Requirement - what kind of business use case are you trying to solve?
Upgrade the otel dependency packages to the latest version.

Current used version of the otel packages is v0.20.0. This version was published Apr 23, 2021. We should upgrade this dependency to the latest version (v1.7.0 when this issue was opened)

volmedo:
Closes #1966

Update OpenTelemetry packages to v1.10.0+v0.32.1, which were published a couple weeks ago.

## Additional considerations
Updating the packages was a bit more involved than anticipated since the API changed substantially. Besides, I wasn't familiar with the previous nor the current design, so please review the changes carefully.


pkg/metrics/bootstrap.go
```diff
@@ -30,28 +23,17 @@ func Bootstrap(ctx context.Context, namespace string, client client.Client) erro
...
-	c := controller.New(
-		processor.New(
-			selector.NewWithHistogramDistribution(
-				histogram.WithExplicitBoundaries(config.DefaultHistogramBoundaries),
-			),
-			export.CumulativeExportKindSelector(),
-			processor.WithMemory(true),
-		),
-		controller.WithResource(resource.NewWithAttributes([]attribute.KeyValue{}...)),
-	)
```
volmedo, line 45: this is one of the most controversial changes I made. I looked through the whole code base and it looks that there are no traces of controllers, processors and selectors anymore. At the same time, I could not find a sensible equivalent or a place where these settings could be configured. A documentation about migrating from the old to the new API would've been great.

I'm not sure if they are not needed anymore or the new packages are using some defaults that are better suited and allow removing the need for configuration on client code.

tests/assert-jobs/reporter-otlp/main.go
```diff
@@ -92,74 +89,42 @@ func initProvider(serviceName string) func() {
...
-		sdktrace.WithSpanProcessor(bsp),
-	)
-
-	cont := controller.New(
-		processor.New(
-			simple.NewWithExactDistribution(),
-			exp,
-		),
-		controller.WithExporter(exp),
-		controller.WithCollectPeriod(2*time.Second),
```
volmedo, line 140: what I said [her](https://github.com/jaegertracing/jaeger-operator/pull/2084#discussion_r990806809)e also applies to these changes.

volmedo:
I issued `make unit-tests` locally and all tests are passing. I'm relaying on the checks in CI for E2E tests to avoid doing a full setup on my environment. Is that a good approach? Something else I'd need do to increase confidence in the changes not breaking anything?

pavolloffay:
@volmedo did you check if traces and metrics are reported?

volmedo:
> It's just my opinion, but basically it's ok to use the CI pipeline for a test, as long as you assume that everything should pass.

@frzifus sure, "rely on the CI pipeline" means that all tests are passing 🙂👍🏻

> did you check if traces and metrics are reported?

@pavolloffay just up to the extent that unit tests and E2E tests in the CI pipeline check they are being reported. Are the E2E tests not enough of a guarantee that everything is working? I haven't installed and configured `minikube` locally but I can resort to manual testing if the automatic checks in place cannot be fully trusted (although I'd suggest improving them if that's the case).

volmedo:
Talking about having all the tests in CI pass, I see some miscellaneous E2E tests are failing
```
+----------------------------------------------+--------+
| miscellaneous/set_custom_img                 | passed |
| miscellaneous/collector_otlp_allinone_http   | failed |
| miscellaneous/collector_otlp_allinone_grpc   | passed |
| miscellaneous/collector_autoscale            | passed |
| miscellaneous/cassandra_spark                | passed |
| miscellaneous/istio                          | passed |
| miscellaneous/outside_cluster                | passed |
| miscellaneous/collector_otlp_production_http | failed |
| miscellaneous/collector_otlp_production_grpc | passed |
+----------------------------------------------+--------+

```

It's curious that only the HTTP version of the tests is failing, while the gRPC one is working. I do see some `unable to get metrics for resource [memory | cpu]`, but it looks like they are warnings and they are present in both the HTTP and the gRPC tests and I have seen them in other successful pipeline runs.

I found this error that is common to both HTTP tests and specific to them:
```
case.go:363: resource Job:kuttl-test-intimate-jaybird/check-span: .status.succeeded: key is missing from map
```

So it seems the `check-span` test fails for the HTTP OTLP provider, which clearly points to [these changes](https://github.com/jaegertracing/jaeger-operator/blob/63df6a4c33a293e96701ca6207bd749ceb9e143c/tests/assert-jobs/reporter-otlp/main.go#L92-L121). As I said before, it's surprising it fails for HTTP but works for gRPC when the configuration for both is that similar.

Any ideas on what the problem may be?

pavolloffay:
TBH I am not sure if there are e2e test that verify telemetry collector. @iblancasa , @rubenvp8510 or @frzifus should know more on this.

iblancasa:
@pavolloffay yes. The `collector-otlp-production-http`, `collector-otlp-production-grpc`, `collector-otlp-allinone-http` and `collector-otlp-allinone-grpc` E2E test verify the OpenTelemetry. They were added as part of #1950, to test automatically the changes introduced in #1948.

@volmedo, the instructions to run the E2E tests are located in the [CONTRIBUTING.md](https://github.com/jaegertracing/jaeger-operator/blob/main/CONTRIBUTING.md#e2e-tests) file. The automation should download all the dependencies needed to build and run the tests. To run the tests that are failing in this PR, you can run:

```sh
$ make run-e2e-tests-miscellaneous VERBOSE=true KUTTL_OPTIONS="--test collector-otlp-production-http"
```

volmedo:
thanks @iblancasa. After tinkering with some of the scripts (I'm on macos) I was able to run the tests locally. However, I got no more information than what I got from CI. The `collector-otlp-production-http` test is failing at the `2-smoke-test` step. I think the reason it's failing is:
```
case.go:363: resource Job:kuttl-test-teaching-martin/check-span: .status.succeeded: key is missing from map
```

I'm not sure what that means or where I should look to try to debug it. I still think [this comment](https://github.com/jaegertracing/jaeger-operator/pull/2084#discussion_r990807128) could have something to do with the issue, although it's surprising that the same tests are working when using `gRPC`.

Could I get some help on this?

iblancasa:
@volmedo the reason why the test fails, as you said, is the job is not success. You can run the test in your machine and check what is happening in that job (you need to find the pod running that job and check for its logs).

volmedo:
Ok, it took a couple debugging sessions but I finally found the issue. From some versions ago, the way the OTLP exporter interprets the `OTEL_EXPORTER_OTLP_ENDPOINT` changed from just using its value as is to [parsing it with `url.Parse`](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlptrace/internal/otlpconfig/envconfig.go#L57-L66). That means that one needs to be much more careful now when setting these environment variables. It also means that the current config:
```
OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=my-jaeger-collector-headless
```
doesn't work anymore, as it gets parsed as:
```
u, _ := url.Parse("my-jaeger-collector-headless")
fmt.Printf("%#v\n", u)
// Prints
// &url.URL{
//     Scheme:"", 
//     Opaque:"", 
//     User:(*url.Userinfo)(nil), 
//     Host:"", 
//     Path:"my-jaeger-collector-headless", 
//     RawPath:"", 
//     OmitHost:false, 
//     ForceQuery:false, 
//     RawQuery:"", 
//     Fragment:"", 
//     RawFragment:""
// }
```

As the value in the variable is not a full URL, it is parsed as just a path. After that, we set the endpoint to that same value plus the corresponding port, which finally results in this URL to report traces to:
```
http://my-jaeger-collector-headless:4318/my-jaeger-collector-headless/v1/traces
```
which obviously returns a 404.

The proper fix is to use full URLs in env variables and omit explicit configuration via options when creating the exporter as they are not needed anymore.

Surprisingly enough, the parsing for the gRPC exporter is done in a slightly different way and it's not by this, but just because the concept of path doesn't exist there.

volmedo:
Hi folks, I think this work is still relevant. The change in the API was deep enough to make this more than a simple package upgrade. And now that the package is in a 1.x.x version, API changes are less likely to take place.

IMO it would be sad this PR falls through the cracks, and the only thing left is having someone do a review. All required tests are passing, which (if you trust your test suites) should indicate the changes can be merged safely.

frzifus:
thank you @volmedo thats great! i will give it a try during the week and if there is no major issue - we can go ahead :)

pkg/metrics/instances_test.go
```diff
@@ -31,22 +30,40 @@ type expectedMetric struct {
...
+	for _, dp := range gauge.DataPoints {
+		if expectedAttrSet.Equals(&dp.Attributes) {
+			matchingDP = dp
+			found = true
+			break
+		}
+	}
+
+	assert.True(t, found, "Metric %s doesn't have expected attributes %v", expectedAttrs)
+	assert.Equal(t, expectedValue, matchingDP.Value,
```
frzifus, line 65: I am glad that there are tests. :smile:

pkg/metrics/bootstrap.go
frzifus: >  A documentation about migrating from the old to the new API would've been great. 

Unfortunately, I very much doubt it. Probably here we would have to follow the otel change log of the last year. :face_exhaling: 

> I'm not sure if they are not needed anymore or the new packages are using some defaults that are better suited and allow removing the need for configuration on client code. 

Ive here no good answer. I made a local tour and could not detect any abnormalities. @rubenvp8510 what are your thoughts?

tests/e2e/render-utils.sh
```diff
@@ -90,8 +90,14 @@ function render_otlp_smoke_test() {
...
 
+    if [ "$reporting_protocol" = "grpc" ]; then
+        reporting_port=":4317"
+    else
+        reporting_port=":4318"
+    fi
+
     export JAEGER_QUERY_ENDPOINT="$protocol$jaeger-query$query_port"
-    export OTEL_EXPORTER_OTLP_ENDPOINT="$jaeger-collector-headless"
+    export OTEL_EXPORTER_OTLP_ENDPOINT="$protocol$jaeger-collector-headless$reporting_port"
```
frzifus, line 100: Do we need to consider this change anywhere else?

frzifus:
@volmedo thanks for your work! i would like to wait for a 2nd review from @rubenvp8510. But then i assume we are ready to go!

PS: tried to bring this branch to a current state and broke something. trying to fix it right now.

frzifus:
mhpf.. @volmedo could you please update your branch? for me it doesnt work :octocat:

volmedo:
I rebased my branch on master and sorted out the conflicts with the dependencies. Everything looks so green now 😄.

frzifus (approved):
Awesome, thank you @volmedo !

wdyt @rubenvp8510 ?

rubenvp8510 (approved):
LGTM! Good work!

tests/e2e/render-utils.sh
rubenvp8510: No as far as I know, but israel should know more about this e2e tests :)

volmedo:
Great seeing this merged! 🎉

Issue #1487, opened 2021-06-24
Cassandra schema job is failed when `options.cassandra.servers` has multi server address list


anarcher:
**Describe the bug**
Cassandra schema job is failed when `options.cassandra.servers` has multi-server address list (Docker image version is: `jaeger-operator:1.22.1`)

logs from `jaeger-cassandra-schema-job`

```
socket.gaierror: [Errno -2] Name or service not known
Cassandra is still not up at scylla-jaegerdb-dc1-a1-0,scylla-jaegerdb-dc1-a1-1,scylla-jaegerdb-dc1-a1-2:9042. Waiting 1 second.
Traceback (most recent call last):
  File "/opt/cassandra/bin/cqlsh.py", line 2458, in <module>
    main(*read_options(sys.argv[1:], os.environ))
  File "/opt/cassandra/bin/cqlsh.py", line 2436, in main
    encoding=options.encoding)
  File "/opt/cassandra/bin/cqlsh.py", line 484, in __init__
    load_balancing_policy=WhiteListRoundRobinPolicy([self.hostname]),
  File "/opt/cassandra/bin/../lib/cassandra-driver-internal-only-3.11.0-bb96859b.zip/cassandra-driver-3.11.0-bb96859b/cassandra/policies.py", line 417, in __init__
socket.gaierror: [Errno -2] Name or service not known
Timed out waiting for Cassandra.
```

`jaeger` CRD

```
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: jaeger
spec:
  allInOne:
    options:
      collector.zipkin.host-port: "9411"
      log-level: debug
  storage:
    cassandraCreateSchema:
      datacenter: dc1
      enabled: true
      mode: test
    options:
      cassandra:
        keyspace: jaeger_v1_dc1
        servers: scylla-jaegerdb-dc1-a1-0,scylla-jaegerdb-dc1-a1-1,scylla-jaegerdb-dc1-a1-2
    type: cassandra
  strategy: allinone
```

**To Reproduce**
1.  
2.
3.

**Expected behavior**

Successfully done of Cassandra schema job

**Screenshots**

**Version (please complete the following information):**
Docker image version is: `jaeger-operator:1.22.1`

**What troubleshooting steps did you try?**
If `cassandra.servers` has single server address, This job runs successfully.

**Additional context**
I am not sure but it is ok for me  that  choosing first `cassandra.servers` when creating job .
What do you think about it? :-> 

[...]

PR #2087, opened 2022-10-11
choose first server address

Efrat19:
Signed-off-by: Efrat19 <efrat890089@gmail.com>

Addressing https://github.com/jaegertracing/jaeger-operator/issues/1487

I think I need some further guidance regarding the unit tests for this one, since [cassandra_dependencies_test.go](https://github.com/jaegertracing/jaeger-operator/blob/main/pkg/storage/cassandra_dependencies_test.go) doesnt seem to cover cassandra properties 

@jpkrohling


pkg/storage/cassandra_dependencies.go
```diff
@@ -44,7 +45,8 @@ func cassandraDeps(jaeger *v1.Jaeger) []batchv1.Job {
 		Value: jaeger.Spec.Storage.CassandraCreateSchema.Datacenter,
 	}}
 
-	host := jaeger.Spec.Storage.Options.StringMap()["cassandra.servers"]
+	servers := jaeger.Spec.Storage.Options.StringMap()["cassandra.servers"]
+	host    := strings.Split(servers,",")[0]  //  <- choose first server address
```
frzifus, line 49: As far as I can see, we are currently returning a `batchv1.job` with the provided `cassandra.servers` address. We would need to create one job per host address. The easiest way would probably be to put the rest of the code of the function into a loop. A more elegant and more testable solution is probably to put the remaining part into an extra function.

frzifus (commented):
Hi @Efrat19, thanks for taking care of that issue. 

> I think I need some further guidance regarding the unit tests for this one, since [cassandra_dependencies_test.go](https://github.com/jaegertracing/jaeger-operator/blob/main/pkg/storage/cassandra_dependencies_test.go) doesnt seem to cover cassandra properties

I assume the current change is probably covered by an integration test. Do you want to  extended your pr that multiple addresses are considered?

pkg/storage/cassandra_dependencies.go
Efrat19: Hey, the func already returns an array `batchv1.job` and the changes remain within its scope, so it seems simple.
I do have a question though -
Once we create a job per host passed from `cassandra.servers`, shouldn't we support different port, user and password for each of them? 
and though its possible to accomplish that, it would require the user to enter hosts, ports and creds in comma-separated lists, something like (when I have to servers `server1:9042` and `server2:9043`)
```yaml
cassandra:
  servers: server1,server2
  port(s): 9042,9043
  ....
```
which seems a bit odd to me unless we change the mapping to connection strings, but that would require far more refactoring.
I can make the changes to create a job per host, but all hosts will have to go by the same port, user and password, unless you have better ideas?

frzifus: > Hey, the func already returns an array batchv1.job and the changes remain within its scope

But the list always contains only [one entry](https://github.com/jaegertracing/jaeger-operator/blob/main/pkg/storage/cassandra_dependencies.go#L154-L196), regardless of the number of server addresses. I thought thats what we want to support. :sweat_smile: 

> Once we create a job per host passed from cassandra.servers, shouldn't we support different port, user and password for each of them? ... I can make the changes to create a job per host, but all hosts will have to go by the same port, user and password, unless you have better ideas?

These are indeed good arguments. I don't really have much experience with cassandra, but I would imagine that not only connection strings and credentials are different, but also certificates used and other options.

Which could be solved by own instance options:
```yaml
cassandra:
  servers: server1
  port(s): 9042
  ....
cassandra/2:
  servers: server2
  port(s): 9043
  ....
```

I don't know how useful that is when we query data later. We could use the first entry as "primary" store. But then it would probably be easier to support a storage list.

```yaml
kind: Jaeger
metadata:
  name: simple-prod
spec:
  strategy: production
  storage:
  ...
  storage/backup:
  ...
```

Better we go with the current change and discuss such things in an extra ticket. :facepalm:

frzifus (approved):

pkg/storage/cassandra_dependencies.go
Efrat19: OK cool
So for now I extracted the validations and job building to a new function I called `cassandraDepsPerHost`, while the original function `cassandraDeps` loops through the hosts list and calls `cassandraDepsPerHost` for each one of them.
I think I can improve its efficiency but just wanted to make sure I am heading the right direction here, what do you think

Efrat19: Of course I can revert back if you preferred to leave the validation change only

frzifus: sounds reasonable. Could we add another test to verify that it produces 2 different jobs? 
Preferably also an integration test. @iblancasa could you help with this?

iblancasa: I don't fully what is the effect of this PR. Please, could you give me a little description? Thanks! :)

Efrat19: Sure @iblancasa the impact of the PR is that the function [`cassandraDeps`](https://github.com/Efrat19/jaeger-operator/blob/1ef0f77a61a744cba891f795b9d130a91ab48ebc/pkg/storage/cassandra_dependencies.go#L16) which used to return a `[]batchv1.job` array with a fixed length of 1 (a single cassandra job) will still return a `[]batchv1.job` but it will be an array of multiple jobs based on the amount of hosts passed to `cassandra.servers`.

Efrat19: Hey the E2Es seem to fail I have no idea why can I get some help?
Alternatively I can revert back to the validation commit and extract these changes to another PR what do you say @frzifus

frzifus: Hi @Efrat19, seems that some kind of cassandra deployment is not found. Could it be caused by the renaming we did here?
```golang
// case.go:363: jobs.batch "test-spark-deps-cassandra-schema-job" not found
truncatedName := util.Truncate("%s-cassandra-schema-job", 63, jaeger.Name, host)
```

> Alternatively I can revert back to the validation commit and extract these changes to another PR what do you say

works both for me. :+1: 

---

Logs: [examples_with_cassandra](https://github.com/jaegertracing/jaeger-operator/actions/runs/3287100490/jobs/5415966164#step:3:2814)
```bash
logger.go:42: 05:58:02 | examples-with-cassandra/0-install | test step completed 0-install
logger.go:42: 05:58:02 | examples-with-cassandra/1-install | starting test step 1-install
logger.go:42: 05:58:02 | examples-with-cassandra/1-install | Jaeger:kuttl-test-noble-jaguar/with-cassandra created
logger.go:42: 06:03:33 | examples-with-cassandra/1-install | test step failed 1-install
case.go:361: failed in step 1-install
case.go:363: deployments.apps "with-cassandra" not found
logger.go:42: 06:03:33 | examples-with-cassandra | Failed to collect events for examples-with-cassandra in ns kuttl-test-noble-jaguar: no matches for kind "Event" in version "events.k8s.io/v1beta1"
```

Logs: [cassandra_spark](https://github.com/jaegertracing/jaeger-operator/actions/runs/3287100488/jobs/5415966302#step:3:2239)
```bash
logger.go:42: 06:02:20 | cassandra-spark/0-install | test step completed 0-install
logger.go:42: 06:02:20 | cassandra-spark/1-install | starting test step 1-install
logger.go:42: 06:02:20 | cassandra-spark/1-install | Jaeger:kuttl-test-clear-turtle/test-spark-deps created
logger.go:42: 06:07:51 | cassandra-spark/1-install | test step failed 1-install
case.go:361: failed in step 1-install
case.go:363: deployments.apps "test-spark-deps" not found
case.go:363: jobs.batch "test-spark-deps-cassandra-schema-job" not found
case.go:363: no resources matched of kind: batch/v1, Kind=Job
logger.go:42: 06:07:51 | cassandra-spark | Failed to collect events for cassandra-spark in ns kuttl-test-clear-turtle: no matches for kind "Event" in version "events.k8s.io/v1beta1"
```

frzifus (approved):
thanks @Efrat19

Issue #1360, opened 2021-01-12
Bad naming convention for service name


lujiajing1126:
Currently the naming convention for the collector service is determined in this function,
https://github.com/jaegertracing/jaeger-operator/blob/010535a2ac314a0106ae0ac31ea136b85dcb7f26/pkg/service/collector.go#L97

which results in some name like `grpc-http(s)`.

However, this will cause confusing for istio as documented in the [`protocol-selection`](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) section, from which I excerpt as follows,

- By the name of the port: `name: <protocol>[-<suffix>]`.
- In Kubernetes 1.18+, by the `appProtocol` field: `appProtocol: <protocol>`.

The result of this bad name is severe, the istio-sidecar will automatically convert grpc protocol to http one which cannot be correctly processed by `collector` service.

Thus, I suggest changing the naming convention to follow the istio rules. I would like to submit a PR to resolve this issue.

jpkrohling:
Could you provide a test for the case you described? In any case, a PR would certainly be appreciated.

lujiajing1126:
Attached there are two captured tcpdump logs for the scenarios with or without istio-sidecar injected to the application pod,
(while in the same time, we do not inject istio-sidecar for jaeger-collector  pod by default)

- [jaeger-agent-with-istio-sidecar.log](https://github.com/jaegertracing/jaeger-operator/files/5801124/jaeger-agent-with-istio-sidecar.log)
- [jaeger-agent-without-istio-sidecar.log](https://github.com/jaegertracing/jaeger-operator/files/5801125/jaeger-agent-without-istio-sidecar.log)

As we can see, if the istio-sidecar is injected, the grpc stream is transcoded to http request automatically.

I suppose I can use this function to reproduce the issue? Or we need a full example?

https://github.com/istio/istio/blob/68ebdc1f1eb9fe832e6ddc26962c0630f9275de1/pkg/config/kube/conversion.go#L47-L79

jpkrohling:
So, this is a problem only for pods with both Jaeger and Istio sidecars? @jkandasa do we have tests for this scenario?

lujiajing1126:
> So, this is a problem only for pods with both Jaeger and Istio sidecars? @jkandasa do we have tests for this scenario?

Exactly.  Specifically,

```
| Pod: Application --(spans)--> Jaeger Agent --(network intercepted by)--> Istio Sidecar | 
[...]

PR #1862, opened 2022-04-19
Add http- prefix to port names in collector and agent services

cnvergence:
Signed-off-by: Karol Szwaj <karol.szwaj@gmail.com>

## Which problem is this PR solving?
- https://github.com/jaegertracing/helm-charts/issues/344#issuecomment-1100166786 (partially)
## Short description of the changes
Adding a prefix to service port names, resolves issues with istioctl 
```bash
Info [IST0118] (Service tracing-jaeger-agent.kyma-system) Port name config-rest (port: 5778, targetPort: 5778) doesn't follow the naming convention of Istio port.
Info [IST0118] (Service tracing-jaeger-collector-headless.kyma-system) Port name c-tchan-trft (port: 14267, targetPort: 14267) doesn't follow the naming convention of Istio port.
Info [IST0118] (Service tracing-jaeger-collector.kyma-system) Port name c-tchan-trft (port: 14267, targetPort: 14267) doesn't follow the naming convention of Istio port.
```
Potentially possible problems similar to this old [issue](https://github.com/jaegertracing/jaeger-operator/issues/907).

PR #1368, opened 2021-01-18
Fix service port naming convention

lujiajing1126:
* [x] Reproduce the issue with e2e test
* [x] Fix unit tests
* [x] Fix issue #1360

PR #2101, opened 2022-10-20
Fix: svc port doesnt match istio convention

frzifus:
This pr changes the service port naming to avoid TLS termination issues when using istio. Istio recognizes which application protocol is used by the port name. Our last name change of the service port name from `https-grpc` to `grcp-https` has resulted in TLS traffic being treated like unencrypted traffic.
Resulting in a handshake error: `authentication handshake failed: tls: first record does not look like a TLS handshake`.

In the analysis, the port names confused me several times. Since I didn't want to just switch grpc and https again, I tried to make the names a bit more descriptive. Note the application protocol is determined by the port name prefix: `<protocol>[-<suffix>]`.
- `grpc-https` (invalid) -> `https-grpc` (valid) -> `https-model-proto` (valid) -> `tls-grpc-jaeger` (valid)
- `grpc-http` (valid) -> `grpc-http`(valid) ->`grpc-model-proto` (valid) -> `grpc-jaeger` (valid)

Jaeger doc:
image

In the past, port names have been changed quite frequently to match the [istio naming conventions](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection). To create some overview, here is a list of all interesting changes:
- https://github.com/jaegertracing/jaeger-operator/pull/911
- https://github.com/jaegertracing/jaeger-operator/pull/1122
- https://github.com/jaegertracing/jaeger-operator/pull/1368
- https://github.com/jaegertracing/jaeger-operator/pull/1862

Istio prefers protocols specified in the [appProtocol](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol) field over the assigned names. Maybe we should think about using this feature in the future. It is supported since kubernetes 1.18. The oldest version we currently support is 1.19.

closes https://github.com/jaegertracing/jaeger-operator/issues/1360

---

cc ) @Stevmann @lautou

### DIFF
--- a/pkg/service/collector_test.go
+++ b/pkg/service/collector_test.go
@@ -66,13 +66,13 @@ func TestCollectorGRPCPortName(t *testing.T) {
 		{
 			"nil",
 			nil,
-			"grpc-http",
+			"grpc-jaeger",
 			false, // in openshift?
 		},
 		{
 			"no-tls",
 			&v1.Jaeger{},
-			"grpc-http",
+			"grpc-jaeger",
 			false, // in openshift?
 		},
 		{
@@ -84,7 +84,7 @@ func TestCollectorGRPCPortName(t *testing.T) {
 					},
 				},
 			},
-			"grpc-http",
+			"grpc-jaeger",
 			false, // in openshift?
 		},
 		{
@@ -96,7 +96,7 @@ func TestCollectorGRPCPortName(t *testing.T) {
 					},
 				},
 			},
-			"grpc-http",
+			"grpc-jaeger",
 			false, // in openshift?
 		},
 		{
@@ -108,13 +108,13 @@ func TestCollectorGRPCPortName(t *testing.T) {
 					},
 				},
 			},
-			"grpc-https",
+			"tls-grpc-jaeger",
 			false, // in openshift?
 		},
 		{
 			"in-openshift",
 			&v1.Jaeger{},
-			"grpc-https",
+			"tls-grpc-jaeger",
 			true, // in openshift?
 		},
 	} {
--- a/pkg/service/collector.go
+++ b/pkg/service/collector.go
@@ -96,17 +96,22 @@ func GetNameForHeadlessCollectorService(jaeger *v1.Jaeger) string {
 	return util.DNSName(util.Truncate("%s-collector-headless", 63, jaeger.Name))
 }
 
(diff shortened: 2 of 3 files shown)


kevinearls (dismissed):

pkg/service/collector_test.go
```diff
@@ -66,13 +66,13 @@ func TestCollectorGRPCPortName(t *testing.T) {
 		{
 			"nil",
 			nil,
-			"grpc-http",
+			"grpc-model-proto",
```
pavolloffay, line 69: I would be more specific and call it `grpc-jaeger`  It as well aligns with OTEL port name. 

There is only a single Jaeger ingestion gprc protocol and I don't think there will be another one.

frzifus: I wonder if then `tls-grpc-jaeger` would work too. I would prefer that name over `https-jaeger` just to avoid future confusion. :x: 

grafik

pavolloffay: Maybe we should ask review form the istio/maistra team on this.

frzifus: y, iam waiting. :)

pavolloffay:
@frzifus did you check the query and all-in-one as well? The query probably exposes 14250 for query V2 and all-in-one all the ports.

Stevmann:
Is there any possiblity/workaround please, how to force operator to rename grpc port name to https-grpc (or any working one)? 
Jaeger collector, based on CLI Flags doesn't support this type of configuration. :(

pkg/service/collector_test.go
frzifus: On a test cluster @iblancasa and I could confirm that replacing the `https` prefix with `tls` worked fine.

frzifus:
@pavolloffay yes, there are not issues.

Query:
https://github.com/jaegertracing/jaeger-operator/blob/b003836ceeec14a313e1e528316efd0403a0b74a/pkg/service/query.go#L88-L94

Agent:
https://github.com/jaegertracing/jaeger-operator/blob/b003836ceeec14a313e1e528316efd0403a0b74a/pkg/service/agent.go#L38-L57

frzifus:
@Stevmann no i dont think so.

frzifus:
Let me close and reopen the pr to trigger the CI again. For some reason its stuck.

iblancasa (approved):

kevinearls (approved):
LGTM

PR #2120, opened 2022-11-16
Fix query ingress when using streaming strategy

### LINKED ISSUE #2088: [Bug]: Query Ingress not properly created with streaming strategy
### What happened?

If you define a path in jaeger query, that path is used for the ingress in deployment strategy == production.
The same doesn't happen if you switch to deployment strategy == streaming.

https://github.com/jaegertracing/jaeger-operator/blob/v1.38.0/pkg/ingress/query.go#L86

### Steps to reproduce

1. Create a Jaeger with production strategy, and set a base path in query. Something like:

```
spec:
  strategy: production
  query:
    options:
      query:
        base-path: /jaeger
...
```

you'll see an ingress with a rule similar to this:
```
spec:
  rules:
  - http:
      paths:
      - backend:
          service:
            name: jaeger-query
            port:
              number: 16686
        path: /jaeger
        pathType: ImplementationSpecific
```

2. Do the same with `strategy: streaming`

Now the ingress only has a default backend service:

```
spec:
  defaultBackend:
    service:
      name: jaeger-query
      port:
        number: 16686
```

### Expected behavior

The ingress should not be different. 

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

1.38.0

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

kubernetes

### Deployment configs

_No response_

kevinearls:

This resolves #2088 If a base path is set in the CR and the streaming strategy is being used, we were not creating the ingress correctly.

### DIFF
--- a/pkg/ingress/query.go
+++ b/pkg/ingress/query.go
@@ -81,10 +81,10 @@ func (i *QueryIngress) Get() *networkingv1.Ingress {
 func (i *QueryIngress) addRulesSpec(spec *networkingv1.IngressSpec, backend *networkingv1.IngressBackend) {
 	path := ""
 
-	if allInOneQueryBasePath, ok := i.jaeger.Spec.AllInOne.Options.StringMap()["query.base-path"]; ok && i.jaeger.Spec.Strategy == v1.DeploymentStrategyAllInOne {
-		path = allInOneQueryBasePath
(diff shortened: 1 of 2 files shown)


frzifus (dismissed):
uff, would it make sense to persist that with a unittest?

kevinearls:
@frzifus Yea, let me take a look.  I don't think there was a test for the current code so I may have to start from scratch.

frzifus (approved):
lgtm :)

frzifus (approved):

PR #1605, opened 2021-11-02
Add support for custom liveness probe

ricoberger:
## Which problem is this PR solving?
- Resolves #1539 

## Short description of the changes
- This PR adds a new property `livenessProbe` which allows a user to overwrite the default liveness probe set by the operator.
- Please see #1539 for the complete description of the problem.

**Notes:**

- I will add some tests, when this change is accepted.
- ~Should this also adapted for the other deployments agent and all-in-one?~
- Should I add the same logic for the readiness probe, just for completeness?

PR #2122, opened 2022-11-21
Fix Liveness Probe for Ingester and Query

ricoberger:

The liveness probes added in https://github.com/jaegertracing/jaeger-operator/pull/1605 are not working for the ingester and query, because of a copy / paste mistake (sorry for that 😞).

This PR fixes the liveness probe configuration for the ingester and query.


frzifus (approved):
oh i see :facepalm: good catch!

Issue #2085, opened 2022-10-10
[Bug]: Jaeger operator cannot reconcile in non-cluster-wide mode


sereneshikari:
### What happened?

I am trying to install the Jaeger operator in non-cluster-wide mode to watch a specific namespace (its own) but it is unable to reconcile the `Jaeger` custom resources created in the same namespace

### Steps to reproduce

Follow the steps in the [documentation](https://www.jaegertracing.io/docs/1.38/operator/#installing-the-operator-on-kubernetes) to install the Jaeger operator to watch a specific namespace:
1. > `kubectl create namespace jaeger`
2. Label the namespace for Istio sidecar injection: `kubectl label ns jaeger istio-injection=enabled` (I don't think this is relevant)
3. Download the [Kubernetes installation manifest for the Jaeger operator](https://github.com/jaegertracing/jaeger-operator/releases/tag/v1.38.0) to a file `jaeger-operator.yaml`
4. > [E]dit the deployment files to change `observability` to the desired namespace value
5. > [C]hange the `ClusterRole` and `ClusterBindingRole` of the operator manifest to `Role` and `RoleBinding`, also set the `WATCH_NAMESPACE` environmebt variable on the jaeger operator Deployment.
6. Remove `Role` and `(Cluster)RoleBinding` resources related to OpenShift metrics (I don't think this is relevant)
7. > `kubectl create -f jaeger-operator.yaml -n jaeger`
(see complete manifest in "Deployment configs" section)
8. Create a simple `Jaeger` CR:
  ```console
  kubectl apply -n jaeger -f - <<EOF
  apiVersion: jaegertracing.io/v1
  kind: Jaeger
  metadata:
    name: simplest
  EOF
  ```

### Expected behavior

As per the [documentation](https://www.jaegertracing.io/docs/1.38/operator/#quick-start---deploying-the-allinone-image):
> In a few seconds, a new in-memory all-in-one instance of Jaeger will be available, suitable for quick demos and development purposes.

### Relevant log output

Logs from the Jaeger operator pod. Please see the "Additional context" section for log "highlights"

[`jaeger-operator-logs.txt`](https://github.com/jaegertracing/jaeger-operator/files/9744354/jaeger-operator-logs.txt)

### Additional context

Environment context:
- The operator is installed into a [Rancher RKE2](https://github.com/rancher/rke2) cluster (`v1.24.6+rke2r1`)
- `cert-manager` (`v1.9.0`) is installed in the cluster
[...]

PR #2126, opened 2022-11-29
limit the get of deployments to WATCH_NAMESPACE on sync

rubenvp8510:

- Fixes https://github.com/jaegertracing/jaeger-operator/issues/2085

-  Check scope of the operator and fetch only deployments that are in the WATCH_NAMESPACES if it's not empty
-  If WATCH_NAMESPACES is not empty (non-cluster wide mode) the deployment webhook only process deployments for those namespaces.


frzifus (commented):
N1, would it make sense to add a e2e test to make sure it doesnt break in the future?

rubenvp8510:
> N1, would it make sense to add a e2e test to make sure it doesnt break in the future?

Yes, I'm on it  :)

config/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - ../default
+
+components:
+  - ./patch
```
frzifus, line 8: missing new line

config/namespaced/patch/kustomization.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+patches:
```
frzifus, line 3: can you explain the patch? ^^

frzifus: Thank you!

frzifus (approved):

frzifus (approved):

PR #2139, opened 2022-12-21
support e2e tests on multi architecture environment

jkandasa:
Signed-off-by: Jeeva Kandasamy <jkandasa@gmail.com>

- currently e2e tests asserts container image is build only for linux/amd64 architecture
- I have updated to the code to generate multi architecture container image

- takes `tests/assert-jobs` directory commit id as image tag to avoid frequent build and push
- if the image is available in the remote registry takes from there. skips local build
- if the image is not available in the remote registry, builds multi arch container image and pushes into remote registry.


tests/e2e/Makefile
```diff
@@ -3,7 +3,9 @@ include $(shell find tests/e2e -name "Makefile" ! -wholename "tests/e2e/Makefile
 export VERTX_IMG ?= jaegertracing/vertx-create-span:operator-e2e-tests
 export ELASTIC_IMG ?= docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.6
 export OPERATOR_IMAGE_NEXT ?= ${IMG_PREFIX}/jaeger-operator:next
-export ASSERT_IMG ?= ${IMG_PREFIX}/asserts-e2e:$(shell date +%s)
+# get "tests/assert-jobs" directry commit id and use that as build tag to avoid frequent build and push container image into a remote registry
+export ASSERT_IMG ?= ${IMG_PREFIX}/asserts-e2e:$(shell git log -n 1 --pretty=format:%h ./tests/assert-jobs/)
```
iblancasa, line 7: This change should not be needed. The image name was stored in a file that avoids triggering the image build. This mechanism should be used again

tests/build-utils/push-assert-e2e-img.sh
```diff
@@ -21,4 +21,4 @@ else
     else
         echo "The image $ASSERT_IMG is in the registry. Not pushing"
     fi
-fi
+fi
```
iblancasa, line 24: Missing new line at the end of the file

tests/build-utils/build-assert-e2e-img.sh
```diff
@@ -6,5 +6,23 @@ set -xe
...
+    manifest=$(docker manifest inspect "$image_name" 2>/dev/null || true)
+    if [ -z "$manifest" ]; then
+        echo "the e2e test asserts container image is not available in the remote registry. locally building and pushing into remote registry"
+        docker buildx build --push \
+            --progress=plain \
+            --platform ${PLATFORMS} \
+            --file Dockerfile.asserts \
+            --tag ${image_name} .
+    else
+        echo "the e2e test asserts container image is available in the remote registry. pulling into local environment"
```
iblancasa, line 23: Why do you need to pull it? The cluster can get it from the registry

iblancasa (changes requested):

tests/e2e/Makefile
jkandasa: in downstream, I do not want to build the image on each run. the name stored in a file only valid for that run.

tests/build-utils/build-assert-e2e-img.sh
jkandasa: in case if you enable `MULTI_ARCH_ASSERT_IMG` in a test environment where `KIND` used
https://github.com/jaegertracing/jaeger-operator/blob/53f3e0ac1bc6209b810c657c39c1fbda3a1aaf10/tests/build-utils/push-assert-e2e-img.sh#L13

tests/e2e/Makefile
iblancasa: In that case, something that can be done in the future is to provide a way to get the image from a registry if its build

tests/build-utils/build-assert-e2e-img.sh
iblancasa: I'm not sure about this. If the images was pushed, it should be able to get it (with the vertx, it happens)

tests/e2e/Makefile
frzifus: @iblancasa you mean something like, build if not exist?

iblancasa: Yes. But it something we can improve in the future

frzifus (approved):

iblancasa (approved):

PR #2243, opened 2023-06-26
[chore] Add Dockerfile for OpenShift containerized test execution on Prow CI.

IshwarKanse:
We will be running end-to-end tests for Jaeger Operator in our [OpenShift release](https://github.com/openshift/release) using Prow CI jobs. These jobs will encompass various cluster configurations, and the Dockerfile will containerize the tests for execution within the Prow CI.

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,0 +1,48 @@
+# The Dockerfile's resulting image is purpose-built for executing Jaeger Operator e2e tests within the OpenShift release (https://github.com/openshift/release) using Prow CI.
+FROM golang:1.20
+
+# Set the user to root
+USER root
+
+# Create the /tmp/go/bin directory, and grant read and write permissions to all users
+RUN mkdir -p /tmp/go/bin \
+    && chmod -R 777 /tmp/go/bin
+
+# Set the build cache directory within /tmp and make it writable
+ENV GOCACHE=/tmp/.cache/go-build
+RUN mkdir -p $GOCACHE \
+    && chmod -R 777 $GOCACHE
+
+# Copy the repository files
+COPY . /tmp/jaeger-operator
+
+# Install kubectl and oc
+RUN curl -L -o oc.tar.gz https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest/openshift-client-linux.tar.gz \
+    && tar -xvzf oc.tar.gz \
+    && chmod +x kubectl oc \
+    && mv oc kubectl /usr/local/bin/ \
+    && rm oc.tar.gz
+
+# Set the working directory
+WORKDIR /tmp/jaeger-operator
+
+# Set the Go path
+ENV GOPATH=/tmp/go
+ENV GOBIN=/tmp/go/bin
+ENV PATH=$PATH:$GOBIN
+
+# Install required dependencies
+RUN ./hack/install/install-kuttl.sh \
+    && ./hack/install/install-golangci-lint.sh \
+    && ./hack/install/install-goimports.sh \
+    && ./hack/install/install-yq.sh \
+    && ./hack/install/install-kustomize.sh \
+    && ./hack/install/install-gomplate.sh \
+    && ./hack/install/install-dependencies.sh \
+    && ./hack/install/install-kubebuilder.sh \
+    && ./hack/install/install-controller-gen.sh \
+    && go install github.com/RH-QE-Distributed-Tracing/junitcli/cmd/junitcli@v1.0.6 \
+    && cp ./bin/kubectl-kuttl /usr/local/bin/kubectl-kuttl
+
+#Make required directories writable as Prow CI doesn't allow root user inside the container.
+RUN chmod -R 777 /tmp/go /tmp/go/bin /tmp/.cache/


tests/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+WORKDIR /tmp
+
+# Install kuttl
+RUN curl -L -o kuttl https://github.com/kudobuilder/kuttl/releases/download/v0.15.0/kubectl-kuttl_0.15.0_linux_x86_64 \
+    && chmod +x kuttl \
+    && mv kuttl /usr/local/bin/kuttl
+
+# Install kubectl and oc
+RUN curl -L -o oc.tar.gz https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest/openshift-client-linux.tar.gz \
```
iblancasa, line 15: Maybe you can use the scripts from this repo to install the tools. Doing that, we ensure we use the same version as other parts of the automation. And we don't have to update multiple sources when increasing the version of one of those tools

IshwarKanse: I have updated the Dockerfile, checking if there are any issues encountered when running the tests. I've skipped building the assert image as we will be providing a pre-built image as env var inside the container.

IshwarKanse: @iblancasa I ran the tests using image built from the Dockerfile and the tests are running fine. Can you help to get this merged. Thanks.

tests/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# The Dockerfile's resulting image is purpose-built for executing Jaeger Operator e2e tests within the OpenShift release (https://github.com/openshift/release) using Prow CI.
+FROM golang:1.20
+
+# Set the user to root
+USER root
+
+# Copy the repository files
+COPY . /tmp/jaeger-operator
+
+WORKDIR /tmp
```
pavolloffay, line 10: This could be removed as it is set below to `WORKDIR /tmp/jaeger-operator`

IshwarKanse: @pavolloffay removed that workdir.

tests/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    && ./hack/install/install-golangci-lint.sh \
+    && ./hack/install/install-goimports.sh \
+    && ./hack/install/install-yq.sh \
+    && ./hack/install/install-kustomize.sh \
+    && ./hack/install/install-gomplate.sh \
+    && ./hack/install/install-dependencies.sh \
+    && go install github.com/RH-QE-Distributed-Tracing/junitcli/cmd/junitcli@v1.0.6 \
+    && cp ./bin/kubectl-kuttl /usr/local/bin/kubectl-kuttl
+
+# Remove the "make prepare-e2e-tests" string from the script ./hack/run-e2e-test-suite.sh
```
pavolloffay, line 30: I would recommend adding this why the removal is needed

tests/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN sed -i 's/make prepare-e2e-tests //g' ./hack/run-e2e-test-suite.sh
+
+# Remove the "load-assert-job" string from the Makefile in the directories in ./tests/e2e/
```
pavolloffay, line 33: I would recommend adding this why the removal is needed

IshwarKanse: The load-assert-job tries to build the assert image which will not work as the tests will be run inside a pod. I have added steps to build the asserts-e2e image in the Prow job and set the env var ASSERT_IMG with the built image, which will be used in the test run.

tests/Dockerfile
IshwarKanse: The target runs 

.PHONY: prepare-e2e-tests
prepare-e2e-tests: start-kind load-operator-image set-image-controller build-assert-job

These are run if we are using kind. Since we are running the tests on OpenShift and not building the assert-e2e image, we need to skip the target. With the target the job fails as it tries to build the assert-e2e image inside a container.

iblancasa: Instead of doing this hack... why don't you add something to the script to skip the execution of that line?

IshwarKanse: @iblancasa Its not just the prepare-e2e-tests, its also load-assert-job in the Makefiles in tests/e2e directories.

tests/e2e/Makefile
```diff
@@ -15,9 +15,16 @@ SKIP_ES_EXTERNAL ?= false
 export SKIP_KAFKA ?= false
 export E2E_TESTS_TIMEOUT ?= 330
 
+#Skip prepare-e2e-tests and load-assert-job when tests are run on Prow CI
+PROW_CI ?= false
```
pavolloffay, line 19: I would avoid try to avoid putting this into the Makefile, so far there is no knowledge in the makefile about the test infra. 

The docker image for prow in this PR can prepare tests for prow or the prow config could actually remove the jobs that do not run on prow.

IshwarKanse: Ok, I'll do the cleanup using the script that executes the tests in Prow. Updated and removed the changes.

pavolloffay (approved):

iblancasa (approved):

PR #2272, opened 2023-07-24
Added changes to respect env variable set from envFrom configMaps

### LINKED ISSUE #1349: Jaeger Operator does not consider variables set via envFrom
Hey there.

When using the sidecar injection of the Jaeger Operator, it checks if the following environment variables are set:
- `JAEGER_SERVICE_NAME`
- `JAEGER_PROPAGATION`

https://github.com/jaegertracing/jaeger-operator/blob/bf1bc275353a0daf99dbebc7a65ea755cd7afaa1/pkg/inject/sidecar.go#L309

However, it only checks if the container spec template directly contains the environment variables.
We manage the environment variables in ConfigMaps which are included in the spec template using `envFrom`.
The `JAEGER_SERVICE_NAME` in the included ConfigMap is not detected, and the default value is set (<app>.<namespace>), which in turn overwrites the value which would be set by the ConfigMap.

P.s.: Thanks for the awesome tools you are providing!

hiteshwani29:
- Resolves #1349 

- We are checking that `JAEGER_SERVICE_NAME` or `JAEGER_PROPAGATION` is present in the envFrom configMaps , if it is present we are not adding these variables in the env variables.

### DIFF
--- a/pkg/inject/sidecar.go
+++ b/pkg/inject/sidecar.go
@@ -43,8 +43,20 @@ const (
 	envVarHostIP      = "HOST_IP"
 )
 
+type SidecarOptions struct {
+	EnvConfigMaps []corev1.ConfigMap
+}
+
+type Options func(f *SidecarOptions)
+
+func WithEnvFromConfigMaps(configMaps []corev1.ConfigMap) Options {
+	return func(s *SidecarOptions) {
+		s.EnvConfigMaps = configMaps
+	}
+}
+
 // Sidecar adds a new container to the deployment, connecting to the given jaeger instance
-func Sidecar(jaeger *v1.Jaeger, dep *appsv1.Deployment) *appsv1.Deployment {
+func Sidecar(jaeger *v1.Jaeger, dep *appsv1.Deployment, opts ...Options) *appsv1.Deployment {
 	deployment.NewAgent(jaeger) // we need some initialization from that, but we don't actually need the agent's instance here
 	logFields := jaeger.Logger().WithValues("deployment", dep.Name)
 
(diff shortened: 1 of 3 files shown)


pkg/inject/sidecar.go
```diff
@@ -485,3 +513,21 @@ func isContainerPortAvailable(port int32, dep *appsv1.Deployment) bool {
 	}
 	return true
 }
+
+// GetEnvFromConfigMaps return configMap which matches with configMapRef
```
iblancasa, line 517:
```suggestion
// GetConfigMapsMatchedEnvFromInDeployment return configMap which matches with configMapRef
```

pkg/inject/sidecar.go
```diff
@@ -485,3 +513,21 @@ func isContainerPortAvailable(port int32, dep *appsv1.Deployment) bool {
...
+func GetConfigMapsMatchedEnvFromInDeplyment(dep appsv1.Deployment, configMaps []corev1.ConfigMap) []corev1.ConfigMap {
+	configMapSearchMap := make(map[string]corev1.ConfigMap)
+	for _, cm := range configMaps {
+		configMapSearchMap[cm.Name] = cm
+	}
```
iblancasa, line 522: I'm not sure if it is a good idea to copy the ConfigMaps in a slice just to do the search. Can we avoid that?

iblancasa (changes requested):

pkg/inject/sidecar.go
hiteshwani29: "Thank you for your feedback! "
Instead of iterating over the entire configMaps slice for each envConfigMap.ConfigMapRef, I've created a map where the keys are the names of the corev1.ConfigMap objects and the values are the corev1.ConfigMap objects themselves.
With this map in place, we can directly find the matching corev1.ConfigMap for a given envConfigMap.ConfigMapRef.Name without the need for multiple iterations over the configMaps slice. This optimization reduces the time complexity of finding a matching corev1.ConfigMap

iblancasa (approved):

iblancasa:
@rubenvp8510 can you take a look?

rubenvp8510 (approved):

PR #2283, opened 2023-08-03
Added the required labels in the ServiceAccount during upgrade

parauliya:
Through this PR the the existing service accounts in the Jaegar namespace will be labelled with following two labels:
```
"app.kubernetes.io/instance":   <same as jaeger>,
"app.kubernetes.io/managed-by": "jaeger-operator",
```

### DIFF
--- a/pkg/controller/jaeger/upgrade.go
+++ b/pkg/controller/jaeger/upgrade.go
@@ -18,6 +18,11 @@ func (r *ReconcileJaeger) applyUpgrades(ctx context.Context, jaeger v1.Jaeger) (
 
 	currentVersions := version.Get()
 
+	err := upgrade.AddLabelsToExistingServiceAccounts(ctx, r.rClient, jaeger)
+	if err != nil {
+		return jaeger, tracing.HandleError(err, span)
+	}
+
 	if len(jaeger.Status.Version) > 0 {
 		if jaeger.Status.Version != currentVersions.Jaeger {
 			// in theory, the version from the Status could be higher than currentVersions.Jaeger, but we let the upgrade routine
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -8,6 +8,7 @@ import (
 	"github.com/Masterminds/semver"
 	"github.com/spf13/viper"
 	"go.opentelemetry.io/otel"
+	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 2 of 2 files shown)


pkg/controller/jaeger/upgrade.go
```diff
@@ -18,6 +18,11 @@ func (r *ReconcileJaeger) applyUpgrades(ctx context.Context, jaeger v1.Jaeger) (
 
 	currentVersions := version.Get()
 
+	err := upgrade.AddLabelsToExistingServiceAccounts(ctx, r.rClient, jaeger)
```
iblancasa, line 21: This applies to all the upgrades. Instead of that, we need to check from what version we're adding the labels to the Service Accounts and create one upgrade routine for that version.

parauliya: This was being checked from version `0.11.0` and the latest version is `1.46.0`. I don't think any one will be using version before a stable one which should be `1.0.0`. So I think this should apply all the future releases.

iblancasa: I agree with the first statement. Maybe we can just add this change to one of the previous versions. I would not like to do this every time because users that are not affected by this, will have their service accounts modified.

Even, I would maybe just document the problem and close the issue.

parauliya: `I would not like to do this every time because users that are not affected by this, will have their service accounts modified.`
In the above case, I think documenting the issue and ask users to have a service account name in Jaeger should not same as their existing service account names.
What do you think and also if we decided to document this where we should documents this, I mean is there any doc file where special cases or prerequisites are mentioned?

iblancasa: Since we don't have any special documentation for the known issues and others, I would add the final explanation to the issue as a comment. Maybe, ask in https://github.com/jaegertracing/documentation to add a section in the K8s Operator section to document this kind of issues.

parauliya: I don't see any K8s Operator section in [https://github.com/jaegertracing/documentation](https://github.com/jaegertracing/documentation), These documents are mostly related to Jaeger Tracing website.
I think as you suggested put a comment to the issue in detail and close the issue.

iblancasa: https://github.com/jaegertracing/documentation/blob/main/content/docs/1.47/operator.md

parauliya: I have created a new [issue](https://github.com/jaegertracing/documentation/issues/664) in the [documentation](https://github.com/jaegertracing/documentation) repo to address this issue.
Now we can close this PR and the issue as well.

parauliya: > I have created a new [issue](https://github.com/jaegertracing/documentation/issues/664) in the [documentation](https://github.com/jaegertracing/documentation) repo to address this issue. Now we can close this PR and the issue as well. @iblancasa , your opinion please?

Hi @iblancasa ,
Please respond on this so that this issue and PR can be closed asap.

iblancasa: Ok, let's close the PR and the issue.

PR #2278, opened 2023-08-01
Protect the platform behind a mutex

iblancasa:
## Which problem is this PR solving?
- Continuation of #2276 to fix #1983

## Short description of the changes
- Added a new structure to manage the operator configuration
- Protect the Platform setting behind a mutex

Note: this work needs to be done for other settings too but will be done in follow-up PRs

PR #2287, opened 2023-08-09
 Protect the ESProvisioning setting behind a mutex

### LINKED ISSUE #1983: fatal error: concurrent map read and map write
**Describe the bug**
The error happens sometimes, i guess some process triggers that.
Please ignore `This field was missing!!!!!!!!` as it is our internal message.

Please find attached file with surrouding logs 

[On_demand_report_2022-07-05T10_54_28.180Z_d7538140-fc50-11ec-84ba-cf7f04fb0981.csv](https://github.com/jaegertracing/jaeger-operator/files/9045750/On_demand_report_2022-07-05T10_54_28.180Z_d7538140-fc50-11ec-84ba-cf7f04fb0981.csv)

iblancasa:
- Continuation of #2278 to fix #1983

- Protect the ESProvisioning setting behind a mutex

### DIFF
--- a/pkg/autodetect/main_test.go
+++ b/pkg/autodetect/main_test.go
@@ -176,7 +176,7 @@ func TestAutoDetectWithServerResourcesForGroupVersionError(t *testing.T) {
 
 func TestAutoDetectOpenShift(t *testing.T) {
 	// prepare
-	viper.Set("platform", "auto-detect")
+	viper.Set("platform", v1.FlagPlatformAutoDetect)
 	defer viper.Reset()
 
 	dcl := &fakeDiscoveryClient{}
@@ -213,7 +213,7 @@ func TestAutoDetectOpenShift(t *testing.T) {
 
 func TestAutoDetectKubernetes(t *testing.T) {
 	// prepare
-	viper.Set("platform", "auto-detect")
+	viper.Set("platform", v1.FlagPlatformAutoDetect)
 	defer viper.Reset()
 
 	dcl := &fakeDiscoveryClient{}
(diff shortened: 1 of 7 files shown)


pkg/autodetect/main_test.go
```diff
@@ -88,7 +85,7 @@ func TestStartContinuesInBackground(t *testing.T) {
 	select {
 	case <-done:
 		assert.False(t, viper.GetBool("auth-delegator-available"))
-	case <-time.After(5 * time.Second):
+	case <-time.After(1 * time.Second):
```
frzifus, line 88: Why this change?

pkg/autodetect/main_test.go
```diff
@@ -100,7 +97,7 @@ func TestStartContinuesInBackground(t *testing.T) {
 			if viper.GetBool("auth-delegator-available") {
 				break
 			}
-			time.Sleep(10 * time.Millisecond)
+			time.Sleep(500 * time.Millisecond)
```
frzifus, line 100: same here?

pkg/autodetect/operatorconfig.go
```diff
@@ -24,6 +24,21 @@ func (p Platform) String() string {
 	return [...]string{"Kubernetes", "OpenShift"}[p]
 }
 
+// ESOperatorIntegration holds the if the ES Operator integration is enabled.
+type ESOperatorIntegration int
+
+const (
+	// ESOperatorIntegrationYes represents the ES Operator integration is enabled.
+	ESOperatorIntegrationYes ESOperatorIntegration = iota
```
frzifus, line 32: Does it make sense to start here with an `ESOperatorIntegrationUnknown` value?

pkg/autodetect/main_test.go
iblancasa: The test is now more resilient. We don't need to wait all that time.

pkg/autodetect/operatorconfig.go
iblancasa: I don't think it is needed

rubenvp8510 (approved):

PR #2356, opened 2023-11-09
Regenerate ES certificated when is close to 1 day for expire

rubenvp8510:
-  This solves the issue in which for some reason the reconciliation process doesn't trigger and the certificates expired, usually the reconciliation process should be triggered due a change in the CR or it's child resources or after 10hours, but it could happens that the certificate expires immediately or very close after a reconciliation call happens. Then we need to wait other 10 hours to have it renewed if nothing else triggers the reconciliation process.

## Description of the changes
-  Check the certificates expiration, if those expires in 1 day, we renew it.


rubenvp8510:
@frzifus Can you review this one please?

frzifus (approved):

iblancasa (approved):

PR #2374, opened 2023-12-07
Choose the newer autoscaling version by default

### LINKED ISSUE #2367: [Bug]: detectAutoscalingVersion always picks `autoscaling/v2beta2`  instead of `autoscaling/v2` if both APIs are available
### What happened?

Hey 👋.

I've been investigating why our `jaeger-operator` instance was still reported by GKE as making calls to the deprecated `autoscaling/v2beta2` even if we updated it to a version that is actually supporting `autoscaling/v2`. 

We're runnign jaeger-operator 1.49.0 on k8s 1.25 and are preparing to upgrade to 1.26

It turns out that [detectAutoscalingVersion](https://github.com/jaegertracing/jaeger-operator/blob/main/pkg/autodetect/main.go#L141) will always pick `autoscaling/v2beta2` if both `v2` and `v2beta2` are available on the cluster.
I suspect it comes from [this break](https://github.com/jaegertracing/jaeger-operator/blob/main/pkg/autodetect/main.go#L156) and the [order in which](https://github.com/jaegertracing/jaeger-operator/blob/main/pkg/autodetect/main.go#L142) the API versions are evaluated. If the cluster exposes both `autoscaling/v2` and `autoscaling/v2beta2`  the latter will always be picked. The "fix" (should this be actually considered as a bug 😅) would be to exit both loops (and not only the inner one) as soon as there is a match.

I'm reporting this as a bug as this behavior is actually causing `jaeger-operator` to be flagged by GKE and it can actually block control plane auto-upgrades.

I would be happy to provide a fix if necessary :)

### Steps to reproduce

1. Run `jaeger-operator` on Kubernetes 1.25
2. See that it stills using the deprecated `autoscaling/v2beta2` API. 

### Expected behavior

As soon as we upgrade jaeger-operator to a version that supports `autoscaling/v2`, we expect it to use it instead of `autosaling/v2beta2`

### Relevant log output

_No response_

### Screenshot

Screenshot 2023-11-27 at 15 00 34

### Additional context

_No response_

### Jaeger backend version

v1.49.0

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

iblancasa:
- Resolves #2367

### DIFF
--- a/pkg/autodetect/main_test.go
+++ b/pkg/autodetect/main_test.go
@@ -534,7 +534,6 @@ func TestAutoDetectAutoscalingVersion(t *testing.T) {
 
 	// verify
 	assert.Equal(t, v1.FlagAutoscalingVersionV2, viper.GetString(v1.FlagAutoscalingVersion))
-
 }
 
 func TestSkipAuthDelegatorNonOpenShift(t *testing.T) {


iblancasa:
@rubenvp8510 can you take a look at this PR? Thanks!

rubenvp8510:
Is it possible to write small unit test for this?

iblancasa:
> Is it possible to write small unit test for this?

Done

rubenvp8510:
small format issue, the CI is complainiing about it

PR #2481, opened 2024-02-22
fixes #2207 by adding server-url to JaegerMetricsStorageSpec

### LINKED ISSUE #2207: [Feature]: Add PROMETHEUS_SERVER_URL to jaeger query spec
### Requirement

 As a Jaeger operator with an independently deployed Prometheus
 I want to specify where my Prometheus lives
 so that I make use of Service Performance Monitoring of my services from within the Jaeger UI

### Problem

[We can specify that we should be using Prometheus, but not how to reach Prometheus](https://github.com/jaegertracing/jaeger-operator/blob/f3eb6284d3bcd3bde75f01820bbd4ca6c7af00e6/pkg/deployment/query.go#L159), despite it being to the `metricsStorage` option.

### Proposal

Add the PROMETHEUS_SERVER_URL setting at the same level as METRICS_STORAGE_TYPE one, a la
```
query:
  metricsStorage: 
    type: prometheus
    server-url: <my-prometheus-here>:9090
```

### Open questions

_No response_

antoniomerlin:
ServerUrl is for providing external prometheus service url. Added test for verifying env variable PROMETHEUS_SERVER_URL is not empty when specifying metric storage type as prometheus for both AllInOne and Query.

- Resolves #2207

## Description of the changes
- ServerUrl is for providing external prometheus service url for Jaeger SPM. And associated tests.

- on k8s multinode cluster

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits
- [x] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully
  - for `jaeger-ui`: `yarn lint` and `yarn test`


cimitan:
thanks!

question: is this example file wrong then? (so we should deprecate)
https://github.com/jaegertracing/jaeger-operator/blob/bbd5b1a2c147911a95c37018dd9929a12df6a75b/examples/all-in-one-with-metrics-query.yaml

note the "server-url" under a different level, although I cannot find references in code

iblancasa:
> thanks!
> 
> question: is this example file wrong then? (so we should deprecate)
> 
> https://github.com/jaegertracing/jaeger-operator/blob/bbd5b1a2c147911a95c37018dd9929a12df6a75b/examples/all-in-one-with-metrics-query.yaml#L14
> 
> note the "server-url" under a different level, although I cannot find references in code

https://github.com/jaegertracing/jaeger-operator/blob/bbd5b1a2c147911a95c37018dd9929a12df6a75b/pkg/deployment/all_in_one.go#L66

iblancasa (approved):

iblancasa:
@rubenvp8510 can you take a look to this PR?

rubenvp8510 (approved):

rubenvp8510:
This looks good, just run make api-docs to regenerate the docs, I think is the reason CI is complaining

Issue #2322, opened 2023-09-11
[Feature]: Add the option to disable the Jaeger Agent component installation  until it is removed completely from code-base 


raviranjanelastisys:
### Requirement

Since the Jaeger Agent is depreciated and Jaeger can accept the traces directly via OTEL SDK , Add the option to disable the Jaeger Agent component installation  until it is removed completely from code-base . Because it is sitting around on the node and using the node resources.

Ex- 

```
jaeger:
  create: true
  spec:
    agent:
      enabled: false  # > This Flag should be present in the Jaeger Operator
```

### Problem

There is no flag `enabled: false` to disable the individual component installation specially `Jaeger-agent`

### Proposal

Add the flag `enabled: false`

iblancasa:
Thanks @raviranjanelastisys for the suggestion. Would you like to send a PR?

raviranjanelastisys:
@iblancasa No , i am not sure how to do this !! But someone from your team would be nice

iblancasa:
#2322

Mactherobot:
What happened to this? Which version was this implemented in?

chinaran:
https://github.com/jaegertracing/jaeger/issues/4739
jaeger-agent has been deprecated.
https://github.com/jaegertracing/jaeger-operator/pull/2611
There is a revert PR here, should it be added back?

yurishkuro:
FYI the jaeger-agent has been removed from Jaeger distributions. https://github.com/jaegertracing/jaeger-operator/issues/2714

PR #2697, opened 2024-10-04
Feature Enabling and Disabling of jaeger-agent

Mactherobot:
Due to deprecation of the jaeger agent, and the fact that jaeger can accept the traces directly via OTEL SDK. I don't know if this change is too simple and will introduce side effects. So please help me, if i'm missing anything.

- Resolves: #2322 

## Description of the changes
- Adds the enabled flag for the jaegerAgentSpec
- Returns nil if the value is not defined or set to false.

- By unit tests

## Checklist
- [X] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [X] I have signed all commits
- [X] I have added unit tests for the new functionality
- [X] I have run lint and test steps successfully
  - for `jaeger-ui`: `yarn lint` and `yarn test`


pkg/deployment/agent.go
```diff
@@ -32,6 +32,15 @@ func NewAgent(jaeger *v1.Jaeger) *Agent {
 
 // Get returns a Agent pod
 func (a *Agent) Get() *appsv1.DaemonSet {
+	// Check if the agent is enabled
+	if a.jaeger.Spec.Agent.Enabled == nil || !*a.jaeger.Spec.Agent.Enabled {
```
iblancasa, line 36:
```suggestion
	if a.jaeger.Spec.Agent.Enabled == nil || !*a.jaeger.Spec.Agent.Enabled {
```
I would enable it by default, since it is the current behavior.

iblancasa (changes requested):
Thanks for your contribution! Please, can add an e2e test

iblancasa (approved):

iblancasa:
You need to regenerate the bundle to make the CI pass.

bundle.Dockerfile-e
```diff
@@ -0,0 +1,19 @@
+FROM scratch
```
iblancasa, line 1: Why is this file needed?

bundle/metadata/annotations.yaml-e
```diff
@@ -0,0 +1,14 @@
+annotations:
```
iblancasa, line 1: Why is this file needed?

iblancasa (changes requested):

bundle.Dockerfile-e
Mactherobot: It was what the make bundle created, so i don't know. Should it just be removed?

iblancasa (approved):

iblancasa:
@Mactherobot I'm not working anymore on this project. Maybe @rubenvp8510 can help you with this.

rubenvp8510:
I Will review this tomorrow. I see the CI is failing but cannot see logs.

frzifus (approved):
LGTM, up to @rubenvp8510

rubenvp8510 (approved):

rubenvp8510:
This LGTM, but basic checks are not passing. Could you please do a force push or push empty commit to trigger it again so we can see why? Thanks

PR #2901, opened 2025-06-06
patch: adding ability to pass labels through to the services for query

drew-viles:
As it stands, we can pass annotations down to the query service using `spec.query.annotations` but any labels reflected in the `spec.query.labels` field only get passed to the deployment. This feels like it may have just been an oversight and to be honest, it's something I actually need as I'm sure some others do.
For my purposes, I need to be able to filter services based on their labels and the labels need to be custom ones such as environment, service, region and more. A long story short, it's for application discovery when using Teleport. so I can ensure certain people can only see certain services.

## Description of the changes
This change sets a new variable called `labels which gets the original labels values of `util.Labels(GetNameForQueryService(jaeger), "service-query", *jaeger)` set as the default.
It then checsk to see if any additional labels have been passed and merges them into the map.

- I've run make test to ensure my tests work as expected.
- I'm running it in my production server right now. I've built and image and pused it to [my repo here](https://hub.docker.com/repository/docker/drewviles/jaeger-operator/general). I've then used this image to launch the operator with and without custom labels and in both cases it has worked as expected.

### With no custom label
image

### With custom label
image

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits
- [x] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully
  - for `jaeger-ui`: `npm run lint` and `npm run test`

### DIFF
--- a/pkg/service/query.go
+++ b/pkg/service/query.go
@@ -17,6 +17,14 @@ func NewQueryService(jaeger *v1.Jaeger, selector map[string]string) *corev1.Serv
 	if jaeger.Spec.Query.Annotations != nil {
 		annotations = jaeger.Spec.Query.Annotations
 	}
+
+	labels := util.Labels(GetNameForQueryService(jaeger), "service-query", *jaeger)
+	if jaeger.Spec.Query.Labels != nil {
+		for k, v := range jaeger.Spec.Query.Labels {
+			labels[k] = v
+		}
+
+	}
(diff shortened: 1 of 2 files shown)


frzifus (approved):
cc @rubenvp8510

PR #2918, opened 2026-08-10
fix: set SPAN_STORAGE_TYPE=grpc for grpc-plugin storage

### LINKED ISSUE #2727: [Bug]: Cannot initialize storage factory: unknown storage type grpc-plugin
### What happened?

I am trying to connect clickhouse using grpc-plugin like here:
https://github.com/jaegertracing/jaeger-clickhouse/blob/main/guide-kubernetes.md

In the jaeger custom resource

if I set storage.type: grpc-plugin

then there is the following error in the jaeger pod log:
```
jaeger 2024/10/31 14:50:09 Cannot initialize storage factory: unknown storage type grpc-plugin. Valid types are [cassandra opensearch elasticsearch memory kafka badger blackhole grpc]
```

if I set storage.type: grpc

then there is the following error in the jaeger-operator pod log:
```
The provided storage type is unknown. Falling back to 'memory'   {"instance": "bm-jaeger", "namespace": "bm-dev-jaeger", "storage": "grpc", "known-options": ["memory","cassandra","elasticsearch","kafka","badger","grpc-plugin"]}
```

### Steps to reproduce

Just install jaeger-operator helm chart to a kubernetes cluster.

### Expected behavior

No errors.

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

clickhouse via grpc plugin

### Operating system

_No response_

### Deployment model

Kubernetes

### Deployment configs

_No response_

pujitha24:
Motivation

When a Jaeger CR sets storage.type: grpc-plugin (the documented way to
connect a gRPC storage plugin such as jaeger-clickhouse), the collector,
query, ingester and all-in-one containers set the SPAN_STORAGE_TYPE
environment variable to the literal string "grpc-plugin". Jaeger's own
storage factory does not recognize "grpc-plugin" as a valid storage
type, only "grpc", so the container fails at startup with:

    Cannot initialize storage factory: unknown storage type grpc-plugin.
    Valid types are [cassandra opensearch elasticsearch memory kafka
    badger blackhole grpc]

"grpc-plugin" is the operator's own CRD-facing identifier for this
storage type (used e.g. for the "grpc-storage-plugin" options prefix
and for wiring the plugin binary init container/volume); it was never
meant to be forwarded verbatim to Jaeger's SPAN_STORAGE_TYPE, which
only understands "grpc".

Approach

Add JaegerStorageType.SpanStorageType() in apis/v1/jaeger_types.go,
mirroring the existing OptionsPrefix() method: it returns "grpc" for
JaegerGRPCPluginStorage and the type unchanged otherwise. Use it at
the four places SPAN_STORAGE_TYPE is set (pkg/deployment/collector.go,
all_in_one.go, query.go, ingester.go) instead of a raw string
conversion. All other uses of the "grpc-plugin" identifier (options
prefix, plugin init-container/volume wiring, CR validation) are
untouched, since only the value forwarded to Jaeger itself was wrong.

Validation

Ran `go build ./...` (passes) and
`go test ./apis/v1/... ./pkg/deployment/...` (passes), including a new
TestSpanStorageType unit test and an extended TestCollectorGRPCPlugin
assertion that the collector container's env includes
SPAN_STORAGE_TYPE=grpc. Confirmed this is a targeted repro: reverting
only the collector.go change and re-running TestCollectorGRPCPlugin
fails with SPAN_STORAGE_TYPE=grpc-plugin (the exact value reported in
the issue), and passes again once the fix is restored. No live cluster
or e2e run was performed; this repo's CONTRIBUTING.md calls unit tests
sufficient for small changes, reserving e2e for new features.

Report: https://github.com/jaegertracing/jaeger-operator/issues/2727

Fixes #2727

### DIFF
--- a/apis/v1/jaeger_types.go
+++ b/apis/v1/jaeger_types.go
@@ -126,6 +126,16 @@ func (storageType JaegerStorageType) OptionsPrefix() string {
 	return string(storageType)
 }
 
+// SpanStorageType returns the value to use for the SPAN_STORAGE_TYPE environment variable
+// passed to the Jaeger components. The operator identifies the gRPC storage plugin as
+// "grpc-plugin", but Jaeger's storage factory only recognizes "grpc".
+func (storageType JaegerStorageType) SpanStorageType() string {
+	if storageType == JaegerGRPCPluginStorage {
+		return "grpc"
+	}
+	return string(storageType)
+}
+
 // JaegerSpec defines the desired state of Jaeger
 type JaegerSpec struct {
 	// +optional
--- a/apis/v1/jaeger_types_test.go
+++ b/apis/v1/jaeger_types_test.go
@@ -14,6 +14,12 @@ func TestElasticsearchPrefix(t *testing.T) {
 	assert.Equal(t, "es", JaegerESStorage.OptionsPrefix())
 }
 
+func TestSpanStorageType(t *testing.T) {
+	assert.Equal(t, "grpc", JaegerGRPCPluginStorage.SpanStorageType())
+	assert.Equal(t, "cassandra", JaegerCassandraStorage.SpanStorageType())
+	assert.Equal(t, "anystorage", JaegerStorageType("anystorage").SpanStorageType())
+}
+
 func TestValidTypes(t *testing.T) {
 	assert.ElementsMatch(t, ValidStorageTypes(),
 		[]JaegerStorageType{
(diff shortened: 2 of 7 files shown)


pujitha24:
This is rebased and green whenever you get a chance to take a look — happy to make any changes that would help.

PR #2919, opened 2026-08-27
Default Ingress path to "/" for Prefix and Exact path types

### LINKED ISSUE #2734: [Bug]:
### What happened?

I used the example https://github.com/jaegertracing/jaeger-operator/blob/main/examples/ingress-with-host-pathType.yaml and here is the final manifest

After applying it, the operator reports an error:
```
2024-11-12T08:49:34Z    ERROR    failed to apply the changes    {"instance": "jaeger", "namespace": "observability", "execution": "2024-11-12T08:49:34Z", "error": "Ingress.extensions \"jaeger-query\" is invalid: spec.rules[0].http.paths[0].path: Invalid value: \"\": must be an absolute path"}
```

What am I doing wrong?

### Steps to reproduce

```
---
kind: Jaeger
metadata:
  name: jaeger
spec:
  allInOne:
    image: nexus.xxxxx:8444/devops/jaeger/all-in-one:1.46.0
  storage:
    type: badger
    options:
      badger:
        ephemeral: false
        directory-key: "/badger/key"
        directory-value: "/badger/data"
        span-store-ttl: "16h"
  volumeMounts:
  - name: data
    mountPath: /badger
  volumes:
  - name: data
    persistentVolumeClaim:
      claimName: jaegerpvc
  query:
    options:
      base-path: "/"
  ingress:
    enabled: true
    pathType: Prefix
    hosts:
      - xxxxx
```

### Expected behavior

Applying ingress.

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

pujitha24:
Motivation:
When spec.ingress.pathType is set to "Prefix" or "Exact" and no
query.base-path option is configured, QueryIngress generates an
IngressRule with an empty path. The Kubernetes API server rejects
this Ingress with "must be an absolute path", since only
PathType ImplementationSpecific permits an empty/arbitrary path;
Prefix and Exact require a path starting with "/". This reproduces
the reported error, e.g. via examples/ingress-with-host-pathType.yaml.

Approach:
In pkg/ingress/query.go's addRulesSpec, once we're in the branch that
builds Ingress Rules (hosts set, or a base-path already configured),
default an empty path to "/" when the resolved pathType is Prefix or
Exact. The defaulting is scoped inside that branch so it does not
change the existing no-hosts/no-base-path behavior, which still falls
back to a single DefaultBackend Ingress.

Validation:
- go build ./... succeeds.
- go test ./pkg/... passes (all packages).
- go test ./pkg/ingress/... -v passes, including two new/updated
  tests: TestQueryIngressWithPathType now asserts the generated path
  is "/" instead of empty (previously it asserted the buggy empty
  path), and a new TestQueryIngressWithExactPathTypeDefaultsPath
  covers PathType Exact. A new
  TestQueryIngressWithoutHostsAndPathTypeKeepsDefaultBackend guards
  against a regression where the defaulting could otherwise leak into
  the no-hosts/no-path DefaultBackend branch.
- Confirmed fail-before/pass-after: reverting only the query.go change
  while keeping the updated tests makes both path-defaulting tests
  fail with "expected: / actual: ", proving the test reproduces the
  reported defect and the fix resolves it.
- go vet ./pkg/ingress/... and gofmt/goimports report no issues on the
  changed files.
- Could not run this repo's pinned golangci-lint (v1.55.2, via
  hack/install/install-golangci-lint.sh) locally: it fails to build
  under the local Go 1.26.5 toolchain. This repo's own CI
  (.github/workflows/base-checks.yaml) runs on Go 1.22, where that
  lint version builds and runs normally, so this is a local toolchain
  limitation, not a defect in this change.
- No model/API/CRD files were touched, so generated-code checks
  (make generate / ensure-generate-is-noop) are unaffected.

Impact: fixes Ingress generation so it is accepted by the Kubernetes
API server when pathType is explicitly set to Prefix or Exact without
a query base-path; behavior for ImplementationSpecific (the default)
and for configurations that already set a base-path is unchanged.

Report: https://github.com/jaegertracing/jaeger-operator/issues/2734
Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #2734

### DIFF
--- a/pkg/ingress/query.go
+++ b/pkg/ingress/query.go
@@ -98,6 +98,12 @@ func (i *QueryIngress) addRulesSpec(spec *networkingv1.IngressSpec, backend *net
 		pathType = networkingv1.PathType(pt)
 	}
 	if len(i.jaeger.Spec.Ingress.Hosts) > 0 || path != "" {
+		// Prefix and Exact path types require an absolute path, unlike ImplementationSpecific.
+		// Default to "/" so the generated Ingress isn't rejected by the API server when no
+		// query base-path has been configured.
+		if path == "" && (pathType == networkingv1.PathTypePrefix || pathType == networkingv1.PathTypeExact) {
+			path = "/"
+		}
 		spec.Rules = append(spec.Rules, getRules(path, &pathType, i.jaeger.Spec.Ingress.Hosts, backend)...)
 	} else {
 		// no hosts and no custom path -> fall back to a single service Ingress
--- a/pkg/ingress/query_test.go
+++ b/pkg/ingress/query_test.go
@@ -160,10 +160,30 @@ func TestQueryIngressWithPathType(t *testing.T) {
 	assert.Len(t, dep.Spec.Rules, 1)
 
 	assert.Len(t, dep.Spec.Rules[0].HTTP.Paths, 1)
-	assert.Empty(t, dep.Spec.Rules[0].HTTP.Paths[0].Path)
+	// Prefix requires an absolute path, so an unset query base-path must default to "/"
+	// rather than producing an Ingress that the API server rejects.
+	assert.Equal(t, "/", dep.Spec.Rules[0].HTTP.Paths[0].Path)
 	assert.Equal(t, networkingv1.PathType("Prefix"), *dep.Spec.Rules[0].HTTP.Paths[0].PathType)
 	assert.Equal(t, "test-host-1", dep.Spec.Rules[0].Host)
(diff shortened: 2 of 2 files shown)


pujitha24:
This has been rebased and green for a few days now — happy to make any changes if something would help move review along.

